PR #25, opened 2016-12-29
Fix Lambda tools readme link

Layoric:
Broken link to Lambda tools readme due to `/src` directory being missed.


PavelSafronov:
All links in the README were broken, so fixed them up in a single commit.

Layoric:
Thanks! 👍

PR #33, opened 2017-01-16
AspNetCoreServer: Update for default 500 error code on exception.

travisgosselin:
Otherwise, even on application exception, a 200 status code is returned to the consumer of your API.


normj:
Thanks for the pull request. I pushed out version 0.8.4-preview1 of the NuGet package that has this fix.

travisgosselin:
Awesome thanks @normj

PR #57, opened 2017-02-05
Update AspNetCoreServer readme

kylegalbraith:
The class to extend is APIGatewayProxyFunction not APIGatewayFunction


normj:
Thanks!

PR #60, opened 2017-02-11
Fixed security group argument instead of subnets.

scott-brady:
I got this error running 'dotnet lambda deploy-function FuncName' with an aws-lambda-tools-defaults.json file populated with 'function-subnets' and 'function-security-groups' for my VPC:

```
Creating new Lambda function FuncName
Error creating Lambda function: Error occurred while DescribeSecurityGroups. EC2
 Error Code: InvalidGroupId.Malformed. EC2 Error Message: Invalid id: "subnet-85
f423ba" (expecting "sg-...")
```

Checked the source and saw it was using the SUBNETS argument instead of SECURITY_GROUPS.


normj:
Thanks! I'll get a release out shortly with the fix.

normj:
I pushed out version 1.2.1-preview1 of Amazon.Lambda.Tools with this fix. Thanks again for the fix.

PR #68, opened 2017-02-22
Add Alexa Event library

kylegalbraith:
Based on what the events are that are recorded in the Amazon developer console for Alexa. Happy to update this to have more specific for attributes, but I couldn't find a predefined event document for Alexa.


normj:
Thanks for the pull request. I haven't had a chance to dive deep in this yet but I'll try and get to that soon.

normj:
Also can you confirm that you are okay with this code being accepted under the Apache 2 license.

kylegalbraith:
Yes I am fine with that.

kylegalbraith:
@normj anything more I can do to drive this forward?

normj:
Sorry, I have been busy with the Visual Studio 2017 and new project system work. I'll try to get to this soon.

normj:
I have been thinking about this for awhile. I think the right long term solution is to port the Java [Alexa Skills Kit](https://github.com/amzn/alexa-skills-kit-java) to .NET Core. By porting the Java high level abstraction documentation for the Java implementation would be valid for a .NET Core implementation as well.

There are several different types of events that Alexa will send and it would be confusing to just have an event class with all possible values for all event types.

Not sure on the timing of when we can get to a port but I would rather aim for that instead of having a basic event implementation now and then another implementation later.

kylegalbraith:
@normj makes sense to me as well. Feel free to close this if it is not of value then, and also please let me know if there is anything I can do to help with the port.

normj:
Thanks for the pull request. I'll let you know when I know more about getting the java abstraction ported to .NET Core.

Issue #74, opened 2017-03-09
Add support for binary response content


ebekker:
As described in [this AWS Developer Forum post](https://forums.aws.amazon.com/thread.jspa?messageID=760374#760374), in order to support binary response content from a Lambda function invoked by API Gateway proxy, the Lambda response needs to include the `isBase64Encoded: true` flag in its JSON response payload.

The most recent revision of the [`APIGatewayProxyResponse`](https://github.com/aws/aws-lambda-dotnet/blob/144df7cb7988ae8fb007d5ea7e172d23faff0d5b/Libraries/src/Amazon.Lambda.APIGatewayEvents/APIGatewayProxyResponse.cs) response model class does not include such a flag and the furthermore, the [`AspNetCoreServer`](https://github.com/aws/aws-lambda-dotnet/tree/master/Libraries/src/Amazon.Lambda.AspNetCoreServer) package would need to provide a means of *triggering* (and *flagging*) the base64-encoding when binary response content is returned from a controller action method.

ebekker:
Adding the additional flag is trivial and would support the basic use-case when authoring C# Lambda functions that handle [`APIGatewayEvents`](https://github.com/aws/aws-lambda-dotnet/tree/master/Libraries/src/Amazon.Lambda.APIGatewayEvents).

However, the `AspNetCoreServer` support is a bit trickier -- a controller action method returns an `IActionResult` instance and its content ultimately gets serialized into a string in all cases.  However, it's not always clear just from that instance, if the raw result should be encoded directly (Base64) or transformed to UTF8.

After some experimentation I'm thinking that a good compromise that would minimize the impact on the normal controller code would be to put some flags and options on the base `APIGatewayProxyFunction` class, since this needs to be sub-classed in any event and at a minimum that subclass would contain Lambda-specific logic such as a call to the extension method `UseApiGateway()`.

I'm thinking about 2 possible approaches to support:
* set a global "always-on" boolean that simply indicates that all responses from all controllers actions should be treated literally and don't do any transformation on anything
* setup a collection of one or more `Content-Type` header values to look for that would *trigger* the Base64-encoding and flagging, for example, when ever a response for `application/octet-stream` or `application/zip` or some image content type -- that would trigger the raw response bytes to be transformed into a Base64-encoding, and it would set the flag `isBase64Encoded` to true in the Labmda response to the API Gateway.

normj:
Closing the issue because it is being tracked in a pull request
[...]

PR #75, opened 2017-03-09
Add support for binary response content

ebekker:
As per #74


Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
```diff
@@ -37,6 +39,16 @@ public abstract class APIGatewayProxyFunction
...
 
         /// <summary>
+        /// If true then all response content will be Base64-encoded before being returned.
+        /// </summary>
+        public bool EnableBinaryResponseContent { get; set; }
+
+        /// <summary>
+        /// An optional set of Content-Types that will trigger encoding the response content in Base64.
+        /// </summary>
+        public IEnumerable<string> BinaryResponseContentTypes { get; set; }
```
normj: What if we switch this around. Maintain a collection of common content types that we don't need to be base 64 encoded like text/html. Don't expose the data collection but add an AddTextContentType operation. Then when processing the response if the content type is not in the collection then base 64 encode it. 

This way we can make it so that must users don't have to anything special to handle binary type.

ebekker: With this change, do you also want to do away with the `XXXContentType` property and only have the one method `AddXXXContentType()` as the sole interface point to this feature?

ebekker: Was thinking about it a bit more -- does it make sense to support it "both ways"?  For example, you can establish a default behavior (to encode or not to encode), and then define a list of *exceptions* -- the content types that alter the default behavior.  And then of course, the default configuration is as you described which would accommodate most users.  Something like:

```csharp
enum ResponseEncoding
{
  String = 0, // Or UTF8 or something...
  Base64 = 1,
}

class APIGatewayProxyFunction
{
  protected Dictionary<string, ResponseEncoding> _responseEncodingForContentType =
      new Dictionary<string, ResponseEncoding>
      {
        // The defaults, such as from https://en.wikipedia.org/wiki/Media_type#Common_examples
        ["text/plain"] = ResponseEncoding.String,
        ["text/html"] = ResponseEncoding.String,
        ["text/css"] = ResponseEncoding.String,
        ["application/xml"] = ResponseEncoding.String,
        ["application/json"] = ResponseEncoding.String,
        // etc...
      };

  public ResponseEncoding DefaultResponseEncoding
  { get; set; } = ResponseEncoding.Base64;

  public void AddResponseEncodingForContentType(string contentType, ResponseEncoding encoding)
  {
    // ...
  }
}
```

Don't want to *explode* this feature change, but just thinking about the possible usage scenarios and what would be most stable for the long-term.

ebekker: Just occurred to me there is one more dependency to consider with whatever interface changes are made to support this, namely that you also have to update (and keep in sync) the [Binary Support](http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings-configure-with-console.html) settings of the API Gateway by enumerating the binary content-types that your app will support.

normj: I like the code changes except I would keep _responseEncodingForContentType private at least for now. I'd rather make it protected later if we find we really need to.

Hopefully the API Gateway setting will be made available in the serverless.template at some point which is where I would expect that to be maintain. Right now that seems beyond the scope of this change.

ebekker:
@normj, I've incorporated all the changes we discussed.

One thing to note, because the default disposition is to treat all responses as binary unless otherwise specified through registrations and/or default encoding property, many of the existing unit tests were failing because they were comparing content that was now being returned Base64-encoded.  For that reason you'll see that most of tests were revised to use a slightly different pattern for looking at the response content body with the help of a new extension method on the response object.

For example, what used to be something like:

```csharp
 Assert.True(response.Body.Length > 0);
```
is now
```csharp
 Assert.True(response.BodyAsString().Length > 0);
```

The extension method now looks at the new `IsBase64Encoded` flag in the response and returns the Body as a string with possible transformation.  As you mention in the comments for the extensions class, this is really only a concern for testing situations, however, it could potentially introduce a breaking change for existing users in their unit tests.

I suppose this is OK, since the AspNetCoreServer package is still considered *experimental*?

normj:
I was testing this feature in my dev branch. I didn't realize that for proxy requests, which is what we use for the ASP.NET Core stuff, the client has to send an Accept header for the binary content type. Seeing as this is likely to be called from a browser which isn't going to send the Accept header is this feature still useful?

ebekker:
I agree the API Gateway (and maybe Lambda?) teams are placing some unusual, and seemingly unnecessary obstacles on making the most of their platforms, but perhaps they'll relax those restrictions at some point.  If they simply allowed for a "default" mapping of binary content type, that would go a long way with many apps, perhaps I'll submit that request...

Regardless, this support for binary content is still quite accessible when dealing with non-browser clients, such as mobile apps or custom clients.  Also, even with conventional browsers, the request `Accept` headers can be managed when dealing with dynamic JavaScript calls (via `XMLHTTPRequest`) which is more and more the norm these days with modern JS app frameworks (Angular, Aurelia, React, etc.) or just good old jQuery.

ebekker:
Actually just discovered something very useful -- it turns out that you can force APIGW + Lambda to always respond with binary content if you simply specify a Binary media type of `*/*` -- in which case all responses will be sent exactly as returned by Lamda, which makes this addition completely relevant for all clients.

ebekker:
@normj, any other suggestions?

muneebm:
@ebekker  for implementing this feature! This would be especially useful for integrating [Swashbuckle](https://github.com/domaindrivendev/Swashbuckle.AspNetCore) to  a net core web API lambda project. We were able to integrate  [Swashbuckle](https://github.com/domaindrivendev/Swashbuckle.AspNetCore)  successfully and can access the swagger documentation through API gateway, the only bottleneck now is getting the swagger ui image files through API gateway and I hope that can be solved by having `IsBase64Encoded` flag in [APIGatewayProxyResponse](https://github.com/aws/aws-lambda-dotnet/blob/144df7cb7988ae8fb007d5ea7e172d23faff0d5b/Libraries/src/Amazon.Lambda.APIGatewayEvents/APIGatewayProxyResponse.cs)

normj:
Sorry I have been off for a bit and coming back to this. 

Initially when I suggested we default to Base64 I assumed everything would be transparent to the user but since we found out it requires additional configuration on the API Gateway side I think we should change `DefaultResponseContentEncoding` to be `Default` instead of `Base64`.

I don't need another pull request for the change. I have things merged into a local branch. I just wanted your thoughts before making that change.

ebekker:
So you're saying that 'DefaultResponseContentEncoding' will be changed, but you would still including the base set of content-type specific mappings (the initial set in `_responseContentEncodingForContentType`)?

I'm OK with that, we would still need to update documentation for folks to make that work (i.e. provide guidance on adding the wildcard `*/*` binary mime-type or a specific type).

Do you have any insight on to the timeline of updating the SAM support for adding the ability to define Binary mappings?  It looks like it stems from a limitation that Binary mappings can't even be defined by CloudFormation at this point, so I guess that team needs to bring that support up to speed first.

normj:
Yes, I just changed DefaultResponseContentEncoding and left _responseContentEncodingForContentType alone. So images will be base64 encoded by default.

Sorry I don't have any information about the SAM support for adding the ability to define binary mappings.

ebekker:
Sounds good to me.

Libraries/src/Amazon.Lambda.AspNetCoreServer/README.md
```diff
@@ -38,3 +38,82 @@ The function handler for the Lambda function will be **TestWebApp::TestWebApp.La
 
 Once the function is deployed configure API Gateway with a HTTP Proxy to call the Lambda Function. Refer to the API Gateway 
 [developer guide](http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-as-simple-proxy.html) for more information.
+
+## Supporting Binary Response Content
+
+The interface between the API Gateway and Lambda provides for and assumes repsonse content to be returned as a UTF-8 string.
```
damianh: typo: `repsonse` -> `response`

Libraries/src/Amazon.Lambda.AspNetCoreServer/README.md
```diff
@@ -38,3 +38,82 @@ The function handler for the Lambda function will be **TestWebApp::TestWebApp.La
...
+In order to return binary content it is necessary to encode the raw response content in Base64 and to set a flag in the
+response object that Base64-encoding has been applied.
+
+In order to facilitate this mechanism, the `APIGatewayProxyFunction` base class maintains a registry of MIME content types
+and how they should be transformed before being returned to the calling API Gateway.  For any binary content types that are
+returned by your applicaiton, you should register them for Base64 tranformation and then the framework will take care of
```
damianh: `applicaiton` -> `application`

Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
```diff
@@ -302,12 +367,44 @@ protected APIGatewayProxyResponse MarshallResponse(IHttpResponseFeature response
...
+                    if (responseFeatures.Body is MemoryStream)
+                    {
+                        bodyBytes = ((MemoryStream)responseFeatures.Body).ToArray();
+                    }
+                    else
+                    {
+                        using (var ms = new MemoryStream())
+                        {
+                            responseFeatures.Body.CopyTo(ms);
+                            bodyBytes = ms.ToArray();
```
damianh: Binary responses (images etc) can ofter be >80KB. Should we be worried about LOH fragementation in the event of the function being re-used? If so, there's a netstandard version of this package https://github.com/Microsoft/Microsoft.IO.RecyclableMemoryStream coming (or just copy the files) that might help.

Libraries/src/Amazon.Lambda.AspNetCoreServer/README.md
normj: Fixed, thanks

Libraries/src/Amazon.Lambda.AspNetCoreServer/README.md
normj: Fixed

Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
normj: Interesting, curious when that package will support .NET Core

normj:
I pushed out version 0.10.0-preview1 of `Amazon.Lambda.AspNetCoreServer` which has this pull request. Thanks for the pull request.

Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
damianh: Shipped earlier today https://github.com/Microsoft/Microsoft.IO.RecyclableMemoryStream/issues/43#issuecomment-297196391

ebekker:
Cool, I'll keep an eye out on progress in CloudFormation and SAM support for the binary mapping, and we can hopefully simplify the interface to this feature in the future.

ebekker:
Regarding the MS pooling, @damianh -- do want to take a crack at incorporating that in?  Let me know if not, I think it could be a valuable addition.

ebekker:
@normj -- can I ask, how did you pull in the changes from this PR?  I'm trying to pull in all upstream changes into my local fork so that I get all caught up, but git is complaining about merge conflicts with all of the commits associated with this PR, even though there are only about a dozen or so lines that differ.

normj:
I don't remember doing anything special. But admittedly I did the merge a while ago and then went on a 2 week vacation:) 

I did have a quite a few merge conflicts I had to deal with because your pull request came in while I was converting all the projects to the new MSBuild format. Probably something with that messed up the commit history compared your fork. Sorry about that.

damianh:
I can take a crack. However `master` branch is not compiling for me. Package restore fails: 

```
C:\dev\aws\aws-lambda-dotnet\Libraries [master-upstream ≡]> dotnet restore
  Restoring packages for C:\dev\aws\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.ConfigEvents\Amazon.Lambda.ConfigEvents.csproj...
  Restoring packages for C:\dev\aws\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.AspNetCoreServer\Amazon.Lambda.AspNetCoreServer.csproj...
  Restoring packages for C:\dev\aws\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.CognitoEvents\Amazon.Lambda.CognitoEvents.csproj...
  Restoring packages for C:\dev\aws\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Core\Amazon.Lambda.Core.csproj...
  Restoring packages for C:\dev\aws\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.APIGatewayEvents\Amazon.Lambda.APIGatewayEvents.csproj...
  Restoring packages for C:\dev\aws\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.KinesisEvents\Amazon.Lambda.KinesisEvents.csproj...
  Restoring packages for C:\dev\aws\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.DynamoDBEvents\Amazon.Lambda.DynamoDBEvents.csproj...
  Restoring packages for C:\dev\aws\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.LexEvents\Amazon.Lambda.LexEvents.csproj...
C:\Program Files\dotnet\sdk\1.0.3\NuGet.targets(97,5): error : Failed to retrieve information about 'AWSSDK.Kinesis' from remote source 'C:\dev\aws\aws-lambda-dotnet\Deployment\nuget-packages'. [C:\dev\aws\aws-lambda-dotnet\Libraries\Libraries.sln]
C:\dev\aws\aws-lambda-dotnet\Libraries [master-upstream ≡]>
```

And compilation `dotnet build` is resulting in a lot of errors. Any guidance?

normj:
Looks like you are missing the official NuGet feed if you can't find AWSSDK.Kinesis.

I am curious how this package improves performance. I'm always leary of adding dependencies so if I have the data showing it improves performance then it makes it easier to decide to take the pull request.

damianh:
> I am curious how this package improves performance. I'm always leary of adding dependencies so if I have the data showing it improves performance then it makes it easier to decide to take the pull request.

Fully agreed. Am dependency allergic also.

damianh:
> Looks like you are missing the official NuGet feed if you can't find AWSSDK.Kinesis.

```
C:\dev\aws\aws-lambda-dotnet\Libraries [master-upstream ≡]> nuget sources
Registered Sources:

  1.  build-artifacts [Enabled]
      C:\dev\aws\aws-lambda-dotnet\./Deployment/nuget-packages
  2.  nuget.org [Enabled]
      https://api.nuget.org/v3/index.json
   ... more ...
```

I have the official feed registered. Looks like it's choking on the the non-existent path defined in the `nuget.config` and I _don't_ have the package already in the cache.

Anyway, will open a separate PR for this, if it turns out to be desirable.

Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
```diff
@@ -302,12 +367,44 @@ protected APIGatewayProxyResponse MarshallResponse(IHttpResponseFeature response
...
                 }
             }
 
             if (responseFeatures.Body != null)
             {
-                if (responseFeatures.Body is MemoryStream)
+                // Figure out how we should treat the response content
+                var rcEncoding = DefaultResponseContentEncoding;
+                if (contentType != null
+                        && _responseContentEncodingForContentType.ContainsKey(contentType))
```
damianh: This is problematic for content types that specify a charset. i.e. `text/plain; charset=utf8`. 

Either specify all variations in `_responseContentEncodingForContentType` dictionary or perhaps ignore the charset altogether `_responseContentEncodingForContentType.ContainsKey(contentType.Split(';')[0])` ?

brendonparker:
I'm trying to leverage this feature for something that needs to return a `multipart/related` SOAP response.

I've added this line:
```
RegisterResponseContentEncodingForContentType("multipart/related", ResponseContentEncoding.Base64);
```
And added `multipart/related` to the _Binary Media Types_ in the API Gateway settings. 

But my response coming back still looks like its getting escaped:
```
&lt;![CDATA[BASE_64_STRING_HERE]]&gt;
```

Been banging my head on this for a while. Thinking I may have to abandon dreams of this being serverless.

PR #77, opened 2017-03-14
Fix ASP.NET Core API tests failing in debug mode

miltador:
The tests where not working under debug mode because it couldn't resolve project path correctly.
More specifically, running tests in debug mode in Visual Studio fails with `FileNotFoundException` because `Directory.GetCurrentDirectory()` equals `C:\Windows\System32`.
This fixes the regression by following Microsoft example here (`TextFixture`):
https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/testing

### DIFF
--- a/Blueprints/BlueprintDefinitions/PreviewAspNetCoreWebAPI/test/Common/TestLambdaEntryPoint.cs
+++ b/Blueprints/BlueprintDefinitions/PreviewAspNetCoreWebAPI/test/Common/TestLambdaEntryPoint.cs
@@ -1,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting;
+using System.IO;
+using System.Reflection;
+using Microsoft.Extensions.PlatformAbstractions;
+
+namespace BLUEPRINT_BASE_NAME.Tests
+{
+    /// <summary>
+    /// This class extends from APIGatewayProxyFunction which contains the method FunctionHandlerAsync which is the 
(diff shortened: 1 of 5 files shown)


normj:
For me the test pass in VS 2015 and 2017 as well as using `dotnet test`. What environment were you seeing the test fail?

miltador:
It passes but have you tried to debug it?08:01, 16 березня 2017 р., Norm Johanson <notifications@github.com>:For me the test pass in VS 2015 and 2017 as well as using dotnet test. What environment were you seeing the test fail?

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.


  
  




-- Надіслано з мобільної програми Яндекс.Пошти

Marcus-L:
I ran into this as well, not being able to debug the tests slowed me down a bit. It'd be great to get that fixed.

socketnorm:
Sorry I haven't responded back sooner. I'm coming back to this after being away for a bit.

I was able to reproduce this before when debugging but now with the latest patches for Visual Studio and .NET Core tooling I'm no longer able to reproduce this in either VS2015 or VS2017. Can you confirm you are still seeing this behavior after applying all the latest updates to VS and .NET Core tooling?

normj:
Sorry I haven't responded back sooner. I'm coming back to

this after being away for a bit.

I was able to reproduce this before when debugging but now with the latest patches for Visual Studio and .NET Core tooling I'm no longer able to reproduce this in either VS2015 or VS2017. Can you confirm you are still seeing this behavior after applying all the latest updates to VS and .NET Core tooling?

normj:
I don't think this is an issue anymore with latest versions of VS and .NET Core tooling. If you believe that is not the case feel free to reopen.

PR #83, opened 2017-03-22
Tidy up new templates

slang25:
2 minor tidy ups
- Removed `using Amazon.Lambda.Serialization;`, it's not needed, the usage is fully qualified (as it should be IMO)
- `LambdaSerializerAttribute` -> `LambdaSerializer` If this was in `AssemblyInfo.cs` this wouldn't matter, but if this is in going in "regular code files", idiomatic C# would usually omit the `Attribute` suffix as it is verbose and redundant.


normj:
Looks good, This change will go out in the next release of the blueprints.

PR #89, opened 2017-03-28
Populate RemoteIpAddress and RemotePort on HttpContext.Connection

### LINKED ISSUE #22: Support IHttpConnectionFeature in AspNetCoreServer
It looks like the implementation of this interface is commented out in `InvokeFeatures` https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/InvokeFeatures.cs#L21

It would be good to support this, mainly to allow easy access to the client's IP address. Currently you have to grab it from the `X-Forwarded-For` header.

Marcus-L:
Fixes #22 

Allows the AspNetCore application to access the remote client ip/port information via HttpContext.Connection, which works when running locally via `dotnet run`.

RemoteIpAddress is obtained from the APIGatewayProxyRequest's RequestContext.Identity.SourceIp
RemotePort is obtained from the APIGatewayProxyRequest's Headers["X-Forwarded-Port"]

For example this would allow for a WebAPI Controller to use the remote info:
```c#
[HttpGet]
public string GetRemote()
{
    var c = HttpContext.Connection;
    return $"{c.RemoteIpAddress}:{c.RemotePort}";
}
```

Updated the test input to provide the source IP, did not add the header "X-Forwarded-Port" in case a better place to get port information from is known by the maintainer.

### DIFF
--- a/Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
+++ b/Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
@@ -7,6 +7,7 @@ using Microsoft.AspNetCore.Http.Features;
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Net;
 using System.Reflection;
 using System.Text;
 using System.Text.Encodings.Web;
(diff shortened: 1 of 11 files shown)


normj:
Thanks for the pull request. I have this merged in a local branch. I'm trying to finish off a couple more things first and then I'll do a release with this change and close out this pull request.

PR #117, opened 2017-05-10
Displaying the lambda tools version for the help command

coreycoto:
`dotnet lambda help` displays the version of the Amazon.Lambda.Tools.

```
AWS Lambda Tools for .NET Core functions (1.5.0)
Project Home: https://github.com/aws/aws-lambda-dotnet
```

### DIFF
--- a/Libraries/src/Amazon.Lambda.Tools/Program.cs
+++ b/Libraries/src/Amazon.Lambda.Tools/Program.cs
@@ -21,9 +21,9 @@ namespace Amazon.Lambda.Tools
                 {
                     attribute = Assembly.GetEntryAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>();
                 }
-                catch (AmbiguousMatchException e)
+                catch (AmbiguousMatchException)
                 {
-                    Console.Error.WriteLine(e.Message);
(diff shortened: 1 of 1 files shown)


Libraries/src/Amazon.Lambda.Tools/Program.cs
```diff
@@ -12,6 +12,23 @@ namespace Amazon.Lambda.Tools
...
+            get
+            {
+                AssemblyInformationalVersionAttribute attribute = null;
+                try
+                {
+                    attribute = Assembly.GetEntryAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>();
+                }
+                catch (AmbiguousMatchException e)
+                {
+                    Console.Error.WriteLine(e.Message);
```
bjorg: you may want to just return null in that case instead of printing a nasty exception

coreycoto: Good call! Updated the PR.

normj:
Thanks for the pull request. It will go out with the next release.

normj:
Version 1.6.0 of Amazon.Lambda.Tools was released with this pull request

PR #142, opened 2017-07-27
Explain project.json to MSBuild/.csproj migration for cross platform users

forestjohnsonpeoplenet:


normj:
Thanks for the pull. I actually forgot that doc section was there. I have actually been moving away from Yeoman in favor of using templates from the `dotnet new` command. That feature was announced back in March, here is the announcement [blog](https://aws.amazon.com/blogs/developer/creating-net-core-aws-lambda-projects-without-visual-studio/). I'm thinking I should put some of those details in the README file instead of Yeoman. 

Given that we have the `dotnet new` starting point which is where I have been keeping the templates updated. Do you think it is worth keeping the Yeoman section or should we keep it in there for Legacy purposes?

normj:
I updated the README.md saying we have are deprecating Yeoman and then how to use templates from the dotnet CLI. 

Thanks for taking the time to update our docs.

PR #184, opened 2017-12-07
Upadate LexEvents.cs, InputTranscript property in not in the correct class

jmeijon:
The InputTranscript property is in the LexCurrentIntent class but must be in the Lex event class
You can confirm it with the Amazon documentation: 
http://docs.aws.amazon.com/lex/latest/dg/lambda-input-response-format.html


normj:
Good catch! I'll get this fix shipped out.

PR #192, opened 2017-12-24
Improve event visibility in tests

toshi0607:
# overview
use json files instead of string in tests

# background
In EventTest,  each test case uses  event deserialized from json files recently. I think it's better than parsing string literal because of visibility and maintainability. So I replaced events from string to json.


normj:
Thanks for the pull request! Working with the JSON in separate files is easier.

PR #231, opened 2018-02-21
Add FSharpJsonSerialiser

rfrerebe:
Hello,

This PR adds a F# Json Converter, that handles proprely F# types (options, discriminated union, ...).
I had to bump version of NewtonSoft.Json


normj:
Thanks for submitting this. I have been looking into porting our Lambda blueprints to F# but my F# skill is pretty poor. 

Like @Kralizek said this change would be a breaking change to .NET Core 1.0 users plus it would add unnecessary dependencies for C# .NET Core 2.0 users in the package bundle as well. 

It looks like the only code change over the existing serializer is adding the **Fable.JsonConverter()** line. If so I wonder if what we should do is have the constructor for the existing serializer take in a collection of converters. Then in the assembly attribute to enable the serializer you could just pass in the **Fable.JsonConverter()**. Does that seem like a fair compromise?

rfrerebe:
Opened #234 instead

PR #232, opened 2018-02-22
Include Kinesis Analytics Events

vlanka:
Including Kinesis Analytics events and response models with tests.


Libraries/src/Amazon.Lambda.KinesisAnalyticsEvents/KinesisAnalyticsFirehoseInputPreprocessingEvent.cs
```diff
@@ -0,0 +1,111 @@
...
+            public class KinesisFirehoseRecordMetadata
+            {
+                /// <summary>
+                /// Gets or sets the approximate arrival timestamp.
+                /// </summary>
+                /// <value>
+                /// The approximate arrival timestamp.
+                /// </value>
+                [DataMember(Name = "approximateArrivalTimestamp")]
+                public long ApproximateArrivalTimestamp { get; set; }
```
normj: Returning longs for dates in .NET makes it hard to use. Users have to understand that it is an Epoch and figure out how to convert from Epoch to DateTime.

We should do what we did for the KinesisFirehoseEvent and rename this property to ApproximateArrivalEpoch and then provide a second read only property called ApproximateArrivalTimestamp that converts the Epoch property to DateTime. Here is the code on how we did this before.

https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.KinesisFirehoseEvents/KinesisFirehoseEvent.cs#L50-L66

Libraries/src/Amazon.Lambda.KinesisAnalyticsEvents/KinesisAnalyticsStreamsInputPreprocessingEvent.cs
```diff
@@ -0,0 +1,137 @@
...
+                public string PartitionKey { get; set; }
+
+                /// <summary>
+                /// Gets or sets the approximate arrival timestamp.
+                /// </summary>
+                /// <value>
+                /// The approximate arrival timestamp.
+                /// </value>
+                [DataMember(Name = "approximateArrivalTimestamp")]
+                public long ApproximateArrivalTimestamp { get; set; }
```
normj: Same comment as before about return dates as long

Libraries/src/Amazon.Lambda.KinesisAnalyticsEvents/KinesisAnalyticsOutputDeliveryEvent.cs
```diff
@@ -0,0 +1,101 @@
...
+            public LambdaDeliveryRecordMetadata RecordMetadata { get; set; }
+
+            /// <summary>
+            /// 
+            /// </summary>
+            [DataContract]
+            public class LambdaDeliveryRecordMetadata
+            {
+                /// <summary>
+                /// Gets or sets the retry hint.
```
normj: Can we add documentation that says retry hints are the number of delivery retries. Otherwise it is a confusing property.

Libraries/src/Amazon.Lambda.KinesisAnalyticsEvents/KinesisAnalyticsFirehoseInputPreprocessingEvent.cs
Kralizek: Wouldn't be cleaner to have the property of the dto of type DateTimeOffset and use a JsonConverter responsible for converting to/from Unix epoch?

Libraries/src/Amazon.Lambda.KinesisAnalyticsEvents/KinesisAnalyticsOutputDeliveryEvent.cs
vlanka: Addressed.

Libraries/src/Amazon.Lambda.KinesisAnalyticsEvents/KinesisAnalyticsFirehoseInputPreprocessingEvent.cs
vlanka: Done.

Libraries/src/Amazon.Lambda.KinesisAnalyticsEvents/KinesisAnalyticsStreamsInputPreprocessingEvent.cs
vlanka: Done

normj:
All merged and NuGet package released https://www.nuget.org/packages/Amazon.Lambda.KinesisAnalyticsEvents/. Thanks for the pull request.

Issue #227, opened 2018-02-13
Move the code out of the APIGatewayProxyFunction in a 'start' method!


arndklocker:
I want to set a variable in my derived class, but at the moment the default constructor is called and all other methods are initialized before my variable in the derived class is set. Is it possible to move the code out of the consturctor in a 'start' method?

Background: I want to set the 'ASPNETCORE_ENVIRONMENT' with the 'InvokedFunctionArn' from 'ILambdaContext'. I have an working implementaton, but I made the changes in my local repository.

normj:
I'd recommend in your derived class to overwrite the `PostCreateContext` method and set the environment variable in that. There are several other PostX methods that you can overwrite to customize the transformation from API Gateway to ASP.NET Core.

arndklocker:
`PostCreateContext` is to late, this method is called in `FunctionHandlerAsync`. I want to set some varialbes before the base constructor does initializing.

`protected APIGatewayProxyFunction(bool autostart = true)
        {
            if(autostart)
            {
                Start();
            }
        }`

normj:
Okay so the code in question is here: https://github.com/aws/aws-lambda-dotnet/blob/0bf318e67ae269e10e61f65811711fafd25d029c/Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs#L83-L97

I assume what your need is have the environment variables set before the WebHost is built, possibly because it affects how the configuration object is built.

And what you need is this code in the constructor to be delayed till the first function invocation with a hook for you to run some code before that with the ILambdaContext. I'd rather not expose the start method because I could see people miss understanding it and calling it for every function invocation and wonder why everything is running so slow. Here is what I think I could do:

- Move the code into a new private start method
- Constructor takes in a autoStart or delayStart boolean, not sure what name I like better, with the default keeping the existing behavior
- Add a new empty virtual method called before processing the request which you could overwrite to set the environment variables.
- In the FunctionHandlerAsync call the new empty virtual method first.
- If the WebHost builder is null then run the new private start method.
- Continuing processing request as done today.

If that works for you I think I can make those changes. Unfortunately I won't be able to get to that till late next week.

arndklocker:
Yes sounds good. Can you make a new nuget package with this feature?

arndklocker:
When will it be done? With this fix, many other current issues can be fixed in derived classes by the community e.g. "set environment at startup"

normj:
[...]

PR #267, opened 2018-04-23
Fix issue #227

arnd27:
*Issue #227*

- Move the code from the consturctor into a new private start method
- Constructor takes in a delayStart boolean, with the default keeping the existing behavior
- If the _server is null then run the new private start method.
- Continuing processing request as done today.

### DIFF
--- a/Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
+++ b/Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
@@ -97,7 +97,7 @@ namespace Amazon.Lambda.AspNetCoreServer
         }
 
         /// <summary>
-        /// Should be called in the derrived constructor 
+        /// Should be called in the derived constructor 
         /// </summary>
         protected void Start()
         {


Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
```diff
@@ -80,7 +80,24 @@ public abstract class APIGatewayProxyFunction
         /// <summary>
         /// Default constructor that AWS Lambda will invoke.
         /// </summary>
-        protected APIGatewayProxyFunction()
+        protected APIGatewayProxyFunction(bool delayStart = false)
+        {
+            if (!delayStart)
+                Start();
```
costleya: Please wrap if bodies with {}, unless it is an exception or a return

Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
```diff
@@ -80,7 +80,24 @@ public abstract class APIGatewayProxyFunction
...
+        private bool IsStarted
+        {
+            get
+            {
+                return _server != null;
+            }
+        }
+
+        /// <summary>
+        /// Should be called in the derrived constructor
```
costleya: spelling - derived

Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
```diff
@@ -173,6 +190,9 @@ protected virtual IWebHostBuilder CreateWebHostBuilder()
         {
             lambdaContext.Logger.LogLine($"Incoming {request.HttpMethod} requests to {request.Path}");
 
+            if (!IsStarted)
+                Start();
```
costleya: if statement body needs { }

Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
```diff
@@ -173,6 +190,9 @@ protected virtual IWebHostBuilder CreateWebHostBuilder()
         {
             lambdaContext.Logger.LogLine($"Incoming {request.HttpMethod} requests to {request.Path}");
```
costleya: You are missing an invocation to a new virtual method that you could override to set the environment variables. 

1. You need a new virtual PreStart method that takes in the ILambdaContext. I wouldn't pass in the request since this method would only be invoked the on the first request, and we want to prevent misuse of this method.

2. You need to call this method in the !IsStarted if statement. Since this method will only be useful for altering the properties of the environment before the service starts, we should isolate this call so it doesn't cause  inadvertent side  effects.

arndklocker:
For what you need a new virtual PreStart method, for me it makes no sense. I use this fix with the following class:

```
    public class APIGatewayProxyFunctionExt : Amazon.Lambda.AspNetCoreServer.APIGatewayProxyFunction
    {
        private string environmentName;

        public APIGatewayProxyFunctionExt(string environmentName, Action<string, IWebHostBuilder> configureDelegate)
            : base(true)
        {
            this.environmentName = environmentName;
            this.configureDelegate = configureDelegate;
            Start();
        }

        /// <summary>
        /// The builder has configuration, logging and Amazon API Gateway already configured. The startup class
        /// needs to be configured in this method using the UseStartup<>() method.
        /// </summary>
        /// <param name="builder"></param>
        protected override void Init(IWebHostBuilder builderRoot)
        {
            configureDelegate(environmentName, builderRoot);
        }

        private Action<string, IWebHostBuilder> configureDelegate;

        protected override IWebHostBuilder CreateWebHostBuilder()
        {
            return new WebHostBuilder()
                .UseApiGateway()
                .ConfigureAppConfiguration((context, builder) =>
                {
                    var env = context.HostingEnvironment;
                    env.EnvironmentName = environmentName ?? "NoEnv";
                });
        }
    }
```

Now I can set the environment.

@costleya: Or should I add this modifications to the base class?

And than I use this such as (I know it's only a workaround):

```
    public abstract class LambdaEntryPointExt
    {        
	private readonly IDictionary<string, APIGatewayProxyFunctionExt> dict;

        public LambdaEntryPointExt()
        {
            this.dict = new ConcurrentDictionary<string, APIGatewayProxyFunctionExt>();
        }

        [LambdaSerializer(typeof(Amazon.Lambda.Serialization.Json.JsonSerializer))]
        public Task<APIGatewayProxyResponse> FunctionHandlerAsync(APIGatewayProxyRequest request, ILambdaContext lambdaContext)
        {
            //arn:aws:lambda:eu-west-1:123456789123:function:demo-web:production
            var envName = lambdaContext.InvokedFunctionArn.Split(':').Last();

            if (!string.IsNullOrEmpty(envName))
                envName = envName.FirstCharToUpper();

            APIGatewayProxyFunctionExt apiGatewayProxyFunctionExt = null;
            if (!dict.Keys.Contains(envName))
            {
                apiGatewayProxyFunctionExt = new APIGatewayProxyFunctionExt(envName, ConfigureContext);
                dict.Add(envName, apiGatewayProxyFunctionExt);
            }
            else
            {
                apiGatewayProxyFunctionExt = dict[envName];
            }              

            return apiGatewayProxyFunctionExt.FunctionHandlerAsync(request, lambdaContext);
        }

        protected abstract void ConfigureContext(string environmentName, IWebHostBuilder builder);
    }
```

The concrete LambdaEntryPoint looks like the following:

```
  public class LambdaEntryPoint : LambdaEntryPointExt
  {
        protected override void ConfigureContext(string environmentName, IWebHostBuilder builder)
        {
            builder
                .AppSharedConfiguration()
                .ConfigureAppConfiguration((context, b) =>
                {
                    b.AddAWSConfigurationFromParameterStore($"/demo-auth/{environmentName.ToLower()}");
                });
        }
  }
```

normj:
Is the main point of this pull request to get the environment name in so you can find the SSM parameter store variables for the environment? If so that seems like an important use case for us to handle and I'm wondering if we have made it too hard to do that.

Also I just discovered you can find the lambda function name during the constructor phase by checking the environment variable ```AWS_LAMBDA_FUNCTION_NAME``` which seems then we don't need to do this change. Here are the list of available environment variables Lambda sets. https://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html

arndklocker:
Thats the problem, but I need the aliase name (e.g. `arn:aws:lambda:eu-west-1:123456789:function:demo-api:<alias>`) and not the `AWS_LAMBDA_FUNCTION_NAME` (e.g.` demo-api`). Your lambda runtime and your library don't set the alias and I don't want to wait for this decision. With my simple pull reqeuest, everyone can extend your library and set what he want.

normj:
Status update. I have merged this into the dev branch to go out with the next release.

PR #308, opened 2018-08-30
Adds More Convenient Typed Startup for ASP.NET Core

chrisoverzero:
*Description of changes:*

Adds an inheritor of `APIGatewayProxyFunction` that allows specifying the application's startup type as a type parameter.

This largely mirrors the method `static IWebHostBuilder CreateDefaultBuilder<TStartup>(string[]) where TStartup : class` provided by ASP.NET Core, and is nice to have for the same reasons as that method is nice to have.

<sub>also corrects the name of the class in a comment</sub>

### DIFF
--- a/Libraries/test/TestWebApp/LambdaFunction.cs
+++ b/Libraries/test/TestWebApp/LambdaFunction.cs
@@ -1,19 +1,8 @@
-﻿using System.IO;
-
-using Amazon.Lambda.AspNetCoreServer;
-using Microsoft.AspNetCore.Hosting;
+﻿using Amazon.Lambda.AspNetCoreServer;
 
 namespace TestWebApp
 {
     public class LambdaFunction : APIGatewayProxyFunction<Startup>
     {
-        public const string BinaryContentType = "application/octet-stream";
-
-        protected override void Init(IWebHostBuilder builder)
-        {
-            builder
-                .UseApiGateway()
-                .UseContentRoot(Directory.GetCurrentDirectory());
-        }
     }
 }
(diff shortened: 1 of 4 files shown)


Libraries/test/TestWebApp/LambdaFunction.cs
```diff
@@ -5,16 +5,15 @@
...
 {
-    public class LambdaFunction : APIGatewayProxyFunction
+    public class LambdaFunction : APIGatewayProxyFunction<Startup>
     {
         public const string BinaryContentType = "application/octet-stream";
 
         protected override void Init(IWebHostBuilder builder)
         {
             builder
                 .UseApiGateway()
```
chrisoverzero: Both of these remaining `Use` methods are already called in `CreateWebHostBuilder`, but it would look odd to have `Init` simply return the builder, I thought, so I left it alone. Unless `Init` became virtual instead of abstract? Way out of scope, I understand.

chrisoverzero: The more I think about it, the more it makes sense. An `IWebHostBuilder` created by the typed variant doesn't _require_ any other setup. I think it should be overridden to provide a default implementation of `Init`.

normj:
Thanks for the pull request. I have merged the change into the dev branch and it will go out with the next release.

chrisoverzero:
That's great. Thank you so much.

normj:
Version 2.1.0 was just released with this pull request. Thanks for the effort!

PR #331, opened 2018-10-02
Update README.md

covamalia:
Some of the Dotnet CLI templates are different to the ones in the Readme, I have updated the list from the results of running `dotnet new -i "Amazon.Lambda.Templates::*"` in powershell today as some are now `serverless.` instead of `lambda.`


normj:
Thanks for the update

PR #371, opened 2018-12-23
Blueprint for Empty Function with Dependency Injection

tonysneed:
*Issue #:* 363

*Description of changes:*
Blueprint for an empty function that includes setup code for .NET Core dependency injection and configuration.

Approach described in the following blog post: https://blog.tonysneed.com/2018/12/16/add-net-core-di-and-config-goodness-to-aws-lambda-functions/

Based on the following GitHub repo: https://github.com/tonysneed/net-core-lambda-dep-resolver

Add a `DependencyResolver` class that accepts a callback for configuring services. Add an `AddConfiguration` extension method for building an `IConfiguration`.  Add a `ConfigureServices` method that builds a standard `IConfiguration` with JSON files and environment variables.


Blueprints/BlueprintDefinitions/Msbuild-NETCore_2_1/EmptyFunctionDepInjection/template/test/BlueprintBaseName.1.Tests/FunctionTest.cs
```diff
@@ -0,0 +1,25 @@
...
+        public void TestConfigFunction()
+        {
+            var expected = "val1";
+            var mockConfig = new Mock<IConfiguration>();
+            mockConfig.Setup(p => p[It.IsAny<string>()]).Returns(expected);
+
+            var function = new Function(mockConfig.Object);
+            var result = function.FunctionHandler("env1", new TestLambdaContext());
+            Assert.Equal(expected, result);
+        }
```
slang25: You could get away with not using a mock here with something like this:
```csharp
var expected = "val1";
var configValues = new Dictionary<string,string>{ ["env1"] = "val1" };
var testConfig = new ConfigurationBuilder().AddInMemoryCollection(configValues).Build();

var function = new Function(testConfig);
var result = function.FunctionHandler("env1", new TestLambdaContext());
Assert.Equal(expected, result);
```
That way you can drop moq and the template becomes a bit more non-prescriptive.

tonysneed: @slang25 Good idea. I'll eliminate the Moq bits.

Blueprints/BlueprintDefinitions/Msbuild-NETCore_2_1/EmptyFunctionDepInjection/template/src/BlueprintBaseName.1/DependencyResolver.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace BlueprintBaseName._1
+{
+    public class DependencyResolver
```
tonysneed: I think I could eliminate `DependencyResolver` and move the `ServiceCollection` code to the `Function` ctor.  Would simplify things. @slang25 What do you think?

tonysneed:
@slang25 Removed Moq and simplified the code by removing DependencyResolver.

One thing I wanted to ask about is **launchSettings.json**. Is it possible to add the following to the "Mock Lambda Test Tool" profile? This is the default for a .NET Core Web API project.

```json
"environmentVariables": {
  "ASPNETCORE_ENVIRONMENT": "Development"
}
```

normj:
I took a quick look and it looks good, thanks for submitting the PR. I'm on Christmas vacation now and will take a deeper look when I get back into the office.

Simplifying was the right call, the simpler we can make these that get developers going without learning new things the better.

For launchSettings.json are you asking when the VS toolkit writes that file to add ASPNETCORE_ENVIRONMENT = Development? If so that should be simple to do.

tonysneed:
> For launchSettings.json are you asking when the VS toolkit writes that file to add ASPNETCORE_ENVIRONMENT = Development? If so that should be simple to do.

Yes, this is what I'm looking for.

> Simplifying was the right call, the simpler we can make these that get developers going without learning new things the better.

In the spirit of simplicity, I would also suggest that we move the `AddConfiguration` extension method to [aws-dotnet-extensions-configuration](https://github.com/aws/aws-dotnet-extensions-configuration) and add a package reference to Amazon.Extensions.Configuration.SystemsManager.  We can then enclose the `ServiceCollectionExtensions` class in the `Microsoft.Extensions.Configuration` namespace. The nice side-effect of this is that devs are then in a better position to insert a call to `AddSystemsManager` and store secrets in a Parameter Store.

Lastly, I wanted to suggest, once we settle on an approach to "Empty Function with Configuration" blueprint, we do the same with "Empty Serverless Application with Configuration".

normj:
Back taking a look at this and a couple questions.

1) Unless there is a reason the Program.cs file should be removed it just adds confusion. You can add `<OutputType>Library</OutputType>` to the project file to avoid the errors about not having a `Main`.
2) The `IServiceProvider` seems like overkill to just create a `Configuration` object. Especially since the `IServiceProvider` is just a local variable. Should we just skip the `IServiceProvider` and then the extension methods and just use the `ConfigurationBuilder` in the constructor? Or am I missing something on your intention of how developers would use the `IServiceProvider`?

tonysneed:
> 1. Unless there is a reason the Program.cs file should be removed it just adds confusion. You can add <OutputType>Library</OutputType> to the project file to avoid the errors about not having a Main.

- Agreed.  Done.

> 2. The IServiceProvider seems like overkill to just create a Configuration object. Especially since the IServiceProvider is just a local variable. Should we just skip the IServiceProvider and then the extension methods and just use the ConfigurationBuilder in the constructor? Or am I missing something on your intention of how developers would use the IServiceProvider?

- The template creates an `IServiceProvider` in order to set up dependency injection and add `IConfiguration` to it, which is the same approach as the standard ASPNET.NET Core template that developers are used to.  This makes it possible for devs to inject an `IConfiguration` into other classes, as well as use DI for registering other services, such as `IRepository` interfaces for communicating with data stores. They would do this in the `ConfigureServices` method, which again is a familiar pattern from the standard template.

tonysneed:
@normj Was my explanation of the need for `IServiceProvider` satisfactory, or can I further clarify?

PR #380, opened 2019-01-17
Added additional S3 Event blueprints

austoonz:
*Issue #, if available:*
There is no issue related to this.

*Description of changes:*
Added additional PowerShell Lambda Blueprints for S3 Events. These include samples for:

- S3 -> SNS -> Lambda
- S3 -> SNS -> SQS -> Lambda
- S3 -> SQS -> Lambda

### DIFF
--- a/PowerShell/Module/Templates/Blueprints/S3EventToSNS/readme.txt
+++ b/PowerShell/Module/Templates/Blueprints/S3EventToSNS/readme.txt
@@ -1,0 +1,13 @@
+This sample creates a Lambda function written in PowerShell that can be subscribed
+to an SNS Topic. For this sample, the SNS Topic would be subscribed to S3 Events,
+for when a new object, or object version is created in a bucket.
+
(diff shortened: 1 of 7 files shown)


normj:
The design of the templates was each template would have a single Lambda function template. This PR adds 3 additional Lambda function to the same template. That would potential overall whelm a new PS Lambda user wondering what should they do with all of these files.

For each template make a separate folder and entry in the `ps-lambda-blueprint-manifest.json` file.

austoonz:
Yep I’m fine with that Norm. I was actually deciding which way to go when I was putting them together. I’ll get an update to this and the SQS one.

austoonz:
Norm - I've updated the commit in my fork with the updates as requested.

austoonz:
I modified "S3EventToSNSToSQS" to support both RAW and non-RAW SNS Subscriptions.

normj:
Merge to dev branch

PR #401, opened 2019-02-19
Including Exception and EventId in Lambda Logger messages

karpikpl:
*Issue #, if available:*
Microsoft.Extensions.Logging interface supports logging **EventId** and **Exceptions**, but when LambdaLogger is configured, those objects are not saved in CloudWatch.

Sample scenario:
```
logger.LogError(LoggingEvents.DbOperationError, ex, $"SelectData: {spName} SqlException Number: {se.Number} Sql Parameter: {spc}");
```
The natural expectation is that both Logging Event and Exception will be available in CloudWatch, which is not the case.

*Description of changes:*
I've included 2 new settings on Lambda Logger Options to enable  **EventId** and **Exceptions** logging.


normj:
Thanks for the PR. I'll take a look

normj:
Much better and just in time as I was pushing out the bits. This went out just now as version 0.9.1.

Issue #408, opened 2019-03-05
APIGatewayProxyFunction: Can't parse Identity from custom lambda authorizer output


LUS1N:
Custom Lambda Authorizer will output the claims directly in the authorizer object, while the Cognito User Pool token validator outputs the claims under authorizer.claims object, which prevents the claims from being loaded at all, cause the code in `PostCreateContext` is looking in Authorizer.Claims.

normj:
Closing in favor tracking this with your PR. https://github.com/aws/aws-lambda-dotnet/pull/409

PR #409, opened 2019-03-05
Get Idenitty claims from Authorizer if Authorizer.Claims is null or empty 

LUS1N:
*Issue #408* 

*Description of changes:*
If `Authorizer.Claims` is null or contains no claims, parse the claims from `Authorizer`, because that is how the claims are returned from the custom lambda authorizer.

Made sure the exist test passes, and made a new one with the custom lambda authorizer input.

My formatter formatted the files, I hope it's ok.


normj:
Thanks for the PR. I pushed out version 3.0.3 of Amazon.Lambda.AspNetCoreServer which has this change.

Issue #415, opened 2019-03-13
Amazon.Lambda.AspNetCoreServer 3.x Query Encoding Regression


alanedwardes:
Hi,

We are seeing a problem when upgrading to the `3.x` version of `Amazon.Lambda.AspNetCoreServer` with regards to query strings needing URL encoding _twice_. Take the following controller:

```csharp
[Route("health")]
public class HealthController : Controller
{
    public DateTimeOffset Index([FromQuery] DateTimeOffset testDateTimeOffset)
    {
        return testDateTimeOffset;
    }
}
```

Snippet from CloudFormation template for API gateway configuration:

```json
"/{proxy+}": {
    "x-amazon-apigateway-any-method": {
        "x-amazon-apigateway-integration": {
            "httpMethod": "POST",
            "type": "aws_proxy",
            "uri": { "Fn::Sub": "arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${ApiLambda.Arn}/invocations" }
        }
    }
}
```

The request is sent to API Gateway, with an Asp.NET Lambda function behind it set as a proxy resource handling all requests.

| Version | /health?testDateTimeOffset={x}            | Response                         |
|------------------|-------------------------------------------|----------------------------------|
| 3.x              | 2019-03-12T16%3A06%3A06.5498170%2B00%3A00 | 0001-01-01T00:00:00+00:00        |
| 2.1.x            | 2019-03-12T16%3A06%3A06.5498170%2B00%3A00 | 2019-03-12T16:06:06.549817+00:00 |

If we double encode the query parameter as `2019-03-12T16%253A06%253A06.5498170%252B00%253A00` with the `3.x` package, the date time is as parsed correctly: `2019-03-12T16:06:06.549817+00:00`.

This seems like incorrect behavior - single encoding should be enough as it was with `2.1.x`.

klaytaybai:
Hi @alanedwardes, thank you for reporting this.  We have had some similar bug reports; this may be a duplicate.  I'll work on getting this prioritized highly.

Rutgerz:
I haven't had time to make a demo for this issue but we encountered the same problem. We had IdentityServer running in lambda (move to elb now due to performance) and sign-in uses a number of redirects. These redirects broke because of the extra url decoding that is done in v3.x.

normj:
I have a repo case and I'm investigating how to fix the issue.

alanedwardes:
[...]

PR #451, opened 2019-04-27
Fixed issue with not handling query string encoding

normj:
This fixed an issue where query string values were not correctly being encoded for requests coming in from API Gateway to ASP.NET Core Lambda functions. This address the following issue https://github.com/aws/aws-lambda-dotnet/issues/415

This behavior was triggered when Application Load Balancer support was added. API Gateway will decode the values before passing them to the Lambda  function and ALB does not decode. There was a mistake when ALB was added thinking both services send decoded values.

The fix is to have API Gateway request UrlEncode values before adding them to the ASP.NET Core request. ASP.NET Core will later UrlDecode the values.

Tests were done by first manually verifying what API Gateway and ALB pass into Lambda functions and then updating the encoding JSON test documents for the unit tests.


normj:
Corrected description of the problem and fixed the typo in all of the test files.

Issue #458, opened 2019-05-12
Make library write out a warning when invoked with the wrong payload from upstream


hvanbakel:
It would be really helpful to log a warning or something similar when an endpoint is used that derives from the `ApplicationLoadBalancerFunction` but is invoked with an Api Gateway payload or vice versa. 

Especially when switching between the 2 this can catch otherwise hard to find issues because just not all properties get populated.

klaytaybai:
@hvanbakel, thanks for the contribution.  I think we're going to need to discuss the best options for handling this before committing to a merge.  For example, instead of logging a warning, we may consider adding a method to verify the request type.

normj:
Closing this as the work is being tracked in the PR https://github.com/aws/aws-lambda-dotnet/pull/459

PR #459, opened 2019-05-12
Log warning if integration doesn't match base class

hvanbakel:
Fix for #458 

Log a warning if we suspect there to be a mismatch between the integration used and the class that the function inherits from.


normj:
Thanks for the PR. I have merge the change to the `dev` branch to go out with the next release.

normj:
Version 3.1.0 of Amazon.Lambda.AspNetCoreServer has been released with this PR. Thanks again for the PR.

PR #471, opened 2019-05-30
Support for Scopes in Logger

karpikpl:
*Issue #, if available:* outdated readme

*Description of changes:* Updated dependencies and added support for scopes

I think this is a super important feature for lambda logger.
Logs without scope are pretty much useless if lambda is getting traffic.

After this change you can do:
```
                using(defaultLogger.BeginScope(awsRequestId))
                {
                    defaultLogger.LogInformation("Hello");

                    using(defaultLogger.BeginScope("Second {0}", "scope456"))
                    {
                        defaultLogger.LogError("In 2nd scope");
                        defaultLogger.LogInformation("that's enough");
                    }
                }
```
and all your logs will have `awsRequestId` in them. It makes it easy to find/filter and read logs in cloudwatch.


karpikpl:
I would appreciate if someone help me run tests - they are failing for me on master

Libraries/src/Amazon.Lambda.Logging.AspNetCore/LambdaILogger.cs
```diff
@@ -9,29 +10,26 @@ internal class LambdaILogger : ILogger
...
         }
 
         // ILogger methods
-        public IDisposable BeginScope<TState>(TState state)
-        {
-            // No support for scopes at this point
-            // https://docs.asp.net/en/latest/fundamentals/logging.html#scopes
-            return new NoOpDisposable();
-        }
+        public IDisposable BeginScope<TState>(TState state) => ScopeProvider?.Push(state) ?? NullScope.Instance;
```
normj: Instead of using NullScope from ASP.NET Core internal namespace we should continue to use the existing NoOpDisposable object

normj:
Thanks for the PR. I'm taking a look and testing in practice. Looks like a good change.

Libraries/src/Amazon.Lambda.Logging.AspNetCore/LambdaILogger.cs
```diff
@@ -45,30 +43,33 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
-            //  {[LogLevel] }{Category: }{EventId: }MessageText {Exception}{\n}
+            //  {[LogLevel] }{ => Scopes : }{Category: }{EventId: }MessageText {Exception}{\n}
 
             var components = new List<string>(4);
             if (_options.IncludeLogLevel)
             {
                 components.Add($"[{logLevel}]");
             }
+
+            GetScopeInformation(components, multiLine: _options.IncludeNewline);
```
normj: I'm seeing scopes always being included regardless of what `IncludeScopes` is set to. Seems like we need an if check before calling GetScopeInformation.

Libraries/src/Amazon.Lambda.Logging.AspNetCore/LambdaILogger.cs
karpikpl: ok. I'll bring back NoOpDisposable

Libraries/src/Amazon.Lambda.Logging.AspNetCore/LambdaILogger.cs
karpikpl: sorry, I've missed it.
I based my implementation on https://github.com/aspnet/Extensions/blob/642b9263abe16ba1cd038f27c8cb4661763ff5bb/src/Logging/Logging.Console/src/ConsoleLogger.cs#L293 but missed that condition

karpikpl: it would not log scopes even without the check, since Scopes would be set to `NullExternalScopeProvider`
I think it's more readable with the check like you suggested

normj:
PR is merged into the dev branch. I'll update and close the PR once the feature has been merged to master and released.

normj:
Version 2.3.0 has been released with this PR. Thanks again for the PR.

PR #474, opened 2019-06-05
Adding new identity middleware project for ALB OpenId Connect authentication

thedevopsmachine:
Application Load Balancers are able to automatically perform OpenId Connect authentication and pass the validated tokens in to the application as headers.

This commit contains a new middleware project (with tests) for ASP.NET Core applications running behind an Application Load Balancer (in Lambda, Kestrel, or IIS) that is configured to use an OpenId Connect identity provider.

The middleware will parse the identity headers injected by ALB, optionally validates the token's signature, and transforms the token into a ClaimsPrincipal that can be accessed in the RequestContent's "User" property.


Libraries/src/Amazon.Lambda.ApplicationLoadBalancerIdentity/ALBIdentityMiddleware.cs
```diff
@@ -0,0 +1,245 @@
...
+        /// Caching related properties
+        private static uint[] byteLookup;
+        private readonly bool cacheIdentities;
+        private readonly MemoryCache cachedIds;
+        private readonly SHA1 sha;
+        private readonly TimeSpan cacheDuration;
+
+        // Validation related properties
+        private readonly string region;
+        private readonly ConcurrentDictionary<string, Task<TokenValidationParameters>> validationParameters;
```
normj: If the `Task` fails for an intermittent error the dictionary will always have the failed task in the dictionary. I would prefer only the TokenValidationParameters is in the dictionary and not the Task.

Libraries/src/Amazon.Lambda.ApplicationLoadBalancerIdentity/ALBIdentityMiddleware.cs
```diff
@@ -0,0 +1,245 @@
...
+                // TimeSpan.MaxValue isn't a real thing. If the underlying library that uses it converts it to an integer (which a lot do),
+                // then it will throw an ArgumentOutOfRange exception if you specify a timespan with more seconds in it than
+                // the max value of an Int32 has. To use an *actual* MaxValue that doesn't fail, you need to create a timespan from
+                // the maximum int size in seconds.
+                var maxCacheLife = TimeSpan.FromSeconds(int.MaxValue);
+
+                this.cachedIds = new MemoryCache(new MemoryCacheOptions
+                {
+                    // Evict 20% of cache entries on max size.
+                    CompactionPercentage = 0.2,
```
normj: Should this be configurable from the options object?

Libraries/src/Amazon.Lambda.ApplicationLoadBalancerIdentity/ALBIdentityMiddleware.cs
```diff
@@ -0,0 +1,245 @@
...
+            await this.next.Invoke(httpContext);
+        }
+
+        private async Task<ClaimsPrincipal> GetUser(JsonWebToken jwt, ICacheEntry cacheEntry)
+        {
+            this.logger?.LogDebug("User Principal is '{0}' (issued by '{1}')", jwt.Subject, jwt.Issuer);
+
+            if (this.options.VerifyTokenSignature)
+            {
+                var validationParameters = await this.validationParameters.GetOrAdd(jwt.Kid, async (key) =>
```
normj: Local variables and member variables having the same name cause too much confusion. Please use different names.

Libraries/src/Amazon.Lambda.ApplicationLoadBalancerIdentity/ALBIdentityMiddlewareOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// </summary>
+        public ushort? MaxCacheSizeMB { get; set; }
+
+        /// <summary>
+        /// Gets or sets the maximum amount of time an identity will remain in the in-memory cache.
+        /// </summary>
+        public TimeSpan? MaxCacheLifetime { get; set; }
+
+        /// <summary>
+        /// Gets or sets a flag indicating whether tokens should be verified.
```
normj: Can we add some text about what verifying the token means? Should the default be true to avoid surprising users that think we would be verifying by default?

Libraries/src/Amazon.Lambda.ApplicationLoadBalancerIdentity/README.md
```diff
@@ -0,0 +1,85 @@
+# Amazon.Lambda.ApplicationLoadBalancerIdentity
+
+This package contains Middleware that can be used in conjunction with ALB's OpenId Connect integration to populate the "User" property in the RequestContext.
+
+While it is in the Amazon.Lambda namespace, it can also be used in other ASP.NET Core deployment scenarios e.g. Fargate or EC2.
```
normj: Should we put this in its own repo so isn't so tied to Lambda. Fargate seems like an excellent use case and could be more common then Lambda.

Libraries/src/Amazon.Lambda.ApplicationLoadBalancerIdentity/ALBIdentityMiddleware.cs
thedevopsmachine: Good catch, I will implement your recommendation.

Libraries/src/Amazon.Lambda.ApplicationLoadBalancerIdentity/ALBIdentityMiddleware.cs
thedevopsmachine: Seems reasonable.

Libraries/src/Amazon.Lambda.ApplicationLoadBalancerIdentity/ALBIdentityMiddleware.cs
thedevopsmachine: Will do

Libraries/src/Amazon.Lambda.ApplicationLoadBalancerIdentity/ALBIdentityMiddlewareOptions.cs
thedevopsmachine: The reason I picked the default as false is that caching is not enabled by default, and enabling validation without caching can be expensive. However, that could easily be explained in the documentation, so I'll do that. I'll also change the default behavior to verify tokens, and enable caching with a moderate cache size limit, to enable a higher security bar.

Libraries/src/Amazon.Lambda.ApplicationLoadBalancerIdentity/README.md
thedevopsmachine: I'd be happy to do that, which repo should it move to?

Libraries/src/Amazon.Lambda.ApplicationLoadBalancerIdentity/Amazon.Lambda.ApplicationLoadBalancerIdentity.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
costleya: Obviously, this will change if we put it in a separate repo and emphasize it is not Lambda specific.

normj:
Closing this PR as it the work is being tracked in a new repo.

Kralizek:
Link to the repo? Or at least the name? 😅

PR #486, opened 2019-07-08
Changed APIGatewayProxyFunction from LambdaContextLogger to use ILogger instead, in order to reduce noise in CloudWatch

CasperSkydt:
*Issue #, if available:*

*Description of changes:*
Remove repeated log statements to CloudWatch.

The following is printed to CloudWatch for each request.

"Configuring HttpContext.User with 2 claims coming from API Gateway's Request Context"

Why not use the  microsoft  _logger instead...


normj:
Thanks for the PR. I have merge the change into the upcoming `aspnetcore30-support` branch which has some slight breaking changes and will be  major version bump. The an change on this branch is to add .NET Core 3.0 support for those using custom runtimes.

PR #496, opened 2019-07-25
Split out SimpleEmailEventsReceiptAction into different action types

craigbrett17:
*Description of changes:*

We were trying to use this library to allow us to have a lambda handle emails being received with an S3 action. However, the library as was only supports Lambda. So we're currently using dynamic to get around this limitation, but I'd rather be type safe.

So, what I've done, is split the SimpleEmailReceiptAction into an interface and 2 different action types. This type can then be passed in as a `<T>` at any level of the nested classes, and will resolve properly. Unit tests confirm.

It would help us be type safe were this request to be confirmed.


craigbrett17:
@normj: Alright, I've gone with what you suggested and implemented some lambda by default backwards compatible types. Hopefully the obsolete message will be enough to encourage people to make the hopefully painless change over. 🤞

normj:
I have merged this into the dev branch to go out with the next release.

normj:
Version 2.0.0 of Amazon.Lambda.SimpleEmailEvents  has been released with this PR

Issue #500, opened 2019-08-02
Changing the serializer settings used by the lambda JsonSerializer


mabead:
As documented [here](https://docs.aws.amazon.com/lambda/latest/dg/dotnet-programming-model-handler-types.html#lambda-dotnet-add-serializer), the `Amazon.Lambda.Serialization.Json.JsonSerializer` is based on Newtonsoft's JSON.Net. To define our serialization settings, we typically use the procedure described [here](https://www.newtonsoft.com/json/help/html/DefaultSettings.htm). For example, we often have something in our programs initialization code to make sure that the `enum` are serialized as strings and not as integers:

```csharp
JsonConvert.DefaultSettings = () => new JsonSerializerSettings
{
    Converters = new List<JsonConverter>
    {
        new StringEnumConverter() { CamelCaseText = false }
    }
};
```
Unfortunately, it looks like the JsonSerializer provided by AWS does not use the default conversion settings. For example, let's consider the following code:

```csharp
using System;
using System.Collections.Generic;
using System.Linq;
using System.Threading.Tasks;
using Amazon.Lambda.Core;
using Newtonsoft.Json;
using Newtonsoft.Json.Converters;

[assembly: LambdaSerializer(typeof(Amazon.Lambda.Serialization.Json.JsonSerializer))]

namespace TestToolTrial
{
    public enum Direction
    {
        Up,
        Down,
    }

    public class SomeResult
    {
        public string Text { get; set; }
        public Direction Direction { get; set; }
    }


    public class Function
    {
        public Function()
        {
            JsonConvert.DefaultSettings = () => new JsonSerializerSettings
            {
                Converters = new List<JsonConverter>
                {
                    new StringEnumConverter() { CamelCaseText = false }
                }
            };
        }

        public SomeResult FunctionHandler(ILambdaContext context)
        {
            var result = new SomeResult
            {
                Text = "Hello",
                Direction = Direction.Down,
            };

            var resultAsString = JsonConvert.SerializeObject(result);
            Console.WriteLine("Result as string is : " + resultAsString);

            return result;
        }
    }
}
```
When I deploy this lambda and invoke it from the AWS console (or if I use it with the Lambda Test Tool), I see that the lambda returns this:
```json
{
  "Text": "Hello",
  "Direction": 1
}
```
As you can see, my `Direction` enum is not serialized as a string. 
[...]

PR #503, opened 2019-08-12
add constructor that allows changing the serializer settings

mabead:
*Issue #500*

*Description of changes:*

As described in related issue, I add a constructor that makes it easy to customize the lambda json serialization settings. 

Here's an example of how the json serialization settings can be changed by using a derived class:

```csharp
public class MySerializer : JsonSerializer
{
    public MySerializer()
        : base(CustomizeSerializerSettings)
    {
    }

    private static void CustomizeSerializerSettings(JsonSerializerSettings serializerSettings)
    {
        serializerSettings.Converters = new List<JsonConverter>
        {
            new StringEnumConverter() { CamelCaseText = false }
        };
    }
}
```

The new serializer could then be used like this:
```csharp
[assembly: LambdaSerializer(typeof(MySerializer))]
```

### DIFF
--- a/Libraries/src/Amazon.Lambda.Serialization.Json/JsonSerializer.cs
+++ b/Libraries/src/Amazon.Lambda.Serialization.Json/JsonSerializer.cs
@@ -22,18 +22,33 @@ namespace Amazon.Lambda.Serialization.Json
         private bool debug;
 
         /// <summary>
-        /// Constructs instance of serializer.
+        /// Constructs instance of serializer. This constructor is usefull to 
+        /// customize the serializer settings.
         /// </summary>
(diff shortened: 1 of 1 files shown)


normj:
Version 1.6.0 of Amazon.Lambda.Serialization.Json is out with the PR. Thanks for the PR!

UrmaGurd:
Is there an example of how to use this to deserialize strings to enums with Lambda functions?

My lambda function is currently using 

`[assembly: LambdaSerializer(typeof(Amazon.Lambda.Serialization.SystemTextJson.DefaultLambdaJsonSerializer))]`

How do I change this to support enums / string conversions?

PR #509, opened 2019-08-20
Added support for netstandard2.0 and .NET Core 2.2

aaltotsky:
Added validation for the netstandard during the config file search.  Switch to netcoreapp2.2 to avoid "System.IO.FileLoadException: Could not load file or assembly 'Microsoft.Extensions.DependencyInjection, Version=2.2.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. Could not find or load a specific file." error.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Tools/LambdaTestTool/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
+++ b/Tools/LambdaTestTool/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
@@ -4,7 +4,7 @@
   
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
     <ToolCommandName>dotnet-lambda-test-tool-2.1</ToolCommandName>
     <IsPackable>true</IsPackable>
     <PackAsTool>true</PackAsTool>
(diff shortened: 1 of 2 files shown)


wellrightamplify:
any update on this pr

devolderg:
Any update here?

normj:
I'm assuming you were developing a lambda function that was using the version Microsoft.Extensions.DependencyInjection that came out with .NET Core 2.2. That is valid to do since that is a .NET Standard 2.0 library not depending on anything special of .NET Core 2.2. You had a conflict in this tool because it is an ASP.NET Core app which loads the version of Microsoft.Extensions.DependencyInjection that came out with .NET Core 2.1. 

That definitely is a limitation of this tool but I don't want to change the target framework of this tool to be 2.2. That would work around your problem but that also makes this tool use a lot more code that doesn't match what is running in Lambda. That is why I kept this tool very restricted on its dependencies so it was only using the assemblies that already in the Lambda environment.

If we look in the future and make the assumption, (i.e. I'm not making a product announcement),  .NET Core 3.1 comes out which is the next LTS and we provide a new Lambda runtime. We could rework this to use some of the newer features added for AssemblyLoadContext to make the Lambda code load in the tool have a separate dependency graph and then there wouldn't be these types of conflicts.

Also what was the reason for adding netstandard2.0 when reading the config file since a Lambda function has to target netcoreappX?

@wellrightamplify and @devolderg what were your use cases for this PR?

normj:
Closing for lack activity and this isn't a change we want to take.

PR #518, opened 2019-09-02
Contract resolvers are stateless. Reuse them

danielmarbach:
ContractResolvers are stateless and can be reused. 

Decided to use the AwsResolver as the root object for the converters and use a simple pattern to initialize the fields once


normj:
Version 1.7.0 of Amazon.Lambda.Serialization.Json has been released with this PR.

danielmarbach:
Nice. Less allocations in AWS makes us all happy campers 😁👍

Issue #519, opened 2019-09-02
Logging LogLevel should be handled as namespace prefix


wdolek:
It is common to use typed `ILogger<T>` in .NET - this will produce category which equals to full type name, so for example `ILogger<HttpClient>` has category `System.Net.Http.HttpClient`.

Usually, this should be enough to have `Debug` level for `HttpClient`:

```javascript
    "LogLevel": {
      "Microsoft": "Information",
      "System": "Warning",
      "System.Net": "Information",
      "System.Net.Http": "Debug"
    }
```

Yet in order to make this work, wildcards must be used like this:

```javascript
    "LogLevel": {
      "Microsoft.*": "Information",
      "System.*": "Warning",
      "System.Net.*": "Information",
      "System.Net.Http.*": "Debug"
    }
```

I spent quite a lot of time debugging my configuration not realizing this. In addition to wildcards, I propose adding simple support of namespace prefixes as well, which is common in .NET.

Framework documentation: [Logging in .NET Core and ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-2.1).

normj:
PR merged and released

PR #520, opened 2019-09-02
#519 Match type namespace prefix as well when building `LogLevels`

wdolek:
Related to #519 

*Description of changes:*
Adding support of namespace prefixes as well, alongside of wildcard support. Namespace prefixing is common in .NET, see issue.

With this change, log levels can be configured like this:

```javascript
    "LogLevel": {
      "Default": "Debug",
      "Microsoft": "Information",
      "System": "Warning",
      "System.Net": "Information",
      "System.Net.Http": "Debug"
    }
```

Wildcards are still supported and evaluated _before_ namespace prefixes.

### DIFF
--- a/Libraries/src/Amazon.Lambda.Logging.AspNetCore/LambdaLoggerOptions.cs
+++ b/Libraries/src/Amazon.Lambda.Logging.AspNetCore/LambdaLoggerOptions.cs
@@ -206,6 +206,7 @@ namespace Microsoft.Extensions.Logging
                     }
 
                     var wildcardLocation = category.IndexOf('*');
+
                     // Wildcards are only supported at the end of a Category name!
                     if (wildcardLocation != category.Length - 1)
                     {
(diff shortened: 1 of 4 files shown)


Libraries/src/Amazon.Lambda.Logging.AspNetCore/LambdaLoggerOptions.cs
```diff
@@ -221,9 +222,18 @@ private static bool TryGetSection(IConfiguration configuration, string key, out
                 }
             }
 
-            // Extract a reverse sorted list of wildcard categories.  This allows us to quickly search for most specific match
+            // Extract reverse sorted list of given categories. This will allow us to quickly search for most specific namespace prefix match
+            // to least specific - notice we add "." (namespace delimiter) at the end, assuming it is namespace prefix.
+            var namespacePrefixCategories = logLevelsMapping.Keys
```
normj: Seems like with this change we are treating every namespace as a wild card if there is not exact match. Which makes sense to me but instead of building 2 separate collection which lives the awkward question which collection takes priority should we just build one sorted collection?

wdolek: @normj good point, I was too afraid to touch existing code, I will prepare change.

normj:
Version 3.0.0 of Amazon.Lambda.Logging.AspNetCore has been released with the PR.

Issue #521, opened 2019-09-02
"Default" category's log level shoud be used when category does not match


wdolek:
## Description

Current AWS Lambda library logging differs from usual .NET Core logging implementation. Difference is in how `Default` log level category is handled.

In .NET Core:

- `Default` category is used for any logger, which name doesn't match other declared categories
- if `Default` is not set and category does not match, minimum level is `Information`

You can read more about logger configuration here: [Logging in .NET Core and ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-2.1).

### Use "`Default`" category if no match

1. Use and load following configuration:

```json
{
  "Lambda.Logging": {
    "LogLevel": {
      "Default": "Warning",
      "System": "Warning"
    }
  }
}
```

2. Using `ILoggerFactory`, create logger with name `Dummy` (not matching `System`)
3. Try to log debug event

#### Expected result
- No message logged, default logging (for categories not matching) is set to `Warning`

#### Actual result
- Message logged - no category match -> any level is enabled: [LambdaLoggerOptions.cs:L250](../blob/master/Libraries/src/Amazon.Lambda.Logging.AspNetCore/LambdaLoggerOptions.cs#L250)

### Using minimum log level `Information` when "`Default`" category not set

1. Use and load following configuration

```json
{
  "Lambda.Logging": {
    "LogLevel": {
      "System": "Warning"
    }
  }
}
```

2. Using `ILoggerFactory`, create logger with name `Dummy` (not matching `System`)
3. Try to log debug event

#### Expected result
- No message logged, default logging (in code) is set to `Information` (If config is missing, we don't want _slow down_ app with trace and debug events logging)

#### Actual result
- Message logged - no category match -> any level is enabled: [LambdaLoggerOptions.cs:L250](../blob/master/Libraries/src/Amazon.Lambda.Logging.AspNetCore/LambdaLoggerOptions.cs#L250)

## Conclusion

This is **breaking** change of course, however it will make life easier by adjusting to standard .NET behavior. (Yes, I spent some time struggling with using "my normal logging config" in Lambda)

madmox:
After reviewing the source code, I suspect the lambda logger implementation is not standard-compliant in multiple ways, one of which explains the observed issue:

**Lambda logger LogLevel categories**
[...]

PR #522, opened 2019-09-02
#521 Adjust handling of `Default` log category, adjusting it to .NET

wdolek:
## Adjust handling of `Default` log category, adjusting it to .NET Core behavior
Related to #521 

*Description of changes:*

**Breaking change**!

- when `Default` logging category is set, it is used for loggers which don't match any configured category
- when `Default` logging category is not set, `Information` is minimum level for categories not matching any config key

See #521 and/or unit tests for more details.


normj:
Version 3.0.0 of Amazon.Lambda.Logging.AspNetCore has been released with the PR.

PR #567, opened 2020-01-16
Support netcoreapp3.1 in Amazon.Lambda.AspNetCoreServer.

mukhtar:
*Description of changes:*
This change adds support for ASP.NET Core 3.1 to the Amazon.Lambda.AspNetCoreServer Library. From the commit history, f87adb7 change seems to have added netcoreapp3.0 support, this change just extends that.


normj:
Since we are not using any new features there is no reason to add a new another target framework to compile for. When a user adds the NuGet package to a ASP.NET Core 3.1 project NuGet will resolve the version to use the .NET Core 3.0 version.

mukhtar:
Thank you for the quick response. Closing.

Issue #544, opened 2019-11-14
System.Text.Json Support for AWS Lambda SDK?


martincostello:
As part of migrating some existing .NET Core 3.0 worker services running on EC2 to AWS Lambda, we've coded-up a custom `ILambdaSerializer` implementation to handle deserializing payloads to the Lambda function, such as `SQSEvent`.

While this is relatively simple to do, it would be nice if a serializer implementation for System.Text.Json was built into the Lambda SDK to save the need for copy-paste across different function codebases.

One shortcoming with the state-of-things as of today is that `PropertyNameCaseInsensitive` has to be set to `true` on the serializer options as the events don't use consistent PascalCase vs. camelCase naming, which incurs a performance penalty (I don't have a quantifier for how much the penalty is). In the case of `SQSEvent` I believe it's `eventSourceARN` that trips it up as it doesn't map over to `EventSourceArn`.

I imagine this could be easily fixed in the event libraries by adding `[JsonPropertyName("...")]` attribute annotations to the properties on the models.

I believe further benefit could be gained if there was asynchronous methods on the interface to leverage the async support of `System.Text.Json.JsonSerializer` and consume the `Stream` directly, but that's a bigger change that just providing an implementation based on the current interface.

Below is the code we're using. It might benefit from further optimization, but as it is right now it's working well for us:

```csharp
using System.IO;
using System.Text.Json;
using Amazon.Lambda.Core;

namespace MyLambdaFunction
{
    public sealed class SystemTextJsonLambdaSerializer : ILambdaSerializer
    {
        private readonly JsonSerializerOptions _options;

        public SystemTextJsonLambdaSerializer()
        {
            // SQSEvent does not use a consistent camel/Pascal case naming convention,
            // so as we cannot annotate with attributes, the options need to be case
            // insensitive to be tolerant for things to work properly when deserializing.
            _options = new JsonSerializerOptions()
            {
                IgnoreNullValues = true,
                PropertyNameCaseInsensitive = true,
            };
        }

        public T Deserialize<T>(Stream requestStream)
        {
            using var copy = new MemoryStream();
            requestStream.CopyTo(copy);

            byte[] utf8Json = copy.ToArray();

            return JsonSerializer.Deserialize<T>(utf8Json, _options);
        }

        public void Serialize<T>(T response, Stream responseStream)
        {
            using var writer = new Utf8JsonWriter(responseStream);
            JsonSerializer.Serialize(writer, response, _options);
        }
    }
}
```

[...]

PR #568, opened 2020-01-18
[NETCore31] Add support for using System.Text.Json as the serialization parser

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/544

*Description of changes:*
.NET Core 3.0 added a new JSON serialization called System.Text.Json. This PR creates new Lambda serialization called Amazon.Lambda.Serialization.SystemTextJson that will be based on System.Text.Json. Unlike Newtonsoft's JSON parser System.Text.Json is built into the framework and is optimized for speed. 
 

This will released as part of the upcoming Lambda .NET Core 3.1 and will only target .NET Core 3.1. There will not be a .NET Standard 2.0 which would make it available to .NET Core 2.1 Lambda functions because the event packages like Amazon.Lambda.S3 would have to take a dependency on the System.Text.Json Nuget package.

In my anecdotal testing I have seen some good cold start performance improvements using this library. For those that are currently trying .NET Core 3.1 on Lambda using the custom runtime feature here are preview versions of the impacted NuGet packages to try out.

https://normj-packages.s3.us-west-2.amazonaws.com/System.Text.Json-2020-01-17.zip


msimpsonnz:
There is a .NET Standard package for this which makes it available to 2.1
Would be nice if the SDK’s could align
https://www.nuget.org/packages/System.Text.Json

Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/AwsNamingPolicy.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text.Json;
+
+namespace Amazon.Lambda.Serialization.SystemTextJson
+{
+    /// <summary>
+    /// Custom AWS naming policy
+    /// </summary>
+    public class AwsNamingPolicy : JsonNamingPolicy
```
costleya: Just brainstorming...

Could we make a custom attribute that stores the same string that JsonPropertyNameAttribute does, and read from that attribute in this class? That way, we don't need all of the macros. Optionally, we could have the string be null by default, in which case this class could read from the DataRow attribute to reduce code duplication.

Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/Converters/DateTimeConverter.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Text.Json.Serialization;
+
+namespace Amazon.Lambda.Serialization.SystemTextJson.Converters
+{
+    /// <summary>
+    /// DateTime converter that handles the JSON read for deserialization might use an epoch time.
+    /// </summary>
+    public class DateTimeConverter : JsonConverter<DateTime>
+    {
+        private static readonly DateTime EPOCH_DATETIME = new DateTime(1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc);
```
martincostello: You can use `DateTime.UnixEpoch` instead.

Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/LambdaJsonSerializer.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        /// <summary>
+        /// Constructs instance of serializer.
+        /// </summary>        
+        public LambdaJsonSerializer()
+        {
+            _options = new JsonSerializerOptions()
+            {
+                IgnoreNullValues = true,
+                PropertyNameCaseInsensitive = true,
```
martincostello: Are there more annotations that could be added that would make it possible to remove this? It comes with a performance penalty to some degree, so would be better if it could be left off.

Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/Converters/DateTimeConverter.cs
normj: Nice, I didn't realize `DateTime.UnixEpoch` was added.

normj:
@msimpsonnz That was my original attention to use the NuGet package for .NET Core 2.1. The problem is the AWS event packages require customization of the .NET Property name to JSON Property name mapping. For Newtonsoft we could add `System.Runtime.Serialization.DataMember` attributes to the event packages like Amazon.Lambda.CloudWatchLogsEvents to handle the customization. For System.Text.Json it doesn't recognize `System.Runtime.Serialization.DataMember` attributes and instead you have to use `System.Text.Json.Serialization.JsonPropertyName`. That means in order to support System.Text.Json for .NET Core 2.1 Lambda functions the event packages would have to take a NuGet package dependency on a System.Text.Json. I don't want to force anybody using .NET Core 2.1 to have to include any other NuGet package especially if they are not using it because they are using the Newtonsoft serializer.

Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/LambdaJsonSerializer.cs
normj: I was thinking users could use the constructor that takes in `Action<JsonSerializerOptions> customizer` to customize or remove these settings. I'm open to suggestions to make it easier.

Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/AwsNamingPolicy.cs
costleya: - Considering that JsonNamingPolicy is limited (you only get a string, no additional context), the solution I was suggesting could be dangerous.
- Using "DataMember" functionality dynamically is complex and messy.
- The tracking issue for having the Json serializer respect the DataMember attribute is now in the 5.0 milestone tracker https://github.com/dotnet/runtime/issues/29975

In light of these facts, this is not a viable suggestion. Approving.

PR #581, opened 2020-02-12
fix null ref in APIGatewayProxyFunction

bo-bac:
*Issue #, if available:*

*Description of changes:*
it comes if apiGatewayRequest.Path is null


bo-bac:
hm.. perdon it must be true.
but in my case apiGatewayRequest.Path was null (I use socket gateway and try to join .core infrastructure to handle socket communication, use auth and so on..) in WebSocket.Connected.Lambda I had the issue.
Thus it might be

wdolek:
Would it be possible to add test for your case? (I _think_ it should be possible, you can get inspiration from `Libraries/test/Amazon.Lambda.AspNetCoreServer.Test/TestApiGatewayHttpApiV2Calls.cs`) Also could you share stack trace so it is possible to inspect proper place where NRE is coming from?

PR #595, opened 2020-03-09
core 3.1 support

tmcFoxen:
Issue #554

* Updated .csproj files to 3.1
* Removed unnecessary package reference to `Microsoft.AspNetCore.All`
* Tweaked streams code to be async due to an error at runtime calling synchronous ReadXXX methods


normj:
Thanks for the PR. I'm assuming since you closed the issue you saw the work is already in process in the [mock-testtool-31](https://github.com/aws/aws-lambda-dotnet/tree/mock-testtool-31) branch.

The work is essentially to create 2 separate tools one for 2.1 and one for 3.1 and move a lot of the code into shared project.  There is also work to load the Lambda code in a separate AssemblyLoadContext to avoid some of the assembly loading issues people are having with the current version.

PR #597, opened 2020-03-11
.NET Core 3.1 support for Lambda Test Tool

normj:
Add support for .NET Core 3.1 with the .NET Lambda Test Tool. This PR looks much larger then it really is because of massive file rearrangements.

The Lambda Test Tool must match the version of .NET Core it is being tested. Currently we ship the NuGet package **Amazon.Lambda.TestTool-2.1** which has a cli command called **dotnet-lambda-test-tool-2.1**. With the 3.1 support we will ship an additional tool called **Amazon.Lambda.TestTool-3.1** containing **dotnet-lambda-test-tool-3.1**.

The project was refactored to put all of the common logic about selected Lambda function into the **Amazon.Lambda.TestTool** project. Then separate ASP.NET Core applications for 2.1 and 3.1 were created. The 2.1 is the same ASP.NET Core that we currently ship for 2.1 support. The 3.1 is a new application using Blazor Server Side technology.

image

Besides supporting .NET Core 3.1 the new features/fixes added are:
* The selected Lambda code is now loaded in a separate AssemblyLoadContext to avoid assembly load collisions that current users are having issues. This logic can be found at `Amazon.Lambda.TestTool.Runtime.LambdaAssemblyLoadContext`.
* The test tool can now be run with the `--no-ui` to avoid using the web interface to select the code and use command line arguments to identify the function. This logic can be found at `Amazon.Lambda.TestTool.TestToolStartup.ExecuteWithNoUi()`.
* A new documentation page was added to the 3.1 version.
* The settings folder `.lambda-test-tool` will now only be created during write operations.

The tests were broken out to separate projects for 2.1 and 3.1 and used a shared source project. This was done instead of multi targeting because mac builds wouldn't work with the conditional project reference includes.

Areas to focus for PR:
* All of the **Amazon.Lambda.TestTool.WebTester31** project.
* The **TestToolStartup** class which contains the logic for the new --no-ui mode.
* The **LambdaAssemblyLoadContext** class containing the custom AssemblyLoadContext


normj:
This PR is too large and unwieldy to work with. Closing this and will come up with a different way to do this PR.

PR #610, opened 2020-04-01
Fix .NET Core version in 3.1 examples

eduherminio:
* Fix .NET Core version in 3.1 examples


normj:
Thanks for the fix. I'll get that deployed out.

PR #611, opened 2020-04-01
Fix RemainingTime returning negative values

martincostello:
Fix `LambdaContext.RemainingTime` returning negative values in a local time zone east of UTC due to local time being used for Unix epoch and clock.

This was working as expected for the last few months in some tests we had that run on our local developer machines, but since the UK moved to British Summer Time (UTC+1), the `RemainingTime` property is now returning a value approximately equal to -1 hour.

This PR fixes that by explicitly specifying that the Unix Epoch is in UTC, rather than the default of `DateTimeKind.Unspecified`, and then uses `UtcNow` instead of `Now`.


normj:
Thanks for the fix. I'll get this released soon.

normj:
Version 1.1.1 of Amazon.Lambda.RuntimeSupport is out with this fix. Thanks again.

martincostello:
Thanks @normj!

PR #676, opened 2020-05-26
Fix definition of details for ECSTaskStateChangeEvent.

### LINKED ISSUE #675: ECSTaskStateChangeEvent cannot be deserialized properly
`ECSTaskStateChangeEvent` cannot be deserialized properly by lambda due to differences in the definition of `Attachment.Details`.  `Attachment.Details` are returned as a List of `{"name" : "name", "value" : "value}` pairs, not a List of `KeyValuePair<string, string>`

This is the exception:

```
Error converting the Lambda event JSON payload to type Amazon.Lambda.CloudWatchEvents.ECSEvents.ECSTaskStateChangeEvent: The JSON value could not be converted to System.Collections.Generic.List`1[System.Collections.Generic.KeyValuePair`2[System.String,System.String]]. Path: $.detail.attachments[0].details[0] | LineNumber: 0 | BytePositionInLine: 412.: JsonSerializerException
   at Amazon.Lambda.Serialization.SystemTextJson.DefaultLambdaJsonSerializer.Deserialize[T](Stream requestStream)
   at lambda_method(Closure , Stream , Stream , LambdaContextInternal )
```

stang-tgs:
Fixes #675

The code is missing all the comments that are available everywhere else, and the extra class is not in its own file, but this is to illustrate a fix for this issue.

*Description of changes:*
Below is a snippet of an `Attachment` that is returned by from an ECSTaskStateChangeEvent Cloudwatch Event.  Notice that the Attachment Details is a list of Name/Value pairs (Not KeyValue) 
``` json
            {
                "id": "a7f304c8-ab44-4c9a-939e-e60e19600d48",
                "type": "eni",
                "status": "DELETED",
                "details": [
                    {
                        "name": "subnetId",
                        "value": "subnet-0a4e7c6ae27b807d5"
                    },
                    {
                        "name": "networkInterfaceId",
                        "value": "eni-0097b893e94948589"
                    },
                    {
                        "name": "macAddress",
                        "value": "12:03:1c:c1:01:3f"
                    },
                    {
                        "name": "privateIPv4Address",
                        "value": "10.1.21.156"
                    }
                ]
            }
```


stang-tgs:
Interestingly, during deserialization, exception isn't thrown on dotnet 2.1, but throws on dotnet 3.1.

Libraries/src/Amazon.Lambda.CloudWatchEvents/ECSEvents/Attachment.cs
```diff
@@ -30,4 +30,10 @@ public class Attachment
         /// </summary>
         public string Type { get; set; }
     }
+    
+    public class NameValuePair
+    {
+        public string Name { get; set; }
+        public string Value { get; set; }
```
zachdean, line 37: this should fix your test errors in 3.1 the DefaultLambdaSerializer is picky when it comes the names since there is a mix of camelCase and PascelCase.

```suggestion
      [DataMember(Name = "name")]
#if NETCOREAPP_3_1
      [System.Text.Json.Serialization.JsonPropertyName("name")]
#endif
        public string Name { get; set; }
        
      [DataMember(Name = "value")]
#if NETCOREAPP_3_1
      [System.Text.Json.Serialization.JsonPropertyName("value")]
#endif
        public string Value { get; set; }
```

ashishdhingra:
This has been implemented as part of larger PR https://github.com/aws/aws-lambda-dotnet/pull/868

Issue #679, opened 2020-05-22
.NET Core 3.1 runtime incompatibility with Amazon.Lambda.Logging.AspNetCore 3.0.0


wdolek:
We are in process of migrating our Lambdas to .NET Core 3.1 and I just found one incompatibility when using `Amazon.Lambda.Logging.AspNetCore` version `3.0.0`.

We are using logger scopes (to decorate all log messages with certain value).

Exception being thrown:
```
Could not load type 'Microsoft.Extensions.Logging.Abstractions.Internal.NullScope' from assembly 'Microsoft.Extensions.Logging.Abstractions, Version=3.1.4.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'.: TypeLoadException
   at Microsoft.Extensions.Logging.Logger.BeginScope[TState](TState state)
   at Microsoft.Extensions.Logging.Logger`1.Microsoft.Extensions.Logging.ILogger.BeginScope[TState](TState state)
   at Acme.Common.Extensions.LoggerExtensions.BeginScopeWithInformationalVersion(ILogger logger, Assembly assembly) in /builds/Acme.Common/Extensions/LoggerExtensions.cs:line 19
   at Acme.Common.Extensions.LoggerExtensions.BeginScopeWithInformationalVersion[T](ILogger logger) in /builds/Acme.Common/Extensions/LoggerExtensions.cs:line 10
   at Acme.Le.LeHandler.Handler(Dictionary`2 input, ILambdaContext context) in /builds/Acme.LeHandler/LeHandler.cs:line 89
   at lambda_method(Closure , Stream , Stream , LambdaContextInternal )
```

Our _csproj_ looks like this:
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netcoreapp3.1</TargetFramework>
    <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
    <AWSProjectType>Lambda</AWSProjectType>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Amazon.Lambda.Core" Version="1.1.0" />
    <PackageReference Include="Amazon.Lambda.Logging.AspNetCore" Version="3.0.0" />
    <PackageReference Include="Amazon.Lambda.Serialization.SystemTextJson" Version="2.0.1" />
    <PackageReference Include="Amazon.Lambda.SQSEvents" Version="1.1.0" />
[...]

PR #683, opened 2020-06-02
#679 Implement `NullScope` for netcore 3.1 packages compatibility

wdolek:
*Issue:*
#679 

*Description of changes:*
Implementing `NullScope` which became `internal` in `Microsoft.Extensions.Logging.Abstractions` 3.1.

Note: I wasn't able to compile whole solution, couldn't find any developer instructions. Unit tests for logging are passing.

---

### DIFF
--- a/Libraries/src/Amazon.Lambda.Logging.AspNetCore/Amazon.Lambda.Logging.AspNetCore.csproj
+++ b/Libraries/src/Amazon.Lambda.Logging.AspNetCore/Amazon.Lambda.Logging.AspNetCore.csproj
@@ -17,9 +17,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.1.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.1.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="2.1.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="[3.1.0,3.2.0)" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="[3.1.0,3.2.0)" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="[3.1.0,3.2.0)" />
   </ItemGroup>
 
 </Project>
--- a/Libraries/src/Amazon.Lambda.Logging.AspNetCore/Amazon.Lambda.Logging.AspNetCore.csproj
+++ b/Libraries/src/Amazon.Lambda.Logging.AspNetCore/Amazon.Lambda.Logging.AspNetCore.csproj
@@ -17,9 +17,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="[3.1.0,3.2.0)" />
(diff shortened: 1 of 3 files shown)


wdolek:
TODO:
- version of package should be bumped at least by minor version to `3.1.x` - where's nuspec?
- version of assembly file should be bumped as well, from `2.0.0.0` to at least `2.1.0.0`, preferably `3.1.0.0` to correspond with package version (`AssemblyInfo.cs`)
- update changelog - what's next release date?

Libraries/src/Amazon.Lambda.Logging.AspNetCore/Amazon.Lambda.Logging.AspNetCore.csproj
```diff
@@ -17,9 +17,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.1.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.1.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="2.1.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="[3.1.0,3.2.0)" />
```
normj, line 20: This library still needs to work with ASP.NET Core 2.1 Lambda functions so i don't want to bump the version to 3.1 as the minimum. Is there a reason we need to do that? Also what is the goal of the using a range blocking at 3.2? That would prevent usage for people that want to use this with .NET 5 preview using custom runtimes.

I think we can just take this part out of the PR and then release the NullScope fix.

wdolek: @normj that's the point of this PR:

- when on `netcore2.x` -> reference `Amazon.Lambda.Logging.AspNetCore` version `3.0.0` 
- when on `netcore3.x` -> reference `Amazon.Lambda.Logging.AspNetCore` version `3.1.0` 

if there are people on 5, we should of course provide another version for it.

The best would be to split code into branches based on runtime version. We should avoid mixing major versions - if I run `netcore3.1` I don't want to see dependency targeting `netcore2.1`, because it can have side effect elsewhere.

normj: Why would we want .NET Core 2.1 users use the versions that came for .NET Core 3.0? In general by targeting a lower version it will work for versions going forward. For example if you look at the Serilog log provider it still references version 2.0.0 of Microsoft.Extensions.Logging.

We would only need to create a separate build for the newer runtimes if we were using features that were only available in the newer runtimes.

In this case we made a poor decision by using a type that was in an "internal" namespace. And that is the part of the PR I want to take which fixes that.

wdolek: Oh, you are right - all of these are `netstandard`! I will revert that change and keep only `NullScope` thingy.

In other hand, what do you think about version range? What happens if my app depends on another package which references different version of one of those packages? (Basically what happened to me with logging)

normj:
Everything looks good. I'll work on pushing out new NuGet packages for the change.

normj:
Version 3.0.1 of Amazon.Lambda.Logging.AspNetCore has been released. And version 5.1.2 of Amazon.Lambda.AspNetCoreServer was released to pick up the new dependency.

PR #706, opened 2020-07-22
Set HOME environment variable for PowerShell Lambda

normj:
This is to work around an issue with using the -Parallel feature. The issue has been reported on the PowerShell repository:
https://github.com/PowerShell/PowerShell/issues/13189

The fix is done in the **Amazon.Lambda.PowerShellHost** package which host PowerShell inside the .NET Core Lambda runtime. The PowerShell module used to create the PowerShell Lambda bundle is updated to use the new version.

### DIFF
--- a/Libraries/src/Amazon.Lambda.PowerShellHost/PowerShellFunctionHost.cs
+++ b/Libraries/src/Amazon.Lambda.PowerShellHost/PowerShellFunctionHost.cs
@@ -166,11 +166,21 @@ $LambdaInput = ConvertFrom-Json -InputObject $LambdaInputString
 
 ";
 
-            var tempFolder = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("LAMBDA_TASK_ROOT")) ? "/tmp" : Path.GetTempPath();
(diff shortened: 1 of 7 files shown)


Libraries/src/Amazon.Lambda.PowerShellHost/PowerShellFunctionHost.cs
```diff
@@ -166,12 +166,22 @@ private IAsyncResult BeginInvoke(string input, ILambdaContext context)
 
 ";
 
-            var tempFolder = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("LAMBDA_TASK_ROOT")) ? "/tmp" : Path.GetTempPath();
+            var isLambda = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("LAMBDA_TASK_ROOT"));
+
+            var tempFolder = isLambda ? "/tmp" : Path.GetTempPath();
```
boblodgett, line 171: Is the ! correct on line 169? This is flipping /tmp and Path.GetTempPath() from what the code used to do when it is used on 171.

normj: It is correct. The old behavior was wrong. I added a new test to cover this missing area in the test coverage.

Issue #702, opened 2020-07-20
Make AbstractAspNetCoreFunction.Init() async


damianh:
Make this method async: https://github.com/aws/aws-lambda-dotnet/blob/b54acef044860345c7c201a078713fe3b026c285/Libraries/src/Amazon.Lambda.AspNetCoreServer/AbstractAspNetCoreFunction.cs#L189

## Describe the Feature

When initializing a webhost it is desirable to be able to load configuration asynchronously. For example, from Secrets Manager or Parameter Store. As this method is Sync, one is forced to use the nasty `GetAwaiter().GetResult()` when calling async APIs from within this methods.

This has a bit of knock on effect, however the function handler:

https://github.com/aws/aws-lambda-dotnet/blob/b54acef044860345c7c201a078713fe3b026c285/Libraries/src/Amazon.Lambda.AspNetCoreServer/AbstractAspNetCoreFunction.cs#L431-L436

... is already async so this change is entirely possible.

### Is your Feature Request related to a problem?

More of a quality-of-life improvement.

## Proposed Solution

- Add an `InitAsync()` methods so it's not breaking.
- Call `InitAsync()` before calling `Init()` - users should only override one but if they override both things should still work.
- Optional - add `[Obsolete("..", false)]` attribute to `Init()`

### Describe alternatives you've considered

Currently using `GetAwaiter().GetResult()` which is not desirable, especially when already in an async context (via the function handler).

## Additional Context

Real world usage - loading config from Secrets Manager and Paramater Store.

Marking is as _potentially_ breaking, though making it non-breaking would be preferable.

### Environment

* [x] :wave: I may be able to implement this feature request
* [x] :warning: This feature might incur a breaking change  

---

This is a :rocket: Feature Request

normj:
This is a valid feature request. I would be willing to take in a PR for this otherwise I will get to it when I can.

damianh:
I've taken a look and it appears it won't be possible without a breaking change. The default mechanism to intitialize the aspnet core webhost is calling `Start()` from a constructor which will always be synchronous.

https://github.com/aws/aws-lambda-dotnet/blob/e2bb81a4b48a402570b0de828772067938ca5671/Libraries/src/Amazon.Lambda.AspNetCoreServer/AbstractAspNetCoreFunction.cs#L103-L111

Ultimately `Init()` is called via here: 
[...]

PR #708, opened 2020-07-25
Proposal - Add Async PreInitAsync to allow async operations such as configuration I

damianh:
*Issue #, if available:*

A proposal to address #702

*Description of changes:*

- Add virtual member `PreInitAsync` to `AbstractAspNetCoreFunction` that allows async operations to happen, such as loading configuration from secrets manager, before the webhost is initialized. (I'm not too fond of the name. I'd prefer just `InitAsync()` and rename `Init()` to `InitWebHost()` but that would be a breaking change...)
- It is **only** invoked when `StartupMode` is set `FirstRequest` which in an asynchronous call stack. It is not possible to call it when `StartupMode` is set to `Constructor` as invoking async methods in a constructor is an anti-pattern. (If there is no specific advantage to using `Constructor` startup mode, could it be argued to remove it? 🤷 )
- Add test to assert `PreInitAsync()` is called. I also noticed there was no tests / test coverage for `StartMode.FirstRequest`; this test now covers it.

### DIFF
--- a/Libraries/src/Amazon.Lambda.AspNetCoreServer/AbstractAspNetCoreFunction.cs
+++ b/Libraries/src/Amazon.Lambda.AspNetCoreServer/AbstractAspNetCoreFunction.cs
@@ -189,6 +189,22 @@ namespace Amazon.Lambda.AspNetCoreServer
         protected virtual void Init(IWebHostBuilder builder) { }
 
         /// <summary>
+        /// Method to perform any action before initializing the web host. This is only invoked if StartupMode is set to FirstRequest.
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// protected override async Task PreInitAsync()
+        /// {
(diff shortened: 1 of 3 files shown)


normj:
I think having this only being invoked if StartupMode is set to `FirstRequest`, which isn't the default, is going to cause to much confusion.

Since users are forced to make async calls in an sync call right now if they need to load things async like secret manager should we just move the `GetAwaiter().GetResult()` to the `AbstractAspNetCoreFunction` and call the `PreInit` method at the same location `Init` is called. In the default case `Task.CompletedTask` is returned so that won't affect performance.

damianh:
Ok, will attempt your suggestion.

damianh:
Added a comment regarding this to original issue https://github.com/aws/aws-lambda-dotnet/issues/702#issuecomment-751465286. Not really happy with this proposal as is so going to close.

Issue #720, opened 2020-08-17
AspNetCoreServer: HTTP API v2 integration double escaped URL parameters


petli:
## Description
`AspNetCoreServer.APIGatewayHttpApiV2ProxyFunction` URL-escapes the query string, but since it is receiving the raw, escaped query string from the API gateway this will result in a double escaping. MVC strips out one level of escaping, but that means that the web API controllers get escaped values rather than unescaped values.

This is done here in the code: https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/Utilities.cs#L121-L148

I have verified with a minimal Python lambda that just returns the 2.0 event as a JSON response that the query string in the event is already escaped.

## Reproduction Steps

Test performed with a minimal controller based on `serverless.AspNetCoreWebAPI` but adapted to use HTTP API v2 instead.

Entry point:

    public class LambdaEntryPoint : Amazon.Lambda.AspNetCoreServer.APIGatewayHttpApiV2ProxyFunction
    {
        protected override void Init(IWebHostBuilder builder)
        {
            builder.UseStartup<Startup>();
        }
    }

Controller:

    [Route("api/[controller]")]
    public class ValuesController : ControllerBase
    {
        // GET api/values
        [HttpGet]
        public IActionResult Get([FromQuery] string a)
        {
            return Ok(new
            {
                A = a,
                RawQuery = Request.QueryString.ToString(),
            });
        }
    }

Setting up a HTTP v2 integration and calling it with `/api/values?a=b%3Dc` produces the following output:

    { 
      "a": "b%3Dc", 
      "rawQuery" : "?a=b%253Dc"
    }

The expected output is that `a` should have the unescaped value `b=c`.

### Environment

Build Version:

    <PackageReference Include="AWSSDK.Extensions.NETCore.Setup" Version="3.3.100.1" />
    <PackageReference Include="Amazon.Lambda.AspNetCoreServer" Version="5.1.1" />

Targeted .NET Platform:  Native .NET Core 3.1 runtime

## Resolution
- [x] :wave: I can/would-like-to implement a fix for this problem myself

---

This is a :bug: bug-report

ashishdhingra:
Hi @petli,

Good evening. 

I was able to reproduce the bug using the steps specified in the issue. I will discuss this issue with the .NET SDK development team, but it could be a while when it could be fixed. I will post any updates to this issue as it progresses.

Thanks,
[...]

Issue #237, opened 2018-02-28
Plus ("+") in URL is replaced with space (" ")


szilvaa:
## Expected Behavior
Plus ("+") warrants no special treatment in the URL outside of the  query string. I should receive plus ("+") in my lambda.  .net core 1.0 lambdas don't do this, only .net core 2.0 exhibits this behavior.

## Current Behavior
It appears the lambda hosting code replaces plus ("+") with space (" ") which then gets further encoded as %20. See cloudwatch log:
```
20:13:09 Incoming GET requests to /bla+bla
20:13:09 ASP.NET Core Request PathBase: , Path: /bla bla 
20:13:09 [Information] Microsoft.AspNetCore.Hosting.Internal.WebHost: Request starting GET https://example.com/bla%20bla
```
## Steps to Reproduce (for bugs)
This can be reproduced with any .net core 2.0 lambdas fronted by a proxy apigateway. Simple issue a request with a plus ("+") in the url i.e. GET https://example.com/bla+bla

normj:
I have reproduced the problem. I'll take a look into a fix.

normj:
Version 2.0.2 of Amazon.Lambda.AspNetCoreServer was just release that fixes this issue. Thanks for letting us know about the issue.

PR #721, opened 2020-08-17
#720: Do not double-escape URL query strings in HTTP API v2 functions

petli:
*Issue #, if available:* #720 

*Description of changes:* Remove code to escape query string, since it is already escaped in HTTP API v2. My guess is that this was implemented because v1 events needed to build a query string from unescaped values.

This may change the behaviour of issues like #237 (though that one must've been filed for API v1). However, + being replaced by space is done by ASP.NET Core MVC and not something that should be attempted to be suppressed by the lambda framework.


ashovlin:
Thank you for the PR! We've tested and are planning to accept it, and may combine with another fix we have in progress internally since they're introducing changes to existing behavior.

petli:
@ashovlin Do you have any update on the progress for accepting this PR, or has that other fix mentioned caused this PR to no longer be relevant?

normj (approved):

normj:
Approved the PR. It makes sense to get the behavior to running ASP.NET Core through Kestrel. Because this is a breaking change we will need to do a major version bump to 6.0.0.

ashovlin (approved):
Minor, but might be nice to update the comment above `CreateQueryStringParametersFromHttpApiV2` to clarify that the raw query string now matches the intended encoding and we're not encoding again.

(I know this sat a long time, so if not quick to revisit I imagine we can also clean it up when adding the commit for the version bump)

petli:
Thanks! I've updated the code comment, and also merged with master to aid merging this PR.

ashovlin:
Thanks! Updated docs look good.

normj:
PR released as part of version 6.0.0

Issue #736, opened 2020-09-23
Authentication info is not available in request context for mTLS


albertocorrales:
## Description
mTLS support was recently delivered for API Gateway. https://aws.amazon.com/blogs/compute/introducing-mutual-tls-authentication-for-amazon-api-gateway/. However, we are not able to get the client certificate, since the new "authorization" object is not available in requestContext object.

![image](https://user-images.githubusercontent.com/33038414/94065277-382f7e80-fdeb-11ea-915d-8f5c700f7710.png)

## Reproduction Steps
Authorization section is not available in ProxyRequestContext object, as we can see in https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.APIGatewayEvents/APIGatewayProxyRequest.cs

### Environment

* Build Version: master
* Targeted .NET Platform:  3.1

## Resolution
In order to solve this issue, object ProxyRequestContext should be extended with new "authorization" section

---

This is a :bug: bug-report

normj:
Thanks for bringing this to our attention. Looks like a pretty simple update to the event class.

normj:
I submitted PR to add the new fields.
https://github.com/aws/aws-lambda-dotnet/pull/737

normj:
Version 2.2.0 of Amazon.Lambda.APIGatewayEvents is out with the new properties for mTLS.

PR #737, opened 2020-09-25
Update for mlts

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/736

*Description of changes:*
Update the API Gateway event object to support the new ClientCert properties for API Gateway's Mutual TLS feature.

### DIFF
--- a/Libraries/src/Amazon.Lambda.APIGatewayEvents/APIGatewayProxyRequest.cs
+++ b/Libraries/src/Amazon.Lambda.APIGatewayEvents/APIGatewayProxyRequest.cs
@@ -393,13 +393,13 @@
             /// The date before which the certificate is invalid. Present when a client accesses an API by using a custom domain name 
             /// that has mutual TLS enabled. Present only in access logs if mutual TLS authentication fails.
             /// </summary>
-            public DateTime NotBefore { get; set; }
+            public string NotBefore { get; set; }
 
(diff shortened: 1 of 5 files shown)


philasmar (approved):

Libraries/src/Amazon.Lambda.APIGatewayEvents/APIGatewayProxyRequest.cs
```diff
@@ -332,6 +337,69 @@ public class RequestIdentity
...
+            /// <summary>
+            /// Properties for a client certificate.
+            /// </summary>
+            public ProxyRequestClientCert ClientCert { get; set; }
+        }
+
+        /// <summary>
+        /// Container for the properties of the client certificate.
+        /// </summary>
+        public class ProxyRequestClientCert
```
boblodgett, line 351: What is the reason the ProxyRequestClientCert, ClientCertValidity, and probably other classes are duplicated?

normj: They are duplicate because I can't control how API Gateway models the events in the future. Currently they are the same but in the future they might differ between the REST API event format and the HTTP API event format.

boblodgett (approved):

PR #748, opened 2020-10-13
Updating primaryOutputs in template.json to remove leading ./

sayedihashimi:
Updating `primaryOutputs` in `template.json` to remove leading `./` from the `primaryOutputs`. There is a bug in the TemplateEngine that this is working around. When there is a leading `./` in `primaryOutputs` this prevents the template from appearing in Visual Studio, see this blog post https://devblogs.microsoft.com/dotnet/net-cli-templates-in-visual-studio/

FYI there will be a few other things that you are likely to want to update to get the templates to behave well in Visual Studio. For example, you'll probably want to add an icon for each template. If you can start to get those icons put together, I can help you get the code changes in place. Fell free to ping me on twitter, when you are ready https://twitter.com/sayedihashimi

### DIFF
--- a/Blueprints/BlueprintDefinitions/netcore2.1/AspNetCoreWebAPI-FSharp/template/.template.config/template.json
+++ b/Blueprints/BlueprintDefinitions/netcore2.1/AspNetCoreWebAPI-FSharp/template/.template.config/template.json
@@ -43,7 +43,7 @@
   },
   "primaryOutputs": [
     {
-      "path": "./src/BlueprintBaseName.1/BlueprintBaseName.1.fsproj"
+      "path": "src/BlueprintBaseName.1/BlueprintBaseName.1.fsproj"
     }
   ]
 }
--- a/Blueprints/BlueprintDefinitions/netcore2.1/AspNetCoreWebAPI/template/.template.config/template.json
+++ b/Blueprints/BlueprintDefinitions/netcore2.1/AspNetCoreWebAPI/template/.template.config/template.json
@@ -43,7 +43,7 @@
   },
   "primaryOutputs": [
     {
-      "path": "./src/BlueprintBaseName.1/BlueprintBaseName.1.csproj"
+      "path": "src/BlueprintBaseName.1/BlueprintBaseName.1.csproj"
     }
   ]
 }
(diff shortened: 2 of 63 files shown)


normj:
Thanks for taking the time to engage with us on our blueprints. I think the support in VS to show CLI templates is great but I'm trying to figure out how it will work with these templates because we already expose them through Visual Studio through our own Project Wizard.

image

Now I think we will get them twice which is going to confuse users. Again as much as I like the feature I'm not sure it will be a better experience then our existing VS tailored experience. Will there be an option for templates that already have a VS experience to turn off the CLI template integration for just our package?

sayedihashimi:
@normj thanks for the info. I didn't realize that you already had an experience for this. I'm assuming that this come in through a VS Extension. 
When a user installs via `dotnet new --install`, would you prefer that we don't show those templates directly in the New Project Dialog? If so, I'll work on a sample of how to do that and share it with you here. Same question for Visual Studio for Mac. I'm not sure if you have an extension there, so I could envision you might want to show them directly in the New Project Dialog in VS for Mac but not in Visual Studio. If you let me know what you prefer, I'll get you a sample. I want to make sure that we have a good experience for these templates.

cc @jmatthiesen (PM working in Visual Studio for Mac)

normj:
You are correct, I would like them to not show up in VS but I would like them to show up in VS for Mac. The templates are exposed via our [AWS Toolkit for Visual Studio](https://marketplace.visualstudio.com/items?itemName=AmazonWebServices.AWSToolkitforVisualStudio2017)

sayedihashimi:
Ok, got it. I will work on a PR that has those changes. We may still need this PR to ensure that it works in VS for Mac. Is it possible to create the template .nupkg from this PR to see if it helps?

normj:
If you run the following command at the root of the repo it will write nupkg to the `Deployment/nuget-packages` directory. Don't worry about the 3.9.0 version. That is for anybody that still needs templates targeting .NET Core 2.1.

```
dotnet msbuild .\buildtools\build.proj  /t:package-blueprints
```

sayedihashimi:
Thanks. @jmatthiesen when you get a chance (and when the build is ready) can you try the AWS templates in VSM? If the project(s) are created, but not added to the solution, try building the .nupkg with changes from this PR and try again.

jmatthiesen:
Thanks for copying me in @sayedihashimi! @normj I'm trying to build the package using the build.proj you mention above, but it's not working on my Mac. Should I be able to build it from a Mac?

ashishdhingra:
@sayedihashimi @jmatthiesen @normj Good afternoon. I was going through the open PR list and stumbled across this PR. Please review if this PR is still valid (I see it referencing `.NET Core 2.1` templates) or if it should be closed.

Thanks,
Ashish

Issue #341, opened 2018-11-02
Is it possible to call AD PowerShell-based commands against MicrosoftAD Directory Service?


michael-crawford:
What I'm interested in doing, is from within a Lambda function - read a CSV file containing AD Group information, and create those groups in MicrosoftAD Directory.

Instead of having to a. create a separate EC2 Windows instance, b. Join this to that domain, c. Install AD Tools and d. run the commands from there.

Is this possible? If so, can someone post some code that would show how this could be done? I think many others might be interested. Examples I'm seeing on a quick browse seem to show how to control AWS via PowerShell, but not how to do what I'm asking. Creation of Users and Groups is not something that can be done via the Directory Service API currently.

jdemate:
Not a great solution, but some of the AWS applications that are integrated with AWS Directory Service have APIs that do user management within the directory.  For example, here is a WorkMail API that can create groups within the hosted Active Directory domain.
https://docs.aws.amazon.com/workmail/latest/APIReference/API_CreateGroup.html

As far as I know WorkMail only charges you for the users with a mailbox.  So as long as you don't register users to WorkMail you should not incur additional costs.

michael-crawford:
The advice from AWS Support, FWIW, is that while it may be technically possible to run PowerShell within a Lambda function to do these extra AD tasks, it's likely be be limited, buggy and just not the best way.

So, I'm currently 80% of the way through a new template that is meant to be a close equivalent to an existing template which builds a pair of Domain Controllers in two AZs, with cfn-init and PowerShell to setup subnets, zones, groups, users, DNS entries, etc. The new template creates a MicrosoftAD service and an associated Windows Instance, joins it to the domain, and will run the stuff AWS hasn't automated via the API on that Instance, again via cfn-init and PowerShell. Hopefully I can swap between these two as needed, that's the goal.

[...]

PR #781, opened 2020-12-15
Add dotnet sample for Active Directory LDAP usage.

smhmhmd:
*Issue #, if available:*

This function can be used with AWS Directory Services or Windows Active Directory for LDAP related operations.
This lambda must be placed in the same VPC as AWS Directory Services.
Also, DNS name must be enabled in the DHCP option set of the VPC
so that the directory can be accessed using directory name.
This is potentially a fix for  https://github.com/aws/aws-lambda-dotnet/issues/341


normj:
Apologize for never getting around to this PR but at this point they are too out of date.

Issue #774, opened 2020-12-09
Issue debugging .net serverless application (.net 5.0 container blueprint)


RNavaneethKumar:
## Description
Trying to run the project gives the following error "Unable to find LambdaDefaultsConfigFileParser.cs". However I have installed the Amazon Lambda Test Tool 5.0 latest version.

## Reproduction Steps
1. From Visual Studio, create a serverless project and choose .net 5 container as the blueprint
2. Run F5

![Error](https://user-images.githubusercontent.com/22679526/101677099-1f497780-3a82-11eb-840c-61f7ed22f6d5.PNG)

ashishdhingra:
Reproducible using the mentioned steps.

thylux:
The problem is reading serverless.template.
Just need to remove the below string from array.

From:
"ImageConfig": {
  "EntryPoint": [
    "/lambda-entrypoint.sh"
   ],
   **"Command": [
     "AWSServerless2::AWSServerless2.Functions::Get"
   ]**
 }

To:
"ImageConfig": {
  "EntryPoint": [
    "/lambda-entrypoint.sh"
   ],
   **"Command": "AWSServerless2::AWSServerless2.Functions::Get"**
 }

Don't believe the idea is to allow for multiple commands in the same lambda.

normj:
Thanks for reporting the issue. I have the pull request our to switch reading the `Command` as an array.

normj:
Version 0.11.1 was released that fixes the issue. Thanks for again for reporting the issue.

PR #785, opened 2020-12-26
Fix issue with incorrectly parsing image config command as a scalar instead of an array

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/774

*Description of changes:*
Fix issue reading serverless.template file for Command of ImageConfig. The code was incorrectly reading the command as a scalar which should be read as an array.


ppittle (approved):

ganeshnj (commented):
It would be nice to have unit tests for this.

ganeshnj (approved):

PR #792, opened 2021-01-05
Rework how Lambda Mock Test Tool Blazor version is packaged

normj:
*Issue #, if available:*

*Description of changes:*
Currently the Blazor version of the test tool used for both .NET Core 3.1 and .NET 5.0 is managed through a single csproj file and built with different build configurations. This is works locally but not with our release pipeline which does a build of the solution in one stage and then another stage does the packaging without a build. It is also designed to work with just the Release configuration.

The change is to use the main **Amazon.Lambda.TestTool.BlazorTester.csproj** project file just for building. It will build both the .NET Core 3.1 and .NET 5.0. Then there are separate **Amazon.Lambda.TestTool.BlazorTester31-pack.csproj** and **Amazon.Lambda.TestTool.BlazorTester50-pack.csproj** project files that are not part of the solution and just used for packaging.

The release pipeline will use compile the project in the build stage with the **Amazon.Lambda.TestTool.BlazorTester.csproj** project which is part of the solution. Then in the package stage of the pipeline the new project files will be used like all of the other project files in this repository for other NuGet packaging. 

### DIFF
--- a/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Amazon.Lambda.TestTool.BlazorTester50-pack.csproj
+++ b/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Amazon.Lambda.TestTool.BlazorTester50-pack.csproj
@@ -1,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <Import Project="..\..\..\..\buildtools\common.props" />
+  
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <Description>A tool to help debug and test your .NET Core 5.0 AWS Lambda functions locally.</Description>
+    <VersionPrefix>0.11.1</VersionPrefix>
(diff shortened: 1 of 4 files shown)


ppittle (approved):

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Amazon.Lambda.TestTool.BlazorTester50-pack.csproj
```diff
@@ -0,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <Import Project="..\..\..\..\buildtools\common.props" />
+  
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <Description>A tool to help debug and test your .NET Core 5.0 AWS Lambda functions locally.</Description>
```
ganeshnj, line 7: nit: .NET ~Core~ 5.0

ganeshnj (approved):

PR #794, opened 2021-01-06
Add dotnet sample for Active Directory LDAP with Kerberos

smhmhmd:
*Issue #, if available:*

*Description of changes:*
This function can be used with AWS Directory Services or Windows Active Directory for LDAP related operations.
LDAP authentication done using a keytab file and a username.
This lambda must be placed in the same VPC as AWS Directory Services.
Also, DNS name must be enabled in the DHCP option set of the VPC so that the directory can be accessed using directory name. Also, add S3 endpoint as per https://aws.amazon.com/blogs/aws/new-vpc-endpoint-for-amazon-s3/

     Example JSON input :
             {
               "directoryName": "XXX.YYY.com",
               "userName": "USER",
               "baseDN": "DC=XXX,DC=YYY,DC=com",
               "keytabFileS3Url": "s3://KEYTABLOCATION/my.keytab",
                "filter": "(objectClass=*)"
            }

This is potentially a fix for #341


normj:
Apologize for never getting around to this PR but at this point they are too out of date.

PR #834, opened 2021-03-13
fix: simulate lambda behavior for empty payload

russau:
When you invoke a function with no payload the Lambda service passes `{}` (an empty json object) to your handler input parameter.  The LambdaTestTool currently sends a null parameter not matching the way the service works.

For example, the function below is configured with `DefaultLambdaJsonSerializer`.  If you invoke with an empty payload in the LambdaTestTool the input parameter is null. Doing the the same in the Lambda service the input parameter is `{}` deserialized, which gives you an instance of `APIGatewayProxyRequest` with none of the properties set.

`public async Task<APIGatewayProxyResponse> FunctionHandler(APIGatewayProxyRequest input, ILambdaContext context)`

*Description of changes:*

* `LambdaExecutor.cs` - BuildParameters now passes `{}` for an empty payload
* `InvokeFunctionTests.cs` - WithEventParameterTest tests a method that takes an `S3Event` with an empty payload
* `FunctionSignatureExamples.csproj` - added reference to `Amazon.Lambda.S3Events`
* `InstanceMethods.cs` - added method that take an `S3Event`

### DIFF
--- a/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LambdaExecutor.cs
+++ b/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LambdaExecutor.cs
@@ -194,9 +194,10 @@ namespace Amazon.Lambda.TestTool.Runtime
                 {
                     parmValues[i] = context;
                 }
-                else if(request.Payload != null)
+                else
                 {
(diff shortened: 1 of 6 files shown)


normj (approved):

normj:
Merged into dev branch to go out in next release.

normj:
This PR was released as part of version 0.11.2 of the test tool.

Issue #570, opened 2020-01-21
HttpContext.Response.StatusCode set to zero when reading from middleware


corcus:
I am creating a logging middleware for an asp.net core mvc 2.2 api that is deployed in lambda.
Using version 4.0.0 of the Amazon.Lambda.AspNetCoreServer package and version 1.1.0 of Amazon.Lambda.RuntimeSupport package

This is the code for the invoke function

        public async Task InvokeAsync(HttpContext context)
        {
            try
            {
                var request = context.Request;

                var requestTime = DateTime.UtcNow;

                var requestBodyContent = await ReadRequestBody(request);

                var originalBodyStream = context.Response.Body;
                using (var responseBody = new MemoryStream())
                {
                    var response = context.Response;
                    //The context.Response.Body memory stream cannot be read from more than once. If we read it here to log it the response to the client will be empty 
                    //Instead we have all subsequent middleware write to the temporary responseBody memory stream which we can later read from
                    response.Body = responseBody;
                    logger.LogInformation($"status before : {response.StatusCode}");
                    //calculate the duration of the call
                    Stopwatch callDurationTimer = new Stopwatch();
                    callDurationTimer.Start();
                    await next(context);
                    callDurationTimer.Stop();
                    logger.LogInformation($"status after : {response.StatusCode}");

                    string responseBodyContent = null;
                    responseBodyContent = await ReadResponseBody(response);

                    //we write the contents of the temporary response memory stream to the original response memory stream in order to return the correct response. 
                    await responseBody.CopyToAsync(originalBodyStream);

                    if (ShouldLog(context))
                    {
                        var companyID = 0;
                        if (context.User.Identity.IsAuthenticated)
                        {
                            companyID = new InfoSerializer<ApiUserInfo>().Deserialize((ClaimsIdentity)context.User.Identity).OrganizationID;
                        }

                        var loggingCommand = FormSafeCreateCommand(
                            companyID,
                            DateTime.UtcNow,
                            context.Request.Path,
                            context.Request.Headers,
                            requestBodyContent,
                            await ReadResponseBody(context.Response),
                            callDurationTimer.Elapsed,
                            context.Response.StatusCode,
                            request.Method);

                        var result = loggingCreateHandler.Handle(loggingCommand); 
                    }
                }
            }
            catch (Exception e)
            {
                logger.LogError(e, "Logging Middleware Error");
                await next(context);
            }
        }

[...]

PR #858, opened 2021-05-12
Initialize StatusCode 200 to match behavior of Kestrel

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/570

*Description of changes:*
The Kestrel implementation of the IHttpResponseFeature initializes the Status code to 200. This fixes the initial issue that middleware was seeing the status code as 0.

The solution was found out by opening an issue on the aspnetcore repo for help which pointed me in the right direction.
https://github.com/dotnet/aspnetcore/issues/32596


Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/InvokeFeatures.cs
```diff
@@ -159,16 +159,16 @@ IEnumerator IEnumerable.GetEnumerator()
 
         IHeaderDictionary IHttpRequestFeature.Headers { get; set; } = new HeaderDictionary();
 
-        Stream IHttpRequestFeature.Body { get; set; } = new MemoryStream();
-
+        Stream IHttpRequestFeature.Body { get; set; } = new MemoryStream();
+
 #endregion
-
+
```
philasmar, line 165: nit: line endings issue?

philasmar (approved):

Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/InvokeFeatures.cs
normj: Updated all of the line endings for InvokeFeatures.cs in a separate commit.

Issue #353, opened 2018-12-03
LexEvent.CurrentIntent.SlotDetails Missing OriginalValue


EpaSafelite:
Is LexEvent.CurrentIntent.SlotDetails.OriginalValue something that's coming soon on .NET SDK?

ashishdhingra:
https://docs.aws.amazon.com/lex/latest/dg/lambda-input-response-format.html

ashishdhingra:
Addition of new property `OriginalValue` to [SlotDetail](https://github.com/aws/aws-lambda-dotnet/blob/20046865db2afd3cf57b1d92d492f1fb4ec08a9d/Libraries/src/Amazon.Lambda.LexEvents/LexEvent.cs#L152).

ashishdhingra:
Fixed in [Amazon.Lambda.LexEvents 3.0.0](https://www.nuget.org/packages/Amazon.Lambda.LexEvents/3.0.0)

PR #863, opened 2021-05-18
Added OriginalValue field to SlotDetail class for LexEvent.

ashishdhingra:
*Issue #:* #353 

*Description of changes:*
Added OriginalValue field to SlotDetail class for LexEvent.

---

### DIFF
--- a/Libraries/src/Amazon.Lambda.LexEvents/LexEvent.cs
+++ b/Libraries/src/Amazon.Lambda.LexEvents/LexEvent.cs
@@ -155,6 +155,11 @@ namespace Amazon.Lambda.LexEvents
             /// The resolutions array contains a list of additional values recognized for the slot.
             /// </summary>
             public IList<Dictionary<string, string>> Resolutions { get; set; }
+
+            /// <summary>
+            /// The originalValue field contains the value that was entered by the user for the slot.
+            /// </summary>
+            public string OriginalValue { get; set; }
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


ashovlin (approved):

Libraries/src/Amazon.Lambda.LexEvents/LexEvent.cs
```diff
@@ -155,6 +155,11 @@ public class SlotDetail
             /// The resolutions array contains a list of additional values recognized for the slot.
             /// </summary>
             public IList<Dictionary<string, string>> Resolutions { get; set; }
+
+            /// <summary>
+            /// The originalValue field contains the value that was entered by the user for the slot.
+            /// </summary>
+            public string OriginalValue { get; set; }
```
normj, line 162: Update the following test to check this value. https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/test/EventsTests.Shared/EventTests.cs#L1063

Looks like the test JSON already has a value set for `original-value` https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/test/EventsTests.Shared/lex-event.json#L21

normj (changes requested):

Libraries/src/Amazon.Lambda.LexEvents/LexEvent.cs
ashishdhingra: Done

normj (approved):

Issue #884, opened 2021-06-30
ApiGatewayLambdaFunction/APIGatewayHttpApiV2ProxyFunction returns status code 200 instead of 500 in case of exception


akurin:
## Description
Lambda functions created using Amazon.Lambda.AspNetCoreServer return incorrect status code if an exception is thrown. Instead of expected status code 500, status code 200 is returned. 

## Reproduction Steps
The tests for the issue described already exist. Just run [TestCallingWebAPI.TestDefaultResponseErrorCode](https://github.com/aws/aws-lambda-dotnet/blob/00009c7d4fb4fb406c355a99d11d142718ba4374/Libraries/test/Amazon.Lambda.AspNetCoreServer.Test/TestCallingWebAPI.cs#L126) or [TestApiGatewayHttpApiV2Calls.TestDefaultResponseErrorCode](https://github.com/aws/aws-lambda-dotnet/blob/00009c7d4fb4fb406c355a99d11d142718ba4374/Libraries/test/Amazon.Lambda.AspNetCoreServer.Test/TestApiGatewayHttpApiV2Calls.cs#L110).

### Environment
* Build Version: Amazon.Lambda.AspNetCoreServer 6.0.2
* OS Info: Fedora Linux 33
* Build Environment: dotnet cli
* Targeted .NET Platform: 3.1

## Resolution
- [ ] :wave: I can/would-like-to implement a fix for this problem myself

I believe the issue was introduced in f8b4df5521d19bc53ade32432f977e7d77e1755e. It is also troubling that this code made it into the master branch at a time when the tests did not pass.

---

This is a :bug: bug-report

ashishdhingra:
Reproducible using test [TestCallingWebAPI.TestDefaultResponseErrorCode](https://github.com/aws/aws-lambda-dotnet/blob/00009c7d4fb4fb406c355a99d11d142718ba4374/Libraries/test/Amazon.Lambda.AspNetCoreServer.Test/TestCallingWebAPI.cs#L126)

normj:
Thanks for reporting the issue and sorry for it happening. I have a PR to get this fixed. https://github.com/aws/aws-lambda-dotnet/pull/891

You are correct about the tests. I'm also investigate to see what is going on with our build job that it is not failing our builds due to test failures.

normj:
Version 6.0.3 of Amazon.Lambda.AspNetCoreServer was released that fixes this issue.

PR #891, opened 2021-07-14
Fix issue with ASP.NET Core Lambda functions that have internal errors to correctly return 500 status code

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/884

*Description of changes:*
The previous PR [858](https://github.com/aws/aws-lambda-dotnet/pull/858) changed the default HTTP status code to 200 to fix issues with ASP.NET Core middleware. That change had unintended consequences of interfering with exceptions logic when in the ProcessRequest method which sends the request into ASP.NET Core. 

The original logic was catching any exceptions and if the status code was not set then set the status code to 500. Since the previous PR always sets the status code to default to 200 then it would never get set to 500 if there was an exception.

Since an exception should only be caught by the ProcessRequest method if nothing handled it in ASP.NET Core then the status code should always be set to 500 if an exception is caught. This is what is being done in this PR to explicitly set the status code if an exception is caught.

### DIFF
--- a/Libraries/src/Amazon.Lambda.AspNetCoreServer/AbstractAspNetCoreFunction.cs
+++ b/Libraries/src/Amazon.Lambda.AspNetCoreServer/AbstractAspNetCoreFunction.cs
@@ -495,7 +495,7 @@ namespace Amazon.Lambda.AspNetCoreServer
                     }
 
                     _logger.LogError(sb.ToString());
-                    defaultStatusCode = 500;
+                    ((IHttpResponseFeature)features).StatusCode = 500;
                 }
                 catch (ReflectionTypeLoadException rex)
                 {
@@ -516,14 +516,14 @@ namespace Amazon.Lambda.AspNetCoreServer
                     }
 
                     _logger.LogError(sb.ToString());
-                    defaultStatusCode = 500;
+                    ((IHttpResponseFeature)features).StatusCode = 500;
                 }
                 catch (Exception e)
                 {
                     ex = e;
                     if (rethrowUnhandledError) throw;
                     _logger.LogError($"Unknown error responding to request: {this.ErrorReport(e)}");
-                    defaultStatusCode = 500;
+                    ((IHttpResponseFeature)features).StatusCode = 500;
                 }
 
                 if (features.ResponseStartingEvents != null)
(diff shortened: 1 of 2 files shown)


ashovlin (approved):
Looks good functionally, some minor/non-blocking comments:
* Comment on 475 doesn't apply anymore?
* Wonder if the `StatusCode = responseFeatures.StatusCode != 0 ? responseFeatures.StatusCode : statusCodeIfNotSet` in the `MarshallResponse` implementations is dead now that `StatusCode` has a default value. I don't think harmful to leave in though.

normj:
> Looks good functionally, some minor/non-blocking comments:
> 
> * Comment on 475 doesn't apply anymore?
> * Wonder if the `StatusCode = responseFeatures.StatusCode != 0 ? responseFeatures.StatusCode : statusCodeIfNotSet` in the `MarshallResponse` implementations is dead now that `StatusCode` has a default value. I don't think harmful to leave in though.

I thought about cleaning up the dead code but since the method is virtual and customers could override it and create their own behavior with expectations on this method I decided to leave it alone.

Libraries/src/Amazon.Lambda.AspNetCoreServer/AbstractAspNetCoreFunction.cs
```diff
@@ -495,7 +495,7 @@ protected async Task<TRESPONSE> ProcessRequest(ILambdaContext lambdaContext, obj
                     }
 
                     _logger.LogError(sb.ToString());
-                    defaultStatusCode = 500;
+                    ((IHttpResponseFeature)features).StatusCode = 500;
```
ganeshnj, line 498: shouldn't it work without casting?

https://github.com/aws/aws-lambda-dotnet/blob/49c1d23754cdb7896664107f46c1e88990b7f060/Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/InvokeFeatures.cs#L27

normj: No because InvokeFeatures implements the interfaces using explicitly interface implementation. It requires you to access the properties and methods of the interface by the variable actually being declared or casted as the interface.

ganeshnj (approved):

PR #922, opened 2021-10-01
wip

ganeshnj:
*Issue #, if available:*

*Description of changes:*


Libraries/src/Amazon.Lambda.Annotations.SourceGenerators/LambdaFunctionCodeGenerator.cs
```diff
@@ -0,0 +1,261 @@
...
+                    var iServiceCollectionSymbol =
+                        _context.Compilation.GetTypeByMetadataName(
+                            "Microsoft.Extensions.DependencyInjection.IServiceCollection");
+
+                    var classModel = _context.Compilation.GetSemanticModel(_startupSyntax.SyntaxTree);
+                    foreach (var member in _startupSyntax.Members.Where(member => member.Kind() == SyntaxKind.MethodDeclaration))
+                    {
+                        var methodSyntax = (MethodDeclarationSyntax)member;
+                        var methodSymbol = classModel.GetDeclaredSymbol(methodSyntax);
+                        if (methodSymbol != null
```
normj, line 69: I think this needs to check if the method is `public`. As I understand this is looking for any method that takes in a single parameter of type `IServiceCollection`. Is this better then looking for a specific named method like `ConfigiureServices?

Libraries/src/Amazon.Lambda.Annotations.SourceGenerators/LambdaFunctionCodeGenerator.cs
```diff
@@ -0,0 +1,261 @@
...
+            }
+        }
+
+        public INamedTypeSymbol ReturnType => LambdaFunctionMethodSymbol.ReturnType as INamedTypeSymbol;
+
+        /// <summary>
+        /// Represents the Configure(IServiceCollection) method in the LambdaStartup attributed class.
+        /// </summary>
+        /// <exception cref="NotSupportedException">Thrown when multiple LambdaStartup attributed classes exist in the project.</exception>
+        public IMethodSymbol ConfigureMethodSymbol
```
normj, line 49: Semantic code searching is very meta and will always be hard to understand. Please add lots of inline comments of exactly what you are attempting. This is a general rule for all generator code.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerators/LambdaFunctionCodeGenerator.cs
```diff
@@ -0,0 +1,261 @@
...
+        /// <summary>
+        /// Generates <see cref="SourceText"/> for the LambdaFunction including setting up the dependency injection.
+        /// </summary>
+        /// <returns>A <see cref="Tuple"/> containing source file name hint and source text.</returns>
+        public (string, SourceText) GenerateSource()
+        {
+            var apiGatewayResponseSymbol =
+                _context.Compilation.GetTypeByMetadataName("Amazon.Lambda.APIGatewayEvents.APIGatewayProxyResponse");
+
+            var returnTypeString = "APIGatewayProxyResponse";
```
normj, line 130: Why are being specific for `APIGatewayProxyResponse`? Is this a temporary thing to get something working?

Libraries/src/Amazon.Lambda.Annotations.SourceGenerators/LambdaFunctionCodeGenerator.cs
```diff
@@ -0,0 +1,261 @@
...
+            if (LambdaFunctionMethodSymbol.IsAsync)
+            {
+                returnTypeString = $"System.Threading.Tasks.Task<{returnTypeString}>";
+            }
+
+            var source = new StringBuilder(
+$@"using System;
+using Amazon.Lambda.Core;
+using Amazon.Lambda.APIGatewayEvents;
+using Microsoft.Extensions.DependencyInjection;
```
normj, line 140: Is it going to be later task to handle a Lambda function that isn't using DI and in that case doesn't have a package reference to Microsoft.Extensions.DependencyInjection?

Libraries/src/Amazon.Lambda.Annotations.SourceGenerators/LambdaFunctionCodeGenerator.cs
```diff
@@ -0,0 +1,261 @@
...
+            serviceProvider = services.BuildServiceProvider();
+        }}
+
+        public{(LambdaFunctionMethodSymbol.IsAsync ? " async " : " ")}{returnTypeString} {LambdaFunctionMethodSymbol.Name}(APIGatewayProxyRequest request, ILambdaContext _context)
+        {{
+            // Create a scope for every request,
+            // this allows creating scoped dependencies without creating a scope manually.
+            using var scope = serviceProvider.CreateScope();
+
+            var {LambdaFunctionMethodSymbol.ContainingType.Name.ToCamelCase()} = scope.ServiceProvider.GetRequiredService<{LambdaFunctionMethodSymbol.ContainingType}>();");
```
normj, line 176: If I understand this would mean the constructor of the type containing the LambdaFunction would be called for every request. I think that would be a big break to the programming model for Lambda. I was thinking the scope would be for each invocation of the method. User's could use the `FromService` attribute on the method to get scoped services otherwise use the constructor for singleton services.

normj (changes requested):
Most of my comments might be because this is an initial version to get started and for work to happen in future tasks.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerators/LambdaFunctionCodeGenerator.cs
ganeshnj: I initially thought of that but then realized, if the method won't be public, customers will get static analyzer warning which they are already familiar with.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerators/LambdaFunctionCodeGenerator.cs
ganeshnj: currently only dealing with APIGateway events, like for SQS, Dynamo etc. This will go away once we other event types.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerators/LambdaFunctionCodeGenerator.cs
ganeshnj: I think DI is must, Startup is optional. This will always be there. We inject customer code as scoped and resolve later, which needs Microsoft.Extensions.DependencyInjection package.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerators/LambdaFunctionCodeGenerator.cs
ganeshnj: That's right, customer code is like a ASP.NET Web API controller which is initialized for every request (it is a scoped dependency like the comment mention in L156).

The generated function follows the current Lambda function model, where constructor is called only once during the invocation and then stays in memory as long as Lambda wants it. We are getting best of both worlds here IMO Lambda + ASP.NET Core.

That said, if customer ever want to change the behavior of their Function (attributed with LambdaFunction), they can configure the lifetime of Function class in LambdaStartup.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerators/LambdaFunctionCodeGenerator.cs
```diff
@@ -0,0 +1,261 @@
+using System;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Text;
+using System.Linq;
+using System.Text;
+using System.Security.Cryptography;
+
+namespace Amazon.Lambda.Annotations.SourceGenerators
```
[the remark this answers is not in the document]
normj: But there is only one C# Source Generator which will have sub components of generation but still one source generator.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerators/LambdaFunctionCodeGenerator.cs
ganeshnj: Will match the exact signature including the method name because customers can have multiple methods with same IServiceCollection parameter and we will end up using a wrong method in cases.

PR #944, opened 2021-10-29
Check for Metatdata Before Modifying the Serverless Template

96malhar:
*Description of changes:*
This PR does the following
* Prevents the `CloudFormationJsonWriter` from modifying functions that were not created using the `Amazon.Lambda.Annotation` package.
* Add Metadata to existing functions in the `TestServerlessApp` to indicate that they were created using the `Amazon.Lambda.Annotation` package

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationJsonWriter.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationJsonWriter.cs
@@ -31,6 +31,8 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Writers
             
             foreach (var lambdaFunction in report.LambdaFunctions)
             {
+                if (!ShouldProcessLambdaFunction(lambdaFunction)) 
+                    continue;
                 ProcessLambdaFunction(lambdaFunction);
                 processedLambdaFunctions.Add(lambdaFunction.Name);
             }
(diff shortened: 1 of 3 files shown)


Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationJsonWriter.cs
```diff
@@ -31,22 +31,30 @@ public void ApplyReport(AnnotationReport report)
...
                 ApplyLambdaFunctionDefaults(lambdaFunctionPath, propertiesPath);
-
+            }
+            else
+            {
+                var creationTool = _jsonWriter.GetToken($"{lambdaFunctionPath}.Metadata.Tool", string.Empty);
+                if (!string.Equals(creationTool.ToObject<string>(), "Amazon.Lambda.Annotations", StringComparison.Ordinal))
+                    return false;
+            }
+
```
ganeshnj, line 57: Extract the logic to check whether a lambda function can be mutated to separate method. `ProcessLambdaFunction` must not be called if metadata is not expected.

something like in L34.
```csharp
if (GetToolMetadata(arg1, ag2, ...) != "Amazon.Lambda.Annotations") 
{
     continue;
}

ProcessLambdaFunction(arg1)
```

ganeshnj (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationJsonWriter.cs
96malhar: Fixed

ganeshnj (approved):

normj (approved):

PR #949, opened 2021-10-31
feat: add support for FromRoute attribute

ganeshnj:
*Issue #, if available:*

*Description of changes:*

- [x] add support for FromRoute attribute
FromRoute attribute is resolved last in the chain because there could be a parameter which is not marked as FromRoute but marked as FromBody. It will be bad if it is matched for FromRoute and lead to an unexpected state.
- [x] Ability to pass route parameter without using FromRoute attribute

**Example 1**
```csharp
[LambdaFunction(Name = "SimpleCalculatorMultiply")]
[RestApi(HttpMethod.Get, "/SimpleCalculator/Multiply/{x}/{y}")]
public string Multiply(int x, int y)
{
    return _simpleCalculatorService.Multiply(x, y).ToString();
}
```
translates to
```csharp
public APIGatewayProxyResponse Multiply(APIGatewayProxyRequest request, ILambdaContext context)
{
    // Create a scope for every request,
    // this allows creating scoped dependencies without creating a scope manually.
    using var scope = serviceProvider.CreateScope();
    var simpleCalculator = scope.ServiceProvider.GetRequiredService<SimpleCalculator>();

    var x = default(int);
    if (request.PathParameters?.ContainsKey("x") == true)
    {
        x = (int)Convert.ChangeType(request.PathParameters["x"], typeof(int));
    }

    var y = default(int);
    if (request.PathParameters?.ContainsKey("y") == true)
    {
        y = (int)Convert.ChangeType(request.PathParameters["y"], typeof(int));
    }

    var response = simpleCalculator.Multiply(x, y);

    return new APIGatewayProxyResponse
    {
        Body = response,
        Headers = new Dictionary<string, string>
        {
            {"Content-Type", "text/plain"}
        },
        StatusCode = 200
    };
}
```

**Example 2**
```csharp
[LambdaFunction(Name = "SimpleCalculatorDivideAsync")]
[RestApi(template: "/SimpleCalculator/DivideAsync/{x}/{y}", method: HttpMethod.Get)]
public async Task<int> DivideAsync([FromRoute(Name = "x")]int first, [FromRoute(Name = "y")]int second)
{
    return await Task.FromResult(_simpleCalculatorService.Divide(first, second));
}
```
translates to
```csharp
public async Task<APIGatewayProxyResponse> DivideAsync(APIGatewayProxyRequest request, ILambdaContext context)
{
    // Create a scope for every request,
    // this allows creating scoped dependencies without creating a scope manually.
    using var scope = serviceProvider.CreateScope();
    var simpleCalculator = scope.ServiceProvider.GetRequiredService<SimpleCalculator>();

    var first = default(int);
    if (request.PathParameters?.ContainsKey("x") == true)
    {
        first = (int)Convert.ChangeType(request.PathParameters["x"], typeof(int));
    }

    var second = default(int);
    if (request.PathParameters?.ContainsKey("y") == true)
    {
        second = (int)Convert.ChangeType(request.PathParameters["y"], typeof(int));
    }

    var response = await simpleCalculator.DivideAsync(first, second);

    var body = System.Text.Json.JsonSerializer.Serialize(response);

    return new APIGatewayProxyResponse
    {
        Body = body,
        Headers = new Dictionary<string, string>
        {
            {"Content-Type", "application/json"}
        },
        StatusCode = 200
    };
}
```

- [x] Test

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/TemplateParametersExtension.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/TemplateParametersExtension.cs
@@ -1,0 +1,48 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Amazon.Lambda.Annotations.SourceGenerator.Models.Attributes
+{
+    public static class TemplateParametersExtension
+    {
+        /// <summary>
+        /// Returns aa <see cref="HashSet{T}"/> of string containing template parameter
+        /// parsed from the <see cref="RestApiAttribute.Template"/>.
+        /// </summary>
+        /// <param name="attribute">this <see cref="RestApiAttribute"/> object.</param>
+        public static HashSet<string> GetTemplateParameters(this RestApiAttribute attribute)
+        {
+            var template = attribute.Template;
+            return GetTemplateParameters(template);
+        }
+
+        /// <summary>
+        /// Returns aa <see cref="HashSet{T}"/> of string containing template parameter
+        /// parsed from the <see cref="HttpApiAttribute.Template"/>.
+        /// </summary>
+        /// <param name="attribute">this <see cref="HttpApiAttribute"/> object.</param>
+        public static HashSet<string> GetTemplateParameters(this HttpApiAttribute attribute)
+        {
+            var template = attribute.Template;
+            return GetTemplateParameters(template);
+        }
+
+        private static HashSet<string> GetTemplateParameters(string template)
+        {
+            var parameters = new HashSet<string>();
+
+            var components = template.Split('/').Where(c => !string.IsNullOrWhiteSpace(c));
+            foreach (var component in components)
+            {
+                if (component.StartsWith("{") && component.EndsWith("}"))
+                {
+                    var parameter = component.Substring(1, component.Length - 2);
+                    parameters.Add(parameter);
+                }
+            }
+
+            return parameters;
+        }
+    }
+}
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
@@ -7,6 +7,7 @@
 <#@ import namespace="Amazon.Lambda.Annotations.SourceGenerator.Extensions" #>
 <#@ import namespace="Amazon.Lambda.Annotations.SourceGenerator.Models.Attributes" #>
 <#@ import namespace="Microsoft.CodeAnalysis" #>
+<#@ import namespace="Amazon.Lambda.Annotations.SourceGenerator.Validation" #>
 <#
     foreach (var ns in _model.GeneratedMethod.Usings)
     {
@@ -79,8 +80,23 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
     if (_model.LambdaMethod.Events.Contains(EventType.API))
     {
         var parameters = string.Join(", ", _model.LambdaMethod.Parameters.Select(p => p.Name));
-        var restApiAttribute = _model.LambdaMethod.Attributes.FirstOrDefault(att => att.Type.FullName == TypeFullNames.RestApiAttribute);
+        var restApiAttribute = _model.LambdaMethod.Attributes.FirstOrDefault(att => att.Type.FullName == TypeFullNames.RestApiAttribute) as AttributeModel<RestApiAttribute>;
         var httpApiAttribute = _model.LambdaMethod.Attributes.FirstOrDefault(att => att.Type.FullName == TypeFullNames.HttpApiAttribute) as AttributeModel<HttpApiAttribute>;
+
+        if (restApiAttribute != null && httpApiAttribute != null)
+        {
+            throw new NotSupportedException($"A method cannot have both {TypeFullNames.RestApiAttribute} and {TypeFullNames.HttpApiAttribute} attribute at the same time.");
+        }
+
+        var routeParameters = restApiAttribute?.Data?.GetTemplateParameters() ?? httpApiAttribute?.Data?.GetTemplateParameters() ?? new HashSet<string>();
+
+        var (routeTemplateValid, missingRouteParams) = RouteParametersValidator.Validate(routeParameters, _model.LambdaMethod.Parameters);
+        if (!routeTemplateValid)
+        {
+            var template = restApiAttribute?.Data?.Template ?? httpApiAttribute?.Data?.Template ?? string.Empty;
+            throw new InvalidOperationException($"Route template {template} is invalid. Missing {string.Join(",", missingRouteParams)} parameters in method definition.");
+        }
+
         foreach (var parameter in _model.LambdaMethod.Parameters)
         {
             if (parameter.Attributes.Any(att => att.Type.FullName == TypeFullNames.FromServiceAttribute))
@@ -151,7 +167,9 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
             }
             else if (parameter.Attributes.Any(att => att.Type.FullName == TypeFullNames.FromHeaderAttribute))
             {
-                var fromHeaderAttribute = parameter.Attributes.First(att => att.Type.FullName == TypeFullNames.FromHeaderAttribute) as AttributeModel<FromHeaderAttribute>;
+                var fromHeaderAttribute =
+                    parameter.Attributes.First(att => att.Type.FullName == TypeFullNames.FromHeaderAttribute) as
+                        AttributeModel<FromHeaderAttribute>;
 
                 // Use parameter name as key, if Name has not specified explicitly in the attribute definition.
                 var headerKey = fromHeaderAttribute?.Data?.Name ?? parameter.Name;
@@ -187,7 +205,8 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
 
                     // Generic types are mapped using Select statement to the target parameter type argument.
                     var typeArgument = parameter.Type.TypeArguments.First();
-                    var select = $".Select(q => ({typeArgument.FullName})Convert.ChangeType(q, typeof({typeArgument.FullName}))).ToList()";
+                    var select =
+                        $".Select(q => ({typeArgument.FullName})Convert.ChangeType(q, typeof({typeArgument.FullName}))).ToList()";
 #>
             if (request.<#= headers #>?.ContainsKey("<#= headerKey #>") == true)
             {
@@ -207,7 +226,6 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
 
 <#
                 }
-
             }
             else if (parameter.Attributes.Any(att => att.Type.FullName == TypeFullNames.FromBodyAttribute))
             {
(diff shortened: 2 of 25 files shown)


Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/Snapshots/Greeter_SayHelloAsync_Generated.g.cs
```diff
@@ -19,9 +19,9 @@ public Greeter_SayHelloAsync_Generated()
         public async Task<APIGatewayProxyResponse> SayHelloAsync(APIGatewayProxyRequest request, ILambdaContext context)
         {
             var firstNames = default(System.Collections.Generic.IEnumerable<string>);
-            if (request.Headers?.ContainsKey("names") == true)
+            if (request.MultiValueHeaders?.ContainsKey("names") == true)
             {
-                firstNames = request.Headers["names"].Split(",").Select(q => (string)Convert.ChangeType(q, typeof(string))).ToList();
+                firstNames = request.MultiValueHeaders["names"].Select(q => (string)Convert.ChangeType(q, typeof(string))).ToList();
```
ganeshnj, line 24: Snapshot was wrong because of a bug in creating FromHeader attribute.

Arguments are passed in constructor than property initializer.

Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/Snapshots/Greeter_SayHello_Generated.g.cs
```diff
@@ -18,9 +18,9 @@ public Greeter_SayHello_Generated()
         public APIGatewayProxyResponse SayHello(APIGatewayProxyRequest request, ILambdaContext context)
         {
             var firstNames = default(System.Collections.Generic.IEnumerable<string>);
-            if (request.QueryStringParameters?.ContainsKey("names") == true)
+            if (request.MultiValueQueryStringParameters?.ContainsKey("names") == true)
             {
-                firstNames = request.QueryStringParameters["names"].Split(",").Select(q => (string)Convert.ChangeType(q, typeof(string))).ToList();
+                firstNames = request.MultiValueQueryStringParameters["names"].Select(q => (string)Convert.ChangeType(q, typeof(string))).ToList();
```
ganeshnj, line 23: Same L24 in above file.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/TemplateParametersExtension.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            var parameters = new HashSet<string>();
+
+            var components = attribute.Split('/').Where(c => !string.IsNullOrWhiteSpace(c));
+            foreach (var component in components)
+            {
+                if (component.StartsWith("{") && component.EndsWith("}"))
+                {
+                    var parameter = component.Substring(1, component.Length - 2);
+                    if (IsValid(parameter))
```
normj, line 40: If there is a `FromRoute` attribute that is mapping the template parameter name to the C# variable name does the template parameter name have to follow C# naming rules?

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
```diff
@@ -79,8 +79,16 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
     if (_model.LambdaMethod.Events.Contains(EventType.API))
     {
         var parameters = string.Join(", ", _model.LambdaMethod.Parameters.Select(p => p.Name));
-        var restApiAttribute = _model.LambdaMethod.Attributes.FirstOrDefault(att => att.Type.FullName == TypeFullNames.RestApiAttribute);
+        var restApiAttribute = _model.LambdaMethod.Attributes.FirstOrDefault(att => att.Type.FullName == TypeFullNames.RestApiAttribute) as AttributeModel<RestApiAttribute>;
         var httpApiAttribute = _model.LambdaMethod.Attributes.FirstOrDefault(att => att.Type.FullName == TypeFullNames.HttpApiAttribute) as AttributeModel<HttpApiAttribute>;
+
+        if (restApiAttribute != null && httpApiAttribute != null)
```
normj, line 85: Not that having both makes a lot of sense but I suspect that we have encoded into our logic that you can only have one event source per LambdaFunction. We should revisit this as we add more event sources to allow users to potentially have more then one event source. Imagine the scenario we have a method that for an API it gets is payload from an the response body but in an SQS event source it gets its payload from the SQS queue. We can tackle this later as we flesh out other event sources but just keep in mind this is a common request we get to have a single Lambda function with multiple event sources.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/TemplateParametersExtension.cs
normj: Also it doesn't every seem valid to skip parameters. If we can't handle the name then we should throw an error.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
```diff
@@ -227,6 +237,21 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
...
+            {
+                var fromRouteAttribute = parameter.Attributes?.FirstOrDefault(att => att.Type.FullName == TypeFullNames.FromRouteAttribute) as AttributeModel<FromRouteAttribute>;
+
+                // Use parameter name as key, if Name has not specified explicitly in the attribute definition.
+                var routeKey = fromRouteAttribute?.Data?.Name ?? parameter.Name;
+#>
+            var <#= parameter.Name #> = default(<#= parameter.Type.FullName #>);
+            if (request.PathParameters?.ContainsKey("<#= routeKey #>") == true)
+            {
+                <#= parameter.Name #> = (<#= parameter.Type.FullName #>)Convert.ChangeType(request.PathParameters["<#= routeKey #>"], typeof(<#= parameter.Type.FullName #>));
```
normj, line 250: We also need to think about all of these `ChangeType` calls how to better handle if the type can't be converted. This is user input data which can be wrong. Should the end result we automatically send back an 400 bad request saying this parameter is the wrong data type?

normj (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
ganeshnj: I think that's fair, a bad request can be sent on type conversion failure. I will check in ASP.NET Core as well to make sure the behavior is same/similar.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/TemplateParametersExtension.cs
ganeshnj: > If there is a FromRoute attribute that is mapping the template parameter name to the C# variable name does the template parameter name have to follow C# naming rules?

No, atleast in ASP.NET Core, the logic needs to modified to accomodate FromRoute attribute with any string. I just tested 
```
        [HttpGet("{1}/{id}")]
        public string Get([FromRoute(Name = "1")]int id2, int id)
        {
            return "value";
        }
```
and it works.

> Also it doesn't every seem valid to skip parameters. If we can't handle the name then we should throw an error.

That's fair, I'm okay throwing an error and let customer know that need to fix the template.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
ganeshnj: The current design has been kept flexible and it can be changed to allow multiple events on a Lambda function. Once we add a new event, we can have a better design to move forward.

I'm even inclined over removing `EventType` enum and only rely on attributes (they were not modeled previously) which gives better flexibility to incorporate multiple events.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/TemplateParametersExtension.cs
ganeshnj: I just removed the validation logic all together.

ganeshnj: Added route parameter validation, now generator makes sure every route parameter is mapped. If any parameter found unmapped, it throws.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Validation/RouteParametersValidator.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            var missingRouteParams = new List<string>();
+            var isValid = true;
+
+            foreach (var routeParam in routeParameters)
+            {
+                var fromRouteMethodParam = lambdaMethodParameters
+                    .FirstOrDefault(p => p.Attributes.Any(att => (att as AttributeModel<FromRouteAttribute>)?.Data.Name == routeParam));
+
+                // There exists a parameter attributed with FromRouteAttribute
```
normj, line 29: This comment is written as a partial thought. The comment should explain why it is worth skipping if we have a `FromRouteAttribute`. Maybe just rewording to something like "Route parameter valid because there is a matching signature parameter with a FromRoute attribute.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Validation/RouteParametersValidator.cs
```diff
@@ -0,0 +1,56 @@
...
+                if (fromRouteMethodParam != null)
+                {
+                    continue;
+                }
+
+                var nameMatchingMethodParam = lambdaMethodParameters.FirstOrDefault(p => p.Name == routeParam);
+
+                // There doesn't exist a parameter with matching name as template parameter
```
normj, line 37: Same comment as above that is is a partial thought comment.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Validation/RouteParametersValidator.cs
```diff
@@ -0,0 +1,56 @@
...
+                if (nameMatchingMethodParam == null)
+                {
+                    missingRouteParams.Add(routeParam);
+                    isValid = false;
+                    continue;
+                }
+
+                // There exists a route parameter matching with method parameter name but also has one of conflicting From attribute
+                var conflictingAttributes = nameMatchingMethodParam.Attributes.Where(att => NonFromRouteAttributes.Contains(att.Type.FullName)).ToList();
```
normj, line 46: Instead of maintaining the `NonFromRouteAttributes` list could you just check to see if the parameter has any attribute other then FromAttribute then it is an error? I worry the `NonFromRouteAttributes` collection is going to be something we will forget to update when we add new attributes.

normj (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Validation/RouteParametersValidator.cs
ganeshnj: Reworded.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Validation/RouteParametersValidator.cs
ganeshnj: Reworded.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Validation/RouteParametersValidator.cs
ganeshnj: Changed the check to be namespace specific.

normj (approved):

96malhar (approved):

PR #952, opened 2021-11-01
fix: replace .NET 3.1 with 5.0 in CI to build test app

ganeshnj:
*Issue #, if available:*
Source generator is not working for sample app in CI.

*Description of changes:*
The issue was because of using .NET 3.1 SDK, to compile a .NET Core 3.1 using source generator, .NET SDK must be 5.0 or above.

To verify, go to Checks -> Source Generator CI ->test-app-build-result, download the uploaded artifacts from CI and open TestServerlessApp.dll. 

image

### DIFF
--- a/.github/workflows/source-generator-ci.yml
+++ b/.github/workflows/source-generator-ci.yml
@@ -41,10 +41,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Setup .NET 3.1
+    - name: Setup .NET 5.0
       uses: actions/setup-dotnet@v1
       with:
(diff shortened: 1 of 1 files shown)


normj (approved):
Yup, this is exactly why I want the library to target .NET 6 for GA so we don't have issues like this.

ganeshnj:
> 
> 
> Yup, this is exactly why I want the library to target .NET 6 for GA so we don't have issues like this.

+1, when we were discussing the target framework, it rang bells in my head on why the test app is failing.

96malhar (approved):

PR #953, opened 2021-11-02
Assert on serverless.template snapshots in unit tests

96malhar:
*Issue #, if available:*
DOTNET-5494

*Description of changes:*
This PR adds assertions on the content of the `serverless.template` in unit tests

**Points to note**
1. This PR excludes the original `serverless.template` file from being copied in the to `bin` directory. This is done so that a new template is created from every unit test. 
2. The `serverless.template` snapshot does not contain event attributes because they are yet to be supported. They will be added as part of DOTNET-5512

### DIFF
--- a/Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/SourceGeneratorTests.cs
+++ b/Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/SourceGeneratorTests.cs
@@ -1,3 +1,4 @@
+using System;
 using System.IO;
 using System.Text;
 using System.Threading.Tasks;
@@ -38,6 +39,10 @@ namespace Amazon.Lambda.Annotations.SourceGenerators.Tests
                     }
                 }
             }.RunAsync();
+
+            var expectedTemplateContent = File.ReadAllText(Path.Combine("Snapshots", "ServerlessTemplates", "greeter.template"));
+            var actualTemplateContent = File.ReadAllText(Path.Combine("TestServerlessApp", "serverless.template"));
+            VerifyServerlessTemplateSnapshot(expectedTemplateContent, actualTemplateContent);
         }
 
         [Fact]
@@ -83,6 +88,10 @@ namespace Amazon.Lambda.Annotations.SourceGenerators.Tests
                     }
                 }
             }.RunAsync();
+
+            var expectedTemplateContent = File.ReadAllText(Path.Combine("Snapshots", "ServerlessTemplates", "simpleCalculator.template"));
+            var actualTemplateContent = File.ReadAllText(Path.Combine("TestServerlessApp", "serverless.template"));
+            VerifyServerlessTemplateSnapshot(expectedTemplateContent, actualTemplateContent);
         }
 
         [Fact]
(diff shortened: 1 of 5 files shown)


Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/SourceGeneratorTests.cs
```diff
@@ -83,6 +90,12 @@ public async Task SimpleCalculator()
                     }
                 }
             }.RunAsync();
+
+            var expectedTemplateContent = File.ReadAllText(Path.Combine("Snapshots", "ServerlessTemplates", "simpleCalculator.template"));
+            var actualTemplateContent = File.ReadAllText(Path.Combine("TestServerlessApp", "serverless.template"));
+            expectedTemplateContent = expectedTemplateContent.Replace("\r\n", "\n");
+            actualTemplateContent = actualTemplateContent.Replace("\r\n", "\n");
```
ganeshnj, line 97: do not modify the actual template. Only modify the expected template to make it compatible with the environment where the test are being run.

Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/SourceGeneratorTests.cs
```diff
@@ -83,6 +90,12 @@ public async Task SimpleCalculator()
                     }
                 }
             }.RunAsync();
+
+            var expectedTemplateContent = File.ReadAllText(Path.Combine("Snapshots", "ServerlessTemplates", "simpleCalculator.template"));
+            var actualTemplateContent = File.ReadAllText(Path.Combine("TestServerlessApp", "serverless.template"));
+            expectedTemplateContent = expectedTemplateContent.Replace("\r\n", "\n");
```
ganeshnj, line 96:
```suggestion
            expectedTemplateContent = expectedTemplateContent.Replace("\r\n", System.Environment.NewLine);
            expectedTemplateContent = expectedTemplateContent.Replace("\n", System.Environment.NewLine);
```

this will ensure that expected template always have system based new line.

ganeshnj (changes requested):

Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/SourceGeneratorTests.cs
96malhar: You will ultimately end up with \r\r\n. I tried this and it didn't work.

ganeshnj: There needs be a better way, It seems solvable without modifying the actual template.

96malhar: Converting everything to unix based endings is the cleanest way (imo). Also, we are just comparing strings and not writing anything back to the template file.

ganeshnj: by converting the actual output to a different format you are changing the expectations (I didn't mean expected output here).

There should not a be a manipulation on the output to make the test pass.

96malhar: Added an extra replace on \r\r\n.

Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/SourceGeneratorTests.cs
96malhar: Fixed

ganeshnj (approved):

normj (approved):

PR #954, opened 2021-11-02
feat: package analyzer and library using nupsec

ganeshnj:
*Issue #, if available:*

*Description of changes:*
nuget packaging for source generator is done via custom nupsec, it has two components

1. Library part: makes sure that all the symbols are resolved in the customer code when they add attributes.
2. Analyzer part: makes sure that source code can run without any issues. It must have all the dependencies that are required to it to run. Analyzers can't depend on the NuGet references.

CI is also updated and now a NuGet package is generated during checks which can be downloaded and tested.

image

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.nuspec
+++ b/Libraries/src/Amazon.Lambda.Annotations.nuspec
@@ -1,0 +1,25 @@
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+    <metadata>
+        <id>Amazon.Lambda.Annotations</id>
+        <version>0.1.0-preview1</version>
+        <authors>Amazon Web Services</authors>
+        <tags>AWS Amazon Lambda</tags>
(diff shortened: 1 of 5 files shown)


Libraries/src/Amazon.Lambda.Annotations.nuspec
```diff
@@ -0,0 +1,25 @@
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+    <metadata>
+        <id>Amazon.Lambda.Annotations</id>
+        <version>0.1.0-preview1</version>
+        <authors>Amazon Web Services</authors>
+        <tags>AWS Amazon Lambda</tags>
+        <description>Annotations that can be added to Lambda projects to generate C# code and CloudFormation templates.</description>
```
ganeshnj, line 7: @ik130 need your input here, can you provider a description that we want to use on nuget.org.

normj (approved):

96malhar (approved):

PR #956, opened 2021-11-03
ci: create nuget package on every push to feature/annotations

ganeshnj:
*Issue #, if available:*

*Description of changes:*

This change allows not only do checks on PRs but also on merge to `feature/annotations` to allow testing nuget package without looking into PRs. Anyone can go any commit and grab the nuget package and use it.


normj (approved):

Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/WriterTests/CloudFormationJsonWriterTests.cs
```diff
@@ -252,7 +252,8 @@ public void DoNotModifyFunctionWithoutRequiredMetadata()
             Assert.Equal(100, rootToken["Resources"]["MethodNotCreatedFromAnnotationsPackage"]["Properties"]["Timeout"]); // unchanged
 
             var policies = rootToken["Resources"]["MethodNotCreatedFromAnnotationsPackage"]["Properties"]["Policies"] as JArray;
-            Assert.Equal(1, policies.Count);
+            Assert.NotNull(policies);
+            Assert.Single(policies);
```
96malhar, line 256: Having only `Assert.Single` will suffice. `Assert.NotNull` is redundant unless I am missing something.

ganeshnj: This is required to silence the static analyzer, technically it can be null but the setup ensures it will never be. 

Just another safe guard.

96malhar (approved):

PR #960, opened 2021-11-05
Delete Redundant Event Attributes Entries from serverless.template

96malhar:
*Description of changes:*
When a customer changes their event attributes (for ex, changing from HTTP GET attribute to a HTTP POST attribute) the previous attributes are still retained in the `severless.template`.

This PR addresses this issue by adding new metadata called `SyncedEvents` which maintains a list of events that are created by the `Amazon.Lambda.Annotations` package

On every compilation of the target application, currently synced events are compared with the previous ones and redundant entries are deleted.

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationJsonWriter.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationJsonWriter.cs
@@ -83,27 +83,48 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Writers
                 _jsonWriter.RemoveToken($"{propertiesPath}.Role");
             }
             
-            if (lambdaFunction.Attributes != null && lambdaFunction.Attributes.Any())
-                ProcessLambdaFunctionAttributes(lambdaFunction);
+            ProcessLambdaFunctionEvents(lambdaFunction);
         }
 
-        private void ProcessLambdaFunctionAttributes(ILambdaFunctionSerializable lambdaFunction)
+        private void ProcessLambdaFunctionEvents(ILambdaFunctionSerializable lambdaFunction)
         {
+            var currentSyncedEvents = new List<string>();
+            string eventName;
+            
             foreach (var attributeModel in lambdaFunction.Attributes)
             {
                 switch (attributeModel)
                 {
                     case AttributeModel<HttpApiAttribute> httpApiAttributeModel:
-                        ProcessHttpApiAttribute(lambdaFunction, httpApiAttributeModel.Data);
+                        eventName = ProcessHttpApiAttribute(lambdaFunction, httpApiAttributeModel.Data);
+                        currentSyncedEvents.Add(eventName);
                         break;
                     case AttributeModel<RestApiAttribute> restApiAttributeModel:
-                        ProcessRestApiAttribute(lambdaFunction, restApiAttributeModel.Data);
+                        eventName = ProcessRestApiAttribute(lambdaFunction, restApiAttributeModel.Data);
+                        currentSyncedEvents.Add(eventName);
                         break;
                 }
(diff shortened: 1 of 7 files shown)


Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationJsonWriter.cs
```diff
@@ -83,6 +83,7 @@ private void ProcessLambdaFunction(ILambdaFunctionSerializable lambdaFunction)
                 _jsonWriter.RemoveToken($"{propertiesPath}.Role");
             }
             
+            _jsonWriter.RemoveToken($"{lambdaFunctionPath}.Properties.Events");
```
normj, line 86: Looks like the logic is every time we do generation delete the events section and then recreate it. That will mess up users  that are customizing the event object in the template. For example we aren't modeling in code all of the authorization fields that can be done for the HTTP event. So they would need to update the field.

We need a better way of knowing when the attribute has been removed then we delete the event. Possibly putting more state in the metadata section of the AWS::Serverless::Function that identifies what fields were last synced.

normj (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationJsonWriter.cs
ganeshnj: Agree with Norm, this too risky to delete the existing events and add every single time there is change. Only the relevant JSON blob must be update (not create delete).

96malhar: Added logic to surgically remove redundant attributes.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationJsonWriter.cs
```diff
@@ -83,27 +83,49 @@ private void ProcessLambdaFunction(ILambdaFunctionSerializable lambdaFunction)
...
+                        eventName = ProcessRestApiAttribute(lambdaFunction, restApiAttributeModel.Data);
+                        currentSyncedEvents.Add(eventName);
                         break;
                 }
             }
+
+            var eventsPath = $"Resources.{lambdaFunction.Name}.Properties.Events";
+            var syncedEventsMetadataPath = $"Resources.{lambdaFunction.Name}.Metadata.SyncedEvents";
+            
+            if (!(_jsonWriter.GetToken(syncedEventsMetadataPath, new JArray()) is JArray previousSyncedEvents))
```
normj, line 112: This logic can be simplified with having only one return and one place of setting the syncevents json token.

```
if (_jsonWriter.GetToken(syncedEventsMetadataPath, new JArray()) is JArray previousSyncedEvents)
{
	foreach (var previousEventName in previousSyncedEvents.Select(x => x.ToObject<string>()))
	{
		if (!currentSyncedEvents.Contains(previousEventName))
			_jsonWriter.RemoveToken($"{eventsPath}.{previousEventName}");
	}
}


if (currentSyncedEvents.Any())
	_jsonWriter.SetToken(syncedEventsMetadataPath, new JArray(currentSyncedEvents));

```

normj (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationJsonWriter.cs
96malhar: Fixed

normj (approved):

ganeshnj (approved):

Issue #973, opened 2021-11-14
ILambdaLogger.LogLine(...) behaviour changed between .Net 5 and .Net 6


El-Gor-do:
## Description
In .Net 5, _ILambdaLogger.LogLine("hello world")_ would be logged to CloudWatch Logs exactly as
```
hello world
```
After changing my Lambda project's TargetFramework from net5.0 to net6.0, the same call to _ILambdaLogger.LogLine_ is logged to CloudWatch Logs as "{timestamp}\t{requestId}\t{level}\t{message}"
```
2021-11-14T21:32:01.022Z	8247f3ba-2335-4b12-bf19-e379605ba146	info	hello world
```

In my real Lambda project, my message is a json string which I subsequently deserialise. With the new behaviour in .Net 6, the message is no longer valid json because of the extra tab-delimited fields in each message.

## Reproduction Steps
In a custom runtime Lambda, set the function handler to this
```
public static DateTime FunctionHandler(ILambdaContext context)
{
    context.Logger.LogLine("hello world");
    return DateTime.UtcNow;
}
```

Deploy this Lambda using TargetFramework _net5.0_ then invoke the function and check the CloudWatch Logs output. The log event's message will be "hello world". Deploy the Lambda again using TargetFramework _net6.0_ then invoke the function again and check the CloudWatch Logs output again. The log event's message will have tab-delimited fields followed by "hello world".

### Environment
* Build Version: Amazon.Lambda.Core 2.1.0
* OS Info: AmazonLinux2
* Build Environment: docker
* Targeted .NET Platform:  net5.0 and net6.0

## Resolution
I would like the .Net 6 behaviour to return to the previous .Net 5 behaviour where the log event's message is exactly as I passed to _ILambdaLogger.LogLine(...)_
---

This is a :bug: bug-report

ashishdhingra:
Reproducible. While we cannot use Visual Studio 2019 for .NET 6 targets, AWS extensions for .NET CLI could be used to deploy the updated project with target as `net6.0`.

El-Gor-do:
The change in behaviour appears to be due to changing the IConsoleLoggerWriter implementation from SimpleLoggerWriter to LogLevelLoggerWriter.

https://github.com/aws/aws-lambda-dotnet/blob/70d6a97b6430d5f43e7cc38950b56d7d0f353841/Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/ConsoleLoggerWriter.cs#L203

[...]

PR #977, opened 2021-11-16
Add AWS_LAMBDA_HANDLER_LOG_FORMAT environment variable

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/973

*Description of changes:*
Add a new environment variable called `AWS_LAMBDA_HANDLER_LOG_FORMAT`. These are the current valid values: `Default` and `Unformatted`. 

This helps users that prefer the pre .NET 6 logging format which didn't added time, request id and log level into each message. Those users can set the environment variable to `Unformatted` and the log message will be written just as the user sent the message into the log api.


ganeshnj (approved):

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/ConsoleLoggerWriter.cs
```diff
@@ -202,25 +217,32 @@ internal void FormattedWriteLine(string message)
...
-                    line = $"{DateTime.UtcNow.ToString("yyyy-MM-ddTHH:mm:ss.fffZ")}\t{CurrentAwsRequestId}\t{displayLevel}\t{message ?? string.Empty}";
+                if (_logFormatType == LogFormatType.Unformatted)
+                {
+                    _innerWriter.WriteLine(message);
                 }
                 else
                 {
-                    line = $"{DateTime.UtcNow.ToString("yyyy-MM-ddTHH:mm:ss.fffZ")}\t{CurrentAwsRequestId}\t{message ?? string.Empty}";
+                    string line;
+                    if (!string.IsNullOrEmpty(level))
```
boblodgett, line 236: Should this be `displayLevel` instead of `level`?

Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.IntegrationTests/CustomRuntimeTests.cs
```diff
@@ -136,7 +137,29 @@ private async Task RunLoggingTestAsync(AmazonLambdaClient lambdaClient, string h
...
+            var invokeResponse = await InvokeFunctionAsync(lambdaClient, JsonConvert.SerializeObject(""));
+            Assert.True(invokeResponse.HttpStatusCode == System.Net.HttpStatusCode.OK);
+            Assert.True(invokeResponse.FunctionError == null);
+
+            var log = System.Text.UTF8Encoding.UTF8.GetString(Convert.FromBase64String(invokeResponse.LogResult));
+
+            Assert.DoesNotContain("info\t", log);
+            Assert.DoesNotContain("warn\t", log);
+            Assert.DoesNotContain("fail\t", log);
+            Assert.DoesNotContain("crit\t", log);
```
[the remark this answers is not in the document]
normj: That is matching what .NET Logging does. I think there pattern was to keep all the log levels 4 characters.

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/ConsoleLoggerWriter.cs
normj: Switched to `displayLevel` although the results will be the same.

boblodgett (approved):

PR #978, opened 2021-11-16
feat: Update execution environment

96malhar:
*Issue #, if available:*
DOTNET-5543

*Description of changes:*
This PR updates the execution environment so that the appropriate metadata can be attached to the user agent string during service calls via the AWS SDK for .NET 

Verified the User-Agent string by making an S3 call in an existing Lambda function in the `TestServerlessApp`
image

image


Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/GeneratedMethodModelBuilder.cs
```diff
@@ -35,7 +35,10 @@ public static class GeneratedMethodModelBuilder
             {
                 "System",
                 "System.Linq",
-                "System.Collections.Generic"
+                "System.Collections.Generic",
+                "System.Text",
+                "System.Reflection",
+                "Amazon.Lambda.Annotations"
```
ganeshnj, line 41: nit: add Amazon using at the end

ganeshnj (approved):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/GeneratedMethodModelBuilder.cs
96malhar: Fixed

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
```diff
@@ -388,5 +389,28 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
     }
 #>
         }
+
+        private static void SetExecutionEnvironment()
+        {
+            const string envName = "AWS_EXECUTION_ENV";
+            const string amazonLambdaAnnotations = "amazon-lambda-annotations";
+
+            var assemblyVersion = typeof(LambdaFunctionAttribute).Assembly
```
normj, line 398: We shouldn't need to do reflection for this. The source generator should be able to determine the version at compile time and put the value there.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/GeneratedMethodModelBuilder.cs
```diff
@@ -35,7 +35,9 @@ public static class GeneratedMethodModelBuilder
             {
                 "System",
                 "System.Linq",
-                "System.Collections.Generic"
+                "System.Collections.Generic",
+                "System.Text",
+                "System.Reflection",
```
normj, line 40: We should not be needing reflection. That is the point of using source generators.

normj (changes requested):

Libraries/test/TestServerlessApp/aws-lambda-tools-defaults.json
```diff
@@ -12,6 +12,6 @@
   "s3-prefix": "TestServerlessApp/",
   "template": "serverless.template",
   "template-parameters": "",
-  "s3-bucket": "test-serverless-app",
-  "stack-name": "test-serverless-app"
-}
\ No newline at end of file
+"s3-bucket" : "test-serverless-app-8bfa3c02",
+"stack-name" : "test-serverless-app-8bfa3c02"
+}
```
ganeshnj, line 17: It must not be committed.

ganeshnj (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
96malhar: Fixed

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/GeneratedMethodModelBuilder.cs
96malhar: Fixed

Libraries/test/TestServerlessApp/aws-lambda-tools-defaults.json
96malhar: Fixed

Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/Snapshots/ComplexCalculator_Add_Generated.g.cs
```diff
@@ -33,5 +35,24 @@ public Amazon.Lambda.APIGatewayEvents.APIGatewayHttpApiV2ProxyResponse Add(Amazo
...
+
+            var envValue = new StringBuilder();
+
+            // If there is an existing execution environment variable add the annotations package as a suffix.
+            if(!string.IsNullOrEmpty(Environment.GetEnvironmentVariable(envName)))
+            {
+                envValue.Append($"{Environment.GetEnvironmentVariable(envName)}_");
+            }
+
+            envValue.Append($"{amazonLambdaAnnotations}_{sourceGeneratorVersion}");
```
normj, line 53: Since this is generated why not generated with the values actually in the Append call to avoid doing the string interpolation. That also means the lines 42 and 43 disappear and this line looks like

```
envValue.Append("amazon-lambda-annotations_0.1.0.0");
```

normj (changes requested):

normj (approved):

PR #1007, opened 2021-12-02
fix: removing InternalCustomResponseExceptionHandling in ASP.NET core…

MDanialSaleem:
fix: removing InternalCustomResponseExceptionHandling in ASP.NET core server that resulted in implementation details being leaked.

*Issue #, if available:* 886

*Description of changes:*
The default behavior of ASP.NET core when an unhandled exception is thrown in a controller is to respond with the type of exception in an HTTP header. This is done by the ```InternalCustomResponseExceptionHandling``` method. This leaks implementation details. Per [this comment](https://github.com/aws/aws-lambda-dotnet/issues/886#issuecomment-877416642) I have removed that behavior.

Prior to this, overriding this behavior required override the ```ProcessRequest``` method that called the ```InternalCustomResponseExceptionHandling``` method since the latter is private protected. This PR does not change this behavior. One still has to override ```ProcessRequest``` in order to change behavior on unhandled exceptions.

[x] I have run the tests and they pass.

### DIFF
--- a/Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayHttpApiV2ProxyFunction.cs
+++ b/Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayHttpApiV2ProxyFunction.cs
@@ -51,10 +51,6 @@ namespace Amazon.Lambda.AspNetCoreServer
             _hostServices = hostedServices;
         }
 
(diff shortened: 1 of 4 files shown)


ashishdhingra:
PR https://github.com/aws/aws-lambda-dotnet/pull/899 looks similar and adds opt-in behavior. Should this PR be closed?

normj:
I ended merging the similar https://github.com/aws/aws-lambda-dotnet/pull/899 PR because it allowed users that want the existing behavior to be able to opt in to that behavior.

PR #1008, opened 2021-12-03
Correct json property names for Cognito identity properties in lambda normj/runtimesupport-cognito-fix

normj:
*Description of changes:*
The Amazon.Lambda.RuntimeSupport package was using the incorrect json properties for mashaling the Cognito properties into the lambda context object.

Also did a few tweaks to make the test more reliable and work in VS 2022.

### DIFF
--- a/Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.IntegrationTests/BaseCustomRuntimeTest.cs
+++ b/Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.IntegrationTests/BaseCustomRuntimeTest.cs
@@ -143,6 +143,7 @@ namespace Amazon.Lambda.RuntimeSupport.IntegrationTests
                     BucketName = bucketName
                 };
                 await s3Client.PutBucketAsync(putBucketRequest);
+                await Task.Delay(10000);
             }
 
             // write or overwrite deployment package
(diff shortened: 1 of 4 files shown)


Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.IntegrationTests/BaseCustomRuntimeTest.cs
```diff
@@ -143,6 +143,7 @@ private async Task CreateBucketWithDeploymentZipAsync(IAmazonS3 s3Client, string
                     BucketName = bucketName
                 };
                 await s3Client.PutBucketAsync(putBucketRequest);
+                await Task.Delay(10000);
```
klementi, line 146: Love the arbitrary delays ;)

Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.IntegrationTests/BaseCustomRuntimeTest.cs
```diff
@@ -243,6 +244,7 @@ protected async Task CreateFunctionAsync(IAmazonLambda lambdaClient, string buck
                     }
                 }
             }
+            await Task.Delay(5000);
```
klementi, line 247: More delays wtf! ;)

klementi (approved):

boblodgett (approved):

Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.IntegrationTests/BaseCustomRuntimeTest.cs
ganeshnj: A better fix would be periodically checking the existence of the bucket with a timeout then using a constant delay which can never be trusted 100%.

PR #1010, opened 2021-12-05
Port .NET Core 3.1 RIC PreJit feature to Amazon.Lambda.RuntimeSupport

normj:
*Issue #, if available:*

*Description of changes:*
The .NET Core 3.1 managed runtime has the feature that when the `AWS_LAMBDA_DOTNET_PREJIT` environment is set or the Lambda function is being started for provisioned concurrency the RIC will warm up additional parts of the .NET runtime and Assemblies to make the first invoke run faster.

This PR ports that feature from the .NET Core 3.1 RIC to Amazon.Lambda.RuntimeSupport. Most of the code is same as the 3.1 RIC with additionally logging added. The internal logger was also reworked to make it easier to access within Amazon.Lambda.RuntimeSupport.

During testing with the feature I discovered that .NET 6 is less forgiving about the culture names being loaded then .NET Core 3.1 was. The locale prewarmed was being determined by the "LANG" environment variable which contains the text encoding as well (i.e. en_US.UTF-8). With .NET 6 this was throwing an exception because of the suffix .UTF-8. The 3.1 RIC code was modified to strip off the text encoding.

Testing was done with building .NET 6 container image with this version of Amazon.Lambda.RuntimeSupport and doing self contained publish with custom runtimes.

### DIFF
--- a/Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/InvokeDelegateBuilder.cs
+++ b/Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/InvokeDelegateBuilder.cs
@@ -34,6 +34,7 @@ namespace Amazon.Lambda.RuntimeSupport.Bootstrap
         private readonly InternalLogger _logger;
         private readonly HandlerInfo _handler;
         private readonly MethodInfo _customerMethodInfo;
+        private Type CustomerOutputType { get; set; }
 
         public InvokeDelegateBuilder(InternalLogger logger, HandlerInfo handler, MethodInfo customerMethodInfo)
         {
@@ -60,7 +61,7 @@ namespace Amazon.Lambda.RuntimeSupport.Bootstrap
         /// <param name="customerObject">Wrapped customer object.</param>
         /// <param name="customerSerializerInstance">Instance of lambda input & output serializer.</param>
         /// <returns>Action delegate pointing to customer's handler.</returns>
-        public Action<Stream, ILambdaContext, Stream> ConstructInvokeDelegate(object customerObject, object customerSerializerInstance)
+        public Action<Stream, ILambdaContext, Stream> ConstructInvokeDelegate(object customerObject, object customerSerializerInstance, bool isPreJit)
         {
             var inStreamParameter = Expression.Parameter(Types.StreamType, "inStream");
             var outStreamParameter = Expression.Parameter(Types.StreamType, "outStream");
@@ -68,6 +69,11 @@ namespace Amazon.Lambda.RuntimeSupport.Bootstrap
 
             _logger.LogDebug($"UCL : Constructing input expression");
             var inputExpression = BuildInputExpressionOrNull(customerSerializerInstance, inStreamParameter, out var iLambdaContextType);
+            if (isPreJit)
+            {
+                _logger.LogInformation("PreJit: inputExpression");
+                UserCodeInit.InitDeserializationAssembly(inputExpression, inStreamParameter);
+            }
 
             _logger.LogDebug($"UCL : Constructing context expression");
             var contextExpression = BuildContextExpressionOrNull(iLambdaContextType, contextParameter);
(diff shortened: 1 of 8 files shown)


Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/InvokeDelegateBuilder.cs
```diff
@@ -60,14 +61,19 @@ public InvokeDelegateBuilder(InternalLogger logger, HandlerInfo handler, MethodI
...
         {
             var inStreamParameter = Expression.Parameter(Types.StreamType, "inStream");
             var outStreamParameter = Expression.Parameter(Types.StreamType, "outStream");
             var contextParameter = Expression.Parameter(typeof(ILambdaContext), "context");
 
             _logger.LogDebug($"UCL : Constructing input expression");
             var inputExpression = BuildInputExpressionOrNull(customerSerializerInstance, inStreamParameter, out var iLambdaContextType);
+            if (isPreJit)
+            {
+                _logger.LogInformation("PreJit: inputExpression");
```
carlzogh, line 74: I think this should rather be logging at debug level to reduce chances of cluttering the function's CW logs for information it may not have asked for - ditto for other places in this PR. Do you think there's value for this as `info` over `debug`? Would this only end up logging if `LAMBDA_RUNTIMESUPPORT_DEBUG` is set?

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/UserCodeInit.cs
```diff
@@ -0,0 +1,194 @@
...
+                                continue;
+
+                            RuntimeHelpers.PrepareMethod(
+                                method.MethodHandle);
+
+                        }
+                        catch (Exception e)
+                        {
+                            InternalLogger.GetDefaultLogger().LogError(e, "PreJit: Error with PrepareAssembly");
+                        }
```
carlzogh, line 176: How likely is this to happen, and how would one go about addressing it? (this exception is swallowed in .NET Core 3.1 RIC)

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/UserCodeInit.cs
```diff
@@ -0,0 +1,194 @@
...
+                    }
+                }
+            }
+            void LoadReferencedAssemblies(Assembly assembly)
```
carlzogh, line 180: nit: missing new line between methods

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/UserCodeInit.cs
```diff
@@ -0,0 +1,194 @@
...
+                else
+                {
+                    /*A dummy operation on string*/
+                    "Lambdalanguages".ToUpper();
+                }
+            }
+            catch (Exception e)
+            {
+                InternalLogger.GetDefaultLogger().LogError(e, "PreJit: Error with LoadStringCultureInfo");
+            }
```
carlzogh, line 131: Noting that in .NET Core 3.1 this is ignored for `CultureNotFoundException` - this would happen when `LANG` contains an invalid value for locale

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/InvokeDelegateBuilder.cs
normj: This logs only show up if `LAMBDA_RUNTIMESUPPORT_DEBUG` is set. They are just visible for developers that want to debug Amazon.Lambda.RuntimeSupport. They won't be visible to normal invokes unless the environment variable is set.

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/UserCodeInit.cs
normj: Probably very unlikely but like since the logging only shows up when debugging Amazon.Lambda.RuntimeSupport. The more logging when debugging this library the better is my opinion.

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/UserCodeInit.cs
normj: Yes, and this logging is just for RuntimeSupport logging and would let us trouble shoot if a customer noticed that the PreJitting isn't have the expected affect. Only shows up if `LAMBDA_RUNTIMESUPPORT_DEBUG` is enabled.

carlzogh (approved):

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/UserCodeInit.cs
normj: Fixed

PR #1011, opened 2021-12-05
Add X-Ray Cause JSON to error response in Amazon.Lambda.RuntimeSupport

normj:
*Description of changes:*
This PR ports the .NET Core 3.1 RIC `LambdaXRayExceptionWriter` to Amazon.Lambda.RuntimeSupport allowing the X-Ray cause section to be allowed to the X-Ray trace. The below image shows a snippet of the X-Ray cause in the X-Ray console.

image

Since Amazon.Lambda.RuntimeSupport uses the HTTP API and the X-Ray cause JSON is passed in the `Lambda-Runtime-Function-XRay-Error-Cause` HTTP headers no newlines can be part of the JSON document. So `LambdaXRayExceptionWriter` was tweaked from the .NET Core 3.1 RIC to make sure newlines were not added.

Amazon.Lambda.RuntimeSupport generates a HTTP client using NSwag and Lambda's OpenAPI doc for the Runtime API. The OpenAPI doc does not model `Lambda-Runtime-Function-XRay-Error-Cause` header. The generated HTTP client is partial class so I copied the existing error reporting method into a new `ErrorWithXRayCauseAsync` and added a few lines to add the X-Ray header.

### DIFF
--- a/Libraries/src/Amazon.Lambda.RuntimeSupport/Client/InternalClientAdapted.custom.cs
+++ b/Libraries/src/Amazon.Lambda.RuntimeSupport/Client/InternalClientAdapted.custom.cs
@@ -1,0 +1,176 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ * 
+ *  http://aws.amazon.com/apache2.0
+ * 
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+
+namespace Amazon.Lambda.RuntimeSupport
+{
+    internal partial interface IInternalRuntimeApiClient
+    {
+        /// <summary>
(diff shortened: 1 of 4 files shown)


Libraries/src/Amazon.Lambda.RuntimeSupport/Client/InternalClientAdapted.custom.cs
```diff
@@ -0,0 +1,176 @@
...
+        /// <summary>
+        /// This is a copy of the generated Error2Async method but adds support for the unmodeled header `Lambda-Runtime-Function-XRay-Error-Cause`.
+        /// </summary>
+        /// <param name="awsRequestId"></param>
+        /// <param name="lambda_Runtime_Function_Error_Type"></param>
+        /// <param name="errorJson"></param>
+        /// <param name="xrayCause"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public async System.Threading.Tasks.Task<SwaggerResponse<StatusResponse>> ErrorWithXRayCauseAsync(string awsRequestId, string lambda_Runtime_Function_Error_Type, string errorJson, string xrayCause, System.Threading.CancellationToken cancellationToken)
```
carlzogh, line 46: Could we possibly reuse/extend relevant parts of `InternalRuntimeApiClient.ErrorAsync()`? (ref. https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.RuntimeSupport/Client/InternalClientAdapted.cs#L109)

carlzogh (approved):

PR #1012, opened 2021-12-05
Replace LitJson with System.Text.Json

carlzogh:
*Description of changes:*

This change replaces usage of `LitJson` with `System.Text.Json`. As such, `Amazon.Lambda.RuntimeSupport/ThirdParty/Json` was removed and an assembly reference to `System.Text.Json` was added to `Amazon.Lambda.RuntimeSupport.csproj`


normj (approved):

normj:
I also made the changes to only use PackageReference for System.Text.Json for .NET Standard 2.0 which is the only target framework that needs it and update the `ErrorWithXRayCauseAsync` method to use System.Text.Json created in my previous PR.

PR #1013, opened 2021-12-05
Tighten (and tidy) up bootstrap.sh

carlzogh:
*Description of changes:*
1. Use `LAMBDA_TASK_ROOT` instead of new variable `USER_LAMBDA_BINARIES_DIR`
1. Look for specific .deps.json and .runtimeconfig.json files based on the assembly name instead of dynamically looking for all occurrences of `*.deps.json` and `*.runtimeconfig.json` in `LAMBDA_TASK_ROOT`
1. Add fallback to look for single occurrence of *.deps.json and *.runtimeconfig.json if not using the assembly name
1. Make exit codes unique across the bootstrap script
1. Rewrite conditionals when `LAMBDA_HANDLER` is not an executable assembly (contains `::`)

### DIFF
--- a/Libraries/src/Amazon.Lambda.RuntimeSupport/bootstrap.sh
+++ b/Libraries/src/Amazon.Lambda.RuntimeSupport/bootstrap.sh
@@ -5,10 +5,15 @@
 # These files are used to add the end-user assembly into context and make the code reachable to the dotnet process
 # Since the file names are not known in advance, we use this shell script to find the files and pass them to the dotnet process as parameters
 # You can improve cold-start performance by setting the LAMBDA_DOTNET_MAIN_ASSEMBLY environment variable and specifying the assembly name
-USER_LAMBDA_BINARIES_DIR="/var/task"
-if [ ! -d "${USER_LAMBDA_BINARIES_DIR}" ]; then
-  echo "Error: .NET binaries for Lambda function are not correctly installed in the ${USER_LAMBDA_BINARIES_DIR} directory of the image when the image was built. The ${USER_LAMBDA_BINARIES_DIR} directory is missing." 1>&2
-  exit 1
+# LAMBDA_TASK_ROOT is inherited from the Lambda execution environment/base image as "/var/task", but can be overridden for use in custom images.
+if [ -z "${LAMBDA_TASK_ROOT}" ]; then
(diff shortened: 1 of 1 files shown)


Libraries/src/Amazon.Lambda.RuntimeSupport/bootstrap.sh
```diff
@@ -44,24 +51,25 @@ if [[ $(expr index "${LAMBDA_HANDLER}" ":") == 0 ]]; then
...
-    if [[ "${ASSEMBLY_NAME}" == *.dll ]]; then
-      ASSEMBLY_NAME="${ASSEMBLY_NAME::-4}"
-    fi
-    DEPS_FILE="${USER_LAMBDA_BINARIES_DIR}/${ASSEMBLY_NAME}.deps.json"
-    RUNTIMECONFIG_FILE="${USER_LAMBDA_BINARIES_DIR}/${ASSEMBLY_NAME}.runtimeconfig.json"
+    ASSEMBLY_NAME="${LAMBDA_HANDLER::${HANDLER_COL_INDEX}-1}"
+  fi
+  
+  DEPS_FILE="${LAMBDA_TASK_ROOT}/${ASSEMBLY_NAME}.deps.json"
+  if ! [ -f "${DEPS_FILE}" ]; then
```
normj, line 65: This will not work if the function handler is referring to an assembly that is not the main assembly of the package bundle. This is a valid scenario in .NET Core 3.1 runtime and the previous `find` call allowed this to work.

We need the logic to do a fallback if this test fails to do a search in the top level directory for any other deps.json file and runtime.config.json file.

normj (changes requested):

Libraries/src/Amazon.Lambda.RuntimeSupport/bootstrap.sh
carlzogh: Thanks @normj  - added a fallback that matches the .NET Core 3.1 RIC where the bootstrap would look for exactly one occurrence of `*.deps.json` and/or `*.runtimeconfig.json` in `LAMBDA_TASK_ROOT` and fail otherwise.

normj (approved):
Besides review code I built the container image locally and ran through the fallback scenario that was missing as well as the normal scenarios.

PR #1017, opened 2021-12-11
Fix deadlock issue in Amazon.Lambda.RuntimeSupport while logging

normj:
*Description of changes:*
This PR fixes an issue when both Console.WriteX and ILambdaLogger are used for logging in different threads it can trigger a deadlock. 

## How was the issue detected?
This was detected with the ASP.NET Core blueprints which have both the Lambda logger and Console Logger added to the log provider. The ASP.NET Core console logger uses a background task to push log messages to the console. Initially the problem showed that the Lambda functions were timing out. After figuring out the issue was with logging I was able to reproduce this issue with a simple logging stress test which has been added to our integration tests.

## How does the deadlock happen?
For .NET 6 we introduced the `LogLevelLoggerWriter` which is where ILambdaLogger sends log message to format the log message and perform the log level determination. It also captures all Console.WriteX and Console.Error.WriteX to make sure they are formatted. 

To capture stdout and stderr `Console.SetOut` and `Console.SetError` are called passing in our own `TextWriter`. Our TextWriter wraps around the original TextWriters that `Console.Out` and `Console.Error` are using. That gives us a chance to format the console messages before forwarding on the log message to the original console `TextWriter`.

The TextWriters that `Console.Out` and `Console.Error` are set to our synchronized `TextWriters` meaning only one thread can make a call through them at a time. When Amazon.Lambda.RuntimeSupport sets new TextWriters using the `Console.SetOut` and `Console.SetError` those commands wrap around the passed in TextWriters with a synchronized TextWriter. This was the unexpected behavior that triggers the deadlock.

This means with a call to Console.WriteX 2 locks are acquired. Once for the synchronized TextWriter created to wrap around our TextWriter and then a second lock when calling to the original TextWriter Console.Out was set to. When using the logging commands from `ILambdaLogger` it is only acquiring the one lock for calling the original TextWriter Console.Out was set to. With `ILambdaLogger` skipping the first lock but Console.WriteX getting 2 locks this can trigger a deadlock writing to the stdout and stderror.

## How to fix the issue?
To fix the issue we need to make sure all code paths get the same locks. This fix makes sure before writing to the original TextWriter that our TextWriter is wrapping we have also lock on Console.Out or Console.Error just like the Console.WriteX commands do.

### DIFF
--- a/Libraries/src/Amazon.Lambda.AspNetCoreServer/Amazon.Lambda.AspNetCoreServer.csproj
+++ b/Libraries/src/Amazon.Lambda.AspNetCoreServer/Amazon.Lambda.AspNetCoreServer.csproj
@@ -25,17 +25,8 @@
     <ProjectReference Include="..\Amazon.Lambda.APIGatewayEvents\Amazon.Lambda.APIGatewayEvents.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.1'">
-    <PackageReference Include="Microsoft.AspNetCore" Version="2.0.0" />
-    <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="2.0.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="10.0.1" />
-    <ProjectReference Include="..\Amazon.Lambda.Serialization.Json\Amazon.Lambda.Serialization.Json.csproj" />
-  </ItemGroup>
-  <ItemGroup Condition="'$(TargetFramework)' != 'netcoreapp2.1'">
+  <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="System.IO.Pipelines" version="4.7.2">
-      <IncludeAssets>compile</IncludeAssets>
-    </PackageReference>
     <ProjectReference Include="..\Amazon.Lambda.Serialization.SystemTextJson\Amazon.Lambda.Serialization.SystemTextJson.csproj" />
   </ItemGroup>
 
--- a/Libraries/src/Amazon.Lambda.AspNetCoreServer/HostBuilderExtensions.cs
+++ b/Libraries/src/Amazon.Lambda.AspNetCoreServer/HostBuilderExtensions.cs
@@ -42,6 +42,7 @@ namespace Microsoft.Extensions.Hosting
                                 }
                                 else
                                 {
+                                    logging.ClearProviders();
                                     logging.AddLambdaLogger(hostingContext.Configuration, "Logging");
                                 }
                             })
--- a/Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/ConsoleLoggerWriter.cs
+++ b/Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/ConsoleLoggerWriter.cs
@@ -114,8 +114,13 @@ namespace Amazon.Lambda.RuntimeSupport.Helpers
         public LogLevelLoggerWriter()
             : this(Console.Out, Console.Error)
         {
+            // SetOut will wrap our WrapperTextWriter with a synchronized TextWriter. Pass in the new synchronized
+            // TextWriter into our writer to make sure we obtain a lock on that instance before writing to the stdout.
             Console.SetOut(_wrappedStdOutWriter);
+            _wrappedStdOutWriter.LockObject = Console.Out;
+
             Console.SetError(_wrappedStdErrorWriter);
+            _wrappedStdErrorWriter.LockObject = Console.Error;
 
             ConfigureLoggingActionField();
         }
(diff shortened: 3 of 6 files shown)


Libraries/src/Amazon.Lambda.AspNetCoreServer/Amazon.Lambda.AspNetCoreServer.csproj
```diff
@@ -25,17 +25,8 @@
     <ProjectReference Include="..\Amazon.Lambda.APIGatewayEvents\Amazon.Lambda.APIGatewayEvents.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.1'">
```
normj, line 28: This was just cleanup sections we don't need anymore. It was originally done when I was in the guessing stage of what was causing the bahavior.

Libraries/src/Amazon.Lambda.AspNetCoreServer/HostBuilderExtensions.cs
```diff
@@ -42,6 +42,7 @@ public static IHostBuilder ConfigureWebHostLambdaDefaults(this IHostBuilder buil
                                 }
                                 else
                                 {
+                                    logging.ClearProviders();
```
normj, line 45: Without this both our logger and the console logger where added which is what triggered the behavior. Now we clear the console logger. Not to work around the deadlock, that is fixed by the locking fix. We clear because it fixes another reported issue about double logs in CloudWatch Logs.

Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.IntegrationTests/BaseCustomRuntimeTest.cs
```diff
@@ -199,8 +199,15 @@ protected async Task UpdateHandlerAsync(IAmazonLambda lambdaClient, string handl
                 }
             };
             await lambdaClient.UpdateFunctionConfigurationAsync(updateFunctionConfigurationRequest);
+
             // Wait for eventual consistency of function change.
-            await Task.Delay(3000);
+            var getConfigurationRequest = new GetFunctionConfigurationRequest { FunctionName = FunctionName };
```
normj, line 204: This is added to make the tests more resilient.

Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.IntegrationTests/BaseCustomRuntimeTest.cs
```diff
@@ -217,7 +224,8 @@ protected async Task CreateFunctionAsync(IAmazonLambda lambdaClient, string buck
                 },
                 Handler = "PingAsync",
                 MemorySize = 512,
-                Runtime = Runtime.Provided,
+                Timeout = 30,
```
normj, line 227: I needed to increase the timeout to give enough to to prove I was not getting a deadlock

Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/CustomRuntimeFunctionTest/CustomRuntimeFunction.cs
```diff
@@ -97,6 +101,61 @@ private static async Task Main(string[] args)
             }
         }
 
+        private static Task<InvocationResponse> LoggingStressTest(InvocationRequest invocation)
```
normj, line 104: This test was failing with a deadlock before this PR fix.

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/ConsoleLoggerWriter.cs
```diff
@@ -217,31 +230,35 @@ internal void FormattedWriteLine(string message)
...
-                    if (!string.IsNullOrEmpty(displayLevel))
+                    if (_logFormatType == LogFormatType.Unformatted)
                     {
-                        line = $"{DateTime.UtcNow.ToString("yyyy-MM-ddTHH:mm:ss.fffZ")}\t{CurrentAwsRequestId}\t{displayLevel}\t{message ?? string.Empty}";
+                        _innerWriter.WriteLine(message);
                     }
                     else
                     {
-                        line = $"{DateTime.UtcNow.ToString("yyyy-MM-ddTHH:mm:ss.fffZ")}\t{CurrentAwsRequestId}\t{message ?? string.Empty}";
+                        string line;
```
normj, line 250: This is mostly whitespace changes due to the new indent of the lock. If you click hide whitespace changes you can see there is very little change.

boblodgett (approved):

krk (approved):
lgtm, thanks!

PR #1022, opened 2021-12-15
fix: Add GetRelativePath method to DirectoryManager class

96malhar:
*Description of changes:*
The `Amazon.Lambda.Annotations.SourceGenerator` project targets `netstandard2.0`

`netstandard2.0` does not have the `Path.GetRelativePath` method as part `System.IO`. It was introduced as part of `netstandard2.1`.

This PR adds the `GetRelativePath` as part of the `DirectoryManager` class. This method mimics the behavior of the [built-in API](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getrelativepath?view=net-6.0)

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/FileIO/DirectoryManager.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/FileIO/DirectoryManager.cs
@@ -1,4 +1,6 @@
+using System;
 using System.IO;
+using System.Text;
 
 namespace Amazon.Lambda.Annotations.SourceGenerator.FileIO
 {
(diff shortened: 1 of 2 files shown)


normj (approved):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/FileIO/DirectoryManager.cs
```diff
@@ -8,16 +10,64 @@ public class DirectoryManager : IDirectoryManager
 
         public bool Exists(string path) => Directory.Exists(path);
 
+        /// <summary>
+        /// This method mimics the behaviour of https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getrelativepath?view=net-6.0
+        /// </summary>
         public string GetRelativePath(string relativeTo, string path)
```
ganeshnj, line 16: Worth shimming the method as it is from System.IO?

https://github.com/dotnet/runtime/blob/09c1a1f7b0c477890b04912d8dd4f742f80faffc/src/libraries/System.Private.CoreLib/src/System/IO/Path.cs#L822

Along with providing clear attribution.

96malhar: I did think about this and discussed it with @normj. He suggested that we port the existing logic in the `dotnet-cli-extensions` to get the relative path.

Existing logic - https://github.com/aws/aws-extensions-for-dotnet-cli/blob/9639f8f4902349d289491b290979a3a1671cc0a5/src/Amazon.Common.DotNetCli.Tools/Utilities.cs#L91

@normj Do you want to reconsider porting the method verbatim (with citations) from System.IO?

ganeshnj: Even if we don't take System.IO change, add a reference to aws-extensions-for-dotnet-cli repo, so we are aware of the change has been taken straight from other place. Future updates will be easier.

96malhar: Fixed

ganeshnj (approved):

PR #1023, opened 2021-12-16
Update targeting to .NET 6

normj:
*Description of changes:*
Update Amazon.Lambda.Annotations to target .NET 6. Since we now have a .NET 6 container image update integ tests to use that. Fixed a issues discovered while doing end to end testing.
* Rename PackageTypeEnum to LambdaPackageType
* Fix issue with PackageTypeEnum being marked as nullable which is not allowed for named attribute parameters
* Fix issue with the enum conversion from attribute data to source generator model. Source generator was seeing the enum attribute data as just an int.
* Both Amazon.Lambda.Annotations and Amazon.Lambda.Annotations.SourceGenerator target netstandard2.0 which is required for source generators to work due to .NET Framework requirements in VS>
* Removed `Runtime` field in serverless.template file when using `Image` based packaging.
* Fix setting Image->Command in serverless.template to be an array which is required for CloudFormation.

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.nuspec
+++ b/Libraries/src/Amazon.Lambda.Annotations.nuspec
@@ -9,17 +9,17 @@
         <projectUrl>https://github.com/aws/aws-lambda-dotnet</projectUrl>
         <license type="expression">Apache-2.0</license>
         <dependencies>
-            <group targetFramework=".netcoreapp3.1"/>
+            <group targetFramework=".net6.0"/>
         </dependencies>
     </metadata>
 
     <files>
         <!-- Dependencies to make sure attributes are available in consuming csproj, this ensures packaged version of customer code have all the dependencies needed. -->
-        <file src="Amazon.Lambda.Annotations\bin\Release\netcoreapp3.1\Amazon.Lambda.Annotations.dll" target="lib/netcoreapp3.1" />
+        <file src="Amazon.Lambda.Annotations\bin\Release\netstandard2.0\Amazon.Lambda.Annotations.dll" target="lib/net6.0" />
 
         <!-- Include every dependency manually for analyzer, whenever a new dependency is added, it has to be added here. -->
-        <file src="Amazon.Lambda.Annotations.SourceGenerator\bin\Release\netcoreapp3.1\Amazon.Lambda.Annotations.dll" target="analyzers\dotnet\cs" />
-        <file src="Amazon.Lambda.Annotations.SourceGenerator\bin\Release\netcoreapp3.1\Amazon.Lambda.Annotations.SourceGenerator.dll" target="analyzers\dotnet\cs" />
-        <file src="Amazon.Lambda.Annotations.SourceGenerator\bin\Release\netcoreapp3.1\Newtonsoft.Json.dll" target="analyzers\dotnet\cs" />
+        <file src="Amazon.Lambda.Annotations.SourceGenerator\bin\Release\netstandard2.0\Amazon.Lambda.Annotations.dll" target="analyzers\dotnet\cs" />
+        <file src="Amazon.Lambda.Annotations.SourceGenerator\bin\Release\netstandard2.0\Amazon.Lambda.Annotations.SourceGenerator.dll" target="analyzers\dotnet\cs" />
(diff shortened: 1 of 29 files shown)


Libraries/src/Amazon.Lambda.Annotations.nuspec
```diff
@@ -9,17 +9,17 @@
...
+        <file src="Amazon.Lambda.Annotations\bin\Release\netstandard2.0\Amazon.Lambda.Annotations.dll" target="lib/net6.0" />
 
         <!-- Include every dependency manually for analyzer, whenever a new dependency is added, it has to be added here. -->
-        <file src="Amazon.Lambda.Annotations.SourceGenerator\bin\Release\netcoreapp3.1\Amazon.Lambda.Annotations.dll" target="analyzers\dotnet\cs" />
-        <file src="Amazon.Lambda.Annotations.SourceGenerator\bin\Release\netcoreapp3.1\Amazon.Lambda.Annotations.SourceGenerator.dll" target="analyzers\dotnet\cs" />
-        <file src="Amazon.Lambda.Annotations.SourceGenerator\bin\Release\netcoreapp3.1\Newtonsoft.Json.dll" target="analyzers\dotnet\cs" />
+        <file src="Amazon.Lambda.Annotations.SourceGenerator\bin\Release\netstandard2.0\Amazon.Lambda.Annotations.dll" target="analyzers\dotnet\cs" />
+        <file src="Amazon.Lambda.Annotations.SourceGenerator\bin\Release\netstandard2.0\Amazon.Lambda.Annotations.SourceGenerator.dll" target="analyzers\dotnet\cs" />
+        <file src="Amazon.Lambda.Annotations.SourceGenerator\bin\Release\netstandard2.0\Newtonsoft.Json.dll" target="analyzers\dotnet\cs" />
```
ganeshnj, line 23: Could you verify that pack contains all the DLLs specified here with a clean solution?

Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/Amazon.Lambda.Annotations.SourceGenerators.Tests.csproj
```diff
@@ -61,7 +61,7 @@
       <ProjectReference Include="..\..\src\Amazon.Lambda.Annotations.SourceGenerator\Amazon.Lambda.Annotations.SourceGenerator.csproj" />
       <ProjectReference Include="..\..\src\Amazon.Lambda.Annotations\Amazon.Lambda.Annotations.csproj" />
       <ProjectReference Include="..\..\src\Amazon.Lambda.APIGatewayEvents\Amazon.Lambda.APIGatewayEvents.csproj" />
-      <ProjectReference Include="..\..\src\Amazon.Lambda.Core\Amazon.Lambda.Core.csproj" />
+      <ProjectReference Include="..\..\src\Amazon.Lambda.Core\Amazon.Lambda.Core.csproj" AdditionalProperties="TargetFramework=netstandard2.0" />
```
ganeshnj, line 64: Please add a comment why AdditionalProperties is required here.

Libraries/test/TestServerlessApp/SimpleCalculator.cs
```diff
@@ -8,7 +8,7 @@
 using TestServerlessApp.Services;
 
 // Assembly attribute to enable the Lambda function's JSON input to be converted into a .NET class.
-[assembly: LambdaSerializer(typeof(Amazon.Lambda.Serialization.SystemTextJson.DefaultLambdaJsonSerializer))]
+//[assembly: LambdaSerializer(typeof(Amazon.Lambda.Serialization.SystemTextJson.DefaultLambdaJsonSerializer))]
```
ganeshnj, line 11: can this be removed completely? (even the comment)

Libraries/src/Amazon.Lambda.Annotations.nuspec
normj: I have verified. This is just switching back from a recent change to netcoreapp3.1 back to the original netstandard2.0.

Libraries/test/TestServerlessApp/SimpleCalculator.cs
normj: I have uncommented the line as it is required.

ganeshnj (approved):

96malhar (approved):

PR #1024, opened 2021-12-16
fix: return early if there are no Lambda functions to process

ganeshnj:
*Issue #, if available:*

*Description of changes:*

Currently in a project where there doesn't exist a Lambda function, the generator fails at writing the serverless.template due to missing information. This change allows to exit early if there doesn't exist a Lambda function and also output a info diagnostics to let customers know that they should add a `LambdaFunction` attribute in order to start consuming the generator.

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
@@ -46,6 +46,12 @@ namespace Amazon.Lambda.Annotations.SourceGenerator
                     return;
                 }
 
+                // If there are no Lambda methods, return early
+                if (!receiver.LambdaMethods.Any())
+                {
+                    return;
+                }
+
                 var semanticModelProvider = new SemanticModelProvider(context);
                 if (receiver.StartupClasses.Count > 1)
                 {


Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
```diff
@@ -46,6 +46,14 @@ public void Execute(GeneratorExecutionContext context)
                     return;
                 }
 
+                // If there are no Lambda methods, return early
+                if (!receiver.LambdaMethods.Any())
+                {
+                    // If there are more than one startup class, report them as errors
```
normj, line 52: This comment doesn't match what is being done. Also do we need to report anything. The message for `LambdaFunctionNotFound` doesn't sound like just info. 

The tone make its sound like something is wrong and they must take action. In my case since it was showing up in the test project that has a dependency on the Lambda project they shouldn't see anything.

normj (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
ganeshnj: I removed the info diag.

normj (approved):

96malhar (approved):

PR #1028, opened 2021-12-18
ci: (followup) remove git config for line endings

ganeshnj:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Libraries/test/TestServerlessApp/serverless.template
+++ b/Libraries/test/TestServerlessApp/serverless.template
@@ -2,6 +2,25 @@
   "AWSTemplateFormatVersion": "2010-09-09",
   "Transform": "AWS::Serverless-2016-10-31",
   "Description": "An AWS Serverless Application.",
+  "Parameters": {
+    "ArchitectureTypeParameter": {
+      "Type": "String",
+      "Default": "x86_64",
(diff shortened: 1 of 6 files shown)


normj (approved):

Libraries/test/TestServerlessApp/serverless.template
```diff
@@ -2,6 +2,25 @@
...
+  },
+  "Globals": {
+    "Function": {
+      "Architectures": [
+        {
+          "Ref": "ArchitectureTypeParameter"
+        }
+      ]
+    }
+  },
```
96malhar, line 23: Should we add these JSON blobs by default in the [CloudFormationJsonWriter.cs](https://github.com/aws/aws-lambda-dotnet/blob/feature/annotations/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationJsonWriter.cs)?

ganeshnj: this is feature we have to support at some point to later to able to configure the arch of the lambda functions. 

Current code change is just a workaround to be able to run the test app on both x86 and ARM.

96malhar (approved):

PR #1034, opened 2021-12-22
Automate ASP.NET Core version update

ganeshnj:
*Issue #, if available:*

*Description of changes:*

This PR sets up automation for updating the Dockerfiles with the latest version of ASP.NET Core version available for .NET 5 and .NET 6. This will allow us to push image updates quicker and reduce Ops load.

On a successful run, the workflow will create a PR similar to https://github.com/ganeshnj/aws-lambda-dotnet/pull/38 which further needs to be merged to push newer images. https://github.com/ganeshnj/aws-lambda-dotnet/pull/38 is for demo only, once this PR is merged, https://github.com/ganeshnj/aws-lambda-dotnet/pull/38 will be discarded and a new PR will be created.

Screenshot 
image

Testing performed

- [x] When there are no new versions available, actions succeeds without executing PR step (https://github.com/ganeshnj/aws-lambda-dotnet/runs/4622876309)
- [x] When a new version available, a PR https://github.com/ganeshnj/aws-lambda-dotnet/pull/38 is created

### DIFF
--- a/.github/workflows/update-Dockerfiles.yml
+++ b/.github/workflows/update-Dockerfiles.yml
@@ -1,0 +1,94 @@
+name: Update Dockerfiles
+
+on:
+  # Allows to run this workflow manually from the Actions tab
+  workflow_dispatch:
+    inputs:
+      NET_5_AMD64:
+        description: ".NET 5 AMD64"
+        type: boolean
+        required: true
+        default: "true"
+      NET_6_AMD64:
+        description: ".NET 6 AMD64"
+        type: boolean
+        required: true
+        default: "true"
+      NET_6_ARM64:
+        description: ".NET 6 ARM64"
+        type: boolean
+        required: true
+        default: "true"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    env:
+      NET_5_AMD64_Dockerfile: "LambdaRuntimeDockerfiles/Images/net5/amd64/Dockerfile"
+      NET_6_AMD64_Dockerfile: "LambdaRuntimeDockerfiles/Images/net6/amd64/Dockerfile"
+      NET_6_ARM64_Dockerfile: "LambdaRuntimeDockerfiles/Images/net6/arm64/Dockerfile"
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      - name: Update .NET 5 AMD64
+        id: update-net5-amd64
+        shell: pwsh
+        run: |
+          .\LambdaRuntimeDockerfiles/update-dockerfile.ps1 -Dockerfile ${{ env.NET_5_AMD64_Dockerfile }}
+        if: ${{ github.event.inputs.NET_5_AMD64 == 'true' }}
+
+      - name: Update .NET 6 AMD64
+        id: update-net6-amd64
+        shell: pwsh
+        run: |
+          .\LambdaRuntimeDockerfiles/update-dockerfile.ps1 -Dockerfile ${{ env.NET_6_AMD64_Dockerfile }}
+        if: ${{ github.event.inputs.NET_6_AMD64 == 'true' }}
+
+      - name: Update .NET 6 ARM64
+        id: update-net6-arm64
+        shell: pwsh
+        run: |
+          .\LambdaRuntimeDockerfiles/update-dockerfile.ps1 -Dockerfile ${{ env.NET_6_ARM64_Dockerfile }}
+        if: ${{ github.event.inputs.NET_6_ARM64 == 'true' }}
+
+      # Update Dockerfiles if newer version of ASP.NET Core is available
+      - name: Commit and Push
+        id: commit-push
+        shell: pwsh
+        run: |
+          git config --global user.email "github-aws-sdk-dotnet-automation@amazon.com"
+          git config --global user.name "aws-sdk-dotnet-automation"
+          $suffix=Get-Date -Format yyyy-mm-dd-HH-mm
+          $branch="chore/update-Dockerfiles-${suffix}"
+          git checkout -b $branch
+          git add "**/*Dockerfile"
+          git commit -m "chore: ASP.NET Core version update in Dockerfiles"
+          git push origin $branch
+          echo "::set-output name=BRANCH::$branch"
+
+      # Create a Pull Request from the pushed branch
+      - name: Pull Request
+        id: pull-request
+        if: ${{ steps.commit-push.outputs.BRANCH }}
+        uses: repo-sync/pull-request@v2
+        with:
+          source_branch: ${{ steps.commit-push.outputs.BRANCH }}
+          destination_branch: "master"
+          pr_title: 'chore: ASP.NET Core version update in Dockerfiles'
+          pr_body: "This PR updates the Dockerfiles to use the latest ASP.NET Core version.
+            Verify listed Dockerfiles that they have correct version and matching SHA512 checksum for ASP.NET Core runtime.
+            \n\nAll .NET versions https://dotnet.microsoft.com/en-us/download/dotnet
+            \n\n*Description of changes:*
+            \n${{ format
+                (
+                  '{0}\n{1}\n{2}',
+                  join(steps.update-net5-amd64.outputs.*, '\n'),
+                  join(steps.update-net6-amd64.outputs.*, '\n'),
+                  join(steps.update-net6-arm64.outputs.*, '\n')
+                )
+            }}"
+          github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


ganeshnj:
Most recent Update Dockerfile workflow is failing because there exists a PR already.

.github/workflows/update-Dockerfiles.yml
```diff
@@ -0,0 +1,56 @@
...
+
+      # Create a Pull Request from the pushed branch
+      - name: Pull Request
+        id: pull-request
+        uses: repo-sync/pull-request@v2
+        with:
+          source_branch: ${{ steps.commit-push.outputs.BRANCH }}
+          destination_branch: "master"
+          pr_title: 'chore: ASP.NET Core version update in Dockerfiles'
+          pr_body: "This PR updates the Dockerfiles to use the latest ASP.NET Core version."
```
normj, line 54: Add to the body of the PR links to where the reviewer should verify the SHA and version and instructions for how they should review this PR. That they should make sure the SHA and version number are correct.

LambdaRuntimeDockerfiles/update-dockerfile.ps1
```diff
@@ -0,0 +1,55 @@
...
+# Updates the Dockerfile with next ASP.NET Core version and sha256 hash if available
+function Update-Dockerfile ([string]$path) {
+    $nextVersion = Get-NextASPNETVersion -Dockerfile $path
+
+    $shaFilePath = "${nextVersion}-sha.txt"
+
+    Invoke-WebRequest -Uri "https://dotnetcli.blob.core.windows.net/dotnet/checksums/${nextVersion}-sha.txt" -OutFile $shaFilePath
+    if (!$?) {
+        Write-Host "Failed to download checksums for ${nextVersion}. ${nextVersion} is not available yet."
+        return
```
normj, line 12: Should it return a negative exit code so that the GitHub action fails if this fails to download?

normj (changes requested):

.github/workflows/update-Dockerfiles.yml
ganeshnj: Updated the description of the generated PR, check on #1036

LambdaRuntimeDockerfiles/update-dockerfile.ps1
ganeshnj: Actually not, it will be bad exit if returned non-zero here. The action checks for all the dockerfiles and then make sure there is git diff. 

In case (which will happen mostly) download for checksum fails, action will finish successfully without creating a PR. There are information logs written for debugging purpose.

Doing the way round is going to create a lot of failures which are not actually failures but unavailability of the newer version.

normj: My concern is if MS changes the locations of the checksum file or some other assumption made in this PS script how will we notice. Am I missing something?

normj (changes requested):

LambdaRuntimeDockerfiles/update-dockerfile.ps1
ganeshnj: You are right, if the location changes we will never get to know unless someone checks it manually which is rare occasion because we will get to know that there are no auto PRs in last 30 days but that is a lot of cognitive load.

Anyway, I removed the scheduled run and set to manual to keep this action just for updating rather than detecting a new version is available.

.github/workflows/update-Dockerfiles.yml
```diff
@@ -0,0 +1,62 @@
...
+      # Create a Pull Request from the pushed branch
+      - name: Pull Request
+        id: pull-request
+        if: ${{ steps.update-dockerfiles.outputs.BRANCH }}
+        uses: repo-sync/pull-request@v2
+        with:
+          source_branch: ${{ steps.update-dockerfiles.outputs.BRANCH }}
+          destination_branch: "master"
+          pr_title: 'chore: ASP.NET Core version update in Dockerfiles'
+          pr_body: "This PR updates the Dockerfiles to use the latest ASP.NET Core version. Verify listed Dockerfiles that they have correct version and matching SHA256 checksum for ASP.NET Core runtime. \n\nAll .NET versions https://dotnet.microsoft.com/en-us/download/dotnet\n\n*Description of changes:*\n${{join(steps.update-dockerfiles.outputs.*, '\n')}}"
```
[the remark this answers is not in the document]
ganeshnj: Fixed

normj (approved):

ppittle:
Is this process documented anywhere?  Anything indicating how/why this does what it does?

LambdaRuntimeDockerfiles/update-dockerfile.ps1
```diff
@@ -0,0 +1,72 @@
...
+    $currentVersion = $line.Split("=")[1]
+    Write-Host "Current ASPNET version: ${currentVersion}"
+
+    $nextVersion = Update-PatchVersion -Version $currentVersion
+    Write-Host "Next ASPNET version: ${nextVersion}"
+
+    return $nextVersion
+}
+
+# Returns the next path version of the given version
```
ppittle, line 61: nit: _next pat**c**h version_

ganeshnj:
> Is this process documented anywhere?  Anything indicating how/why this does what it does?

PS script code should be easy to follow what it does. Regarding why, this has been a manual process for quite some this, automating it will save some dev time.

LambdaRuntimeDockerfiles/update-dockerfile.ps1
```diff
@@ -0,0 +1,72 @@
+param ([Parameter(Mandatory)]$Dockerfile)
+
+# Updates the Dockerfile with next ASP.NET Core version and checksum256 hash if available
```
ppittle, line 3: nit: isn't this updating checksum512?

LambdaRuntimeDockerfiles/update-dockerfile.ps1
```diff
@@ -0,0 +1,72 @@
...
+function Update-Dockerfile ([string]$path) {
```
ppittle, line 4: Can you add some documentation on how this process works and what it can and can't do?

If I'm correctly reading `Update-Dockerfile` -> `Get-NextASPNETVersion` -> `Update-PatchVersion`, then won't this script break when Microsoft updates a Minor version?  Is that by design?

ppittle, line 4: Why does `Update-Dockerfile` name its parameter `$path`  when `$Dockerfile` is used everywhere else?

ganeshnj: fixed with consistently using DockerfilePath

ganeshnj: Added top-level docs about the process.

That's correct, it will break in case of a minor version but Microsoft will never (in theory they can) release a minor version based on the history of updates. If they do, then there will be a complete different conversion on supporting new minor version, may be a new dockerfile etc etc. In short, yes script is designed to update only patch version.

philasmar (approved):

LambdaRuntimeDockerfiles/update-dockerfile.ps1
ppittle: > Microsoft will never ... release a minor version based on the history of updates:

I think there's pretty strong historical evidence that Microsoft does release minor updates:

.NET Core 3.1
.NET Core 3.0
.NET Core 2.2
.NET Core 2.1
.NET Core 2.0
.NET Core 1.1
.NET Core 1.0
.NET Framework 4.8
.NET Framework 4.7
.NET Framework 4.6
.NET Framework 4.5
.NET Framework 4
.NET Framework 3.5
.NET Framework 3.0

LambdaRuntimeDockerfiles/update-dockerfile.ps1
```diff
@@ -0,0 +1,78 @@
+# This script allows to update Dockerfiles with next ASP.NET Core patch version.
+# It greps the current version from Dockerfile, increment the patch version and fetches checksum file from Microsoft server.
+# If checksum file is available, it will update the Dockerfile next version and its checksum.
+# NOTE: Microsoft only releases patch updates for runtimes, in case of a minor update the Dockerfile a manual update will be required.
```
ppittle, line 4: Can you tweak the documentation so it's very clear that this _guesses_ the next version,  It doesn't query Microsoft (ie https://versionsof.net/) for that version.

ppittle (commented):
Small request to update documentation

LambdaRuntimeDockerfiles/update-dockerfile.ps1
ganeshnj: Not after they release a stable version, like 3.1, 2.1 (we don't support 2.2). If we ever do, we have to make necessary change like mentioned, it will be a new dockerfile.

Issue #1033, opened 2021-12-22
Add SQSBatchResponse class


Dreamescaper:
I this article I can see you can return partial success for SQS batch via SQSBatchResponse class, but seems like it is not available for .NET. 
https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html

Is it possible to add it? Thanks!

Dreamescaper:
This request has no relation to AspNetCore, we don't even use it.

ashishdhingra:
PR https://github.com/aws/aws-lambda-dotnet/pull/1039 is merged. The update has been released in version `2.1.0` of the `Amazon.Lambda.SQSEvents`.

PR #1039, opened 2021-12-31
Adding SQSBatchResponse class to support partial success for SQS batches

jon-armen:
*Issue #, if available: #1033 

*Description of changes: Adding SQSBatchResponse class to support partial success for SQS batches.  See https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html

### DIFF
--- a/Libraries/src/Amazon.Lambda.SQSEvents/SQSBatchResponse.cs
+++ b/Libraries/src/Amazon.Lambda.SQSEvents/SQSBatchResponse.cs
@@ -1,0 +1,54 @@
+﻿using System.Collections.Generic;
+using System.Runtime.Serialization;
+
+namespace Amazon.Lambda.SQSEvents
+{
+    /// This class can be used as the return type for Lambda functions that have partially
+    /// succeeded by supplying a list of message IDs that have failed to process.
+    /// https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#services-sqs-batchfailurereporting
+    [DataContract]
+    public class SQSBatchResponse
+    {
+        /// <summary>
+        /// Creates a new instance of <see cref="SQSBatchResponse"/>
+        /// </summary>
+        public SQSBatchResponse()
+            : this(new List<BatchItemFailure>())
+        {
+        }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="SQSBatchResponse"/>
+        /// </summary>
+        /// <param name="batchItemFailures">A list of batch item failures</param>
+        public SQSBatchResponse(List<BatchItemFailure> batchItemFailures)
+        {
+            BatchItemFailures = batchItemFailures;
+        }
+
+        /// <summary>
+        /// Gets or sets the message failures within the batch failures
+        /// </summary>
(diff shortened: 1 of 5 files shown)


slang25:
I just started coding this 20 minutes ago, what are the odds 😆

Libraries/src/Amazon.Lambda.SQSEvents/SQSBatchResponse.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System.Collections.Generic;
+
+namespace Amazon.Lambda.SQSEvents
+{
+    /// <summary>
+    /// Simle Queue Service response
```
slang25, line 6:
```suggestion
    /// This class can be used as the return type for Lambda functions that have partially succeeded by supplying a list of message IDs that have failed to process.
    /// https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#services-sqs-batchfailurereporting
```

slang25: I'm not great with wording, but something like this

jon-armen: I think that works, I see my typo above.  I had kept with the general style on the SQSEvent class, but I always like more comments myself.

slang25 (approved):
This is a super useful addition, looks good

Libraries/test/EventsTests.Shared/EventTests.cs
```diff
@@ -754,6 +754,35 @@ private static void Handle(SQSEvent sqsEvent)
...
+#if NETCOREAPP3_1_OR_GREATER        
+        [InlineData(typeof(Amazon.Lambda.Serialization.SystemTextJson.LambdaJsonSerializer))]
+        [InlineData(typeof(Amazon.Lambda.Serialization.SystemTextJson.DefaultLambdaJsonSerializer))]
+#endif
+        public void SQSBatchResponseTest(Type serializerType)
+        {
+            var serializer = Activator.CreateInstance(serializerType) as ILambdaSerializer;
+            using (var fileStream = LoadJsonTestFile("sqs-response.json"))
+            {
+                var sqsBatchResponse = serializer.Deserialize<SQSBatchResponse>(fileStream);
```
normj, line 768: Since this is a response the test has to be written in reverse. Create a instance of `SQSBatchResponse` and then serialize it to JSON to and make sure the JSON is correct, especially the casing.

normj (changes requested):
Thanks for the PR! Just one small comment about the test.

Libraries/test/EventsTests.Shared/EventTests.cs
jon-armen: I've updated the test following patterns observed in other tests.

normj (approved):

ganeshnj (approved):

normj:
Thanks again for the PR @jon-armen. The update has been released in version 2.1.0 of the Amazon.Lambda.SQSEvents packages. Also thanks @slang25 for help reviewing the PR.

PR #1043, opened 2022-01-05
Update tests to use System.Text.Json over Newtonsoft.Json

TobbenTM:
*Description of changes:*

This PR updates the default tests in blueprints to use System.Text.Json for the samples instead of Newtonsoft.Json.

While the Newtonsoft.Json package might work for most things, it'll fail as soon as you try testing functions with authorizer claims, since the marshalling is expecting the Claims property to be populated: https://github.com/aws/aws-lambda-dotnet/blob/baa10d4788f339bd35137ea1b5c01df0a5b82afc/Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs#L125

This is because the Claims getter is looking explicitly for a `JsonElement` claims object, it'll fail if you pass a Newtonsoft.Json deserialized payload as that will have a `JObject` type there instead: https://github.com/aws/aws-lambda-dotnet/blob/087590ce99274e16e26d37e1dfd73b0b71d1230a/Libraries/src/Amazon.Lambda.APIGatewayEvents/APIGatewayCustomAuthorizerContext.cs#L136

Additionally, I expanded the current test in `TestCallingWebAPI.cs` to validate this issue. When using Newtonsoft.Json on the `TestGetCustomAuthorizerValue` test, it'll fail with an empty `Claims` dictionary, even if it exists in the sample request.

I did not add directives to support `NETSTANDARD_2_0` since the test project as a whole does not have a configuration for that.

### DIFF
--- a/Blueprints/BlueprintDefinitions/vs2017/AspNetCoreWebAPI/template/test/BlueprintBaseName.1.Tests/S3ProxyControllerTests.cs
+++ b/Blueprints/BlueprintDefinitions/vs2017/AspNetCoreWebAPI/template/test/BlueprintBaseName.1.Tests/S3ProxyControllerTests.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.Json;
 using System.Threading.Tasks;
 
 using Xunit;
@@ -10,7 +11,6 @@ using Amazon.Lambda.TestUtilities;
 using Amazon.Lambda.APIGatewayEvents;
 
 using Microsoft.Extensions.Configuration;
-using Newtonsoft.Json;
 
 using Amazon;
 using Amazon.S3;
@@ -54,7 +54,10 @@ namespace BlueprintBaseName._1.Tests
 
             // Use sample API Gateway request that uploads an object with object key "foo.txt" and content of "Hello World".
             var requestStr = File.ReadAllText("./SampleRequests/S3ProxyController-Put.json");
-            var request = JsonConvert.DeserializeObject<APIGatewayProxyRequest>(requestStr);
+            var request = JsonSerializer.Deserialize<APIGatewayProxyRequest>(requestStr, new JsonSerializerOptions
+            {
+                PropertyNameCaseInsensitive = true
+            });
             var context = new TestLambdaContext();
             var response = await lambdaFunction.FunctionHandlerAsync(request, context);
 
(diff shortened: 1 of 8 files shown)


Blueprints/BlueprintDefinitions/vs2017/AspNetCoreWebAPI/template/test/BlueprintBaseName.1.Tests/S3ProxyControllerTests.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.Json;
```
normj, line 5: The blueprints in VS2017 folder target .NET Core 2.1 which doesn't have System.Text.Json. I suggest leaving the 2017 blueprints alone.

Blueprints/BlueprintDefinitions/vs2017/AspNetCoreWebAPI/template/test/BlueprintBaseName.1.Tests/ValuesControllerTests.cs
```diff
@@ -2,15 +2,14 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.Json;
```
normj, line 5: Same comment as above. Don't change the VS 2017 blueprints.

normj (changes requested):

Blueprints/BlueprintDefinitions/vs2017/AspNetCoreWebAPI/template/test/BlueprintBaseName.1.Tests/S3ProxyControllerTests.cs
TobbenTM: Okay, reverted those changes now 👍

Blueprints/BlueprintDefinitions/vs2017/AspNetCoreWebAPI/template/test/BlueprintBaseName.1.Tests/ValuesControllerTests.cs
TobbenTM: Reverted.

philasmar (approved):

normj (approved):

normj:
Thank you for the PR

Issue #1053, opened 2022-01-19
Amazon.Lambda.Annotations.HttpMethod conflicts with System.Net.Http.HttpMethod when using implicit using statements


martincostello:
## Describe the Feature

If the Implicit Using Statements feature of the .NET 6 SDK is enabled in a project using Amazon.Lambda.Annotations, any attempt to use the `[RestApi]` attribute will result in a compiler error due to ambiguity caused by the default inclusion of the `System.Net.Http` namespace as a using statement in all C# files in the project.

```sh
Error	CS0104	'HttpMethod' is an ambiguous reference between 'Amazon.Lambda.Annotations.HttpMethod' and 'System.Net.Http.HttpMethod'
```

This creates friction to the developer experience when using the annotations library, and may confuse less experienced developers, particularly those new to .NET and adopting it from .NET 6.

## Proposed Solution

Suggestions include:

* Rename the type to remove the ambiguity.
* Automatically add the alias below if implicit using statements are enabled and the package is referenced.
* Automatically add the alias to any Lambda templates using the library in the .NET CLI for new project creation.
* Document how to resolve the ambiguity.

### Describe alternatives you've considered

Manually resolve the ambiguity using an alias in relevant code files:

```csharp
using HttpMethod = Amazon.Lambda.Annotations.HttpMethod;
```

Manually add an additional implicit using statement to apply an alias in all C# files to resolve the ambiguity.

```xml
<ItemGroup>
  <Using Include="Amazon.Lambda.Annotations.HttpMethod" Alias="HttpMethod" />
</ItemGroup>
```

### Environment

* [ ] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change  

---

This is a :rocket: Feature Request

normj:
Agreed the implicit namespaces in .NET 6 makes this awkward. I pushed a breaking PR to rename the enum to `LambdaHttpMethod`. I'm open to better names. https://github.com/aws/aws-lambda-dotnet/pull/1056

ganeshnj:
https://www.nuget.org/packages/Amazon.Lambda.Annotations/0.4.3-preview is out containing the fix.

PR #1056, opened 2022-01-19
Rename HttpMethod to LambdaHttpMethod

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/1053

*Description of changes:*
Rename the `HttpMethod` enum to `LambdaHttpMethod` to avoid ambigituty with `System.Net.Http.HttpMethod`. This is especially important for .NET 6 which has implicit usings for `System.Net.Http`


ganeshnj (approved):

ganeshnj:
Fine with renaming, I was also thinking of scoping the types using nested types? It wouldn't matter a lot but we can make it as a convention for conflicting types.

PR #1063, opened 2022-02-01
Minor performance improvements to Amazon.Lambda.RuntimeSupport

normj:
*Description of changes:*

Minor tweaks while profiling Amazon.Lambda.RuntimeSupport to reduce time spent in this package during coldstarts. I would have liked to do more style/whitespace cleanup to `InternalClientAdapted` but I don't want the logic changes to be unnoticeable in a file with all lines changed.

* Added additional internal logging statements to help with understanding where time is being spent. These logging statements are a noop unless the `LAMBDA_RUNTIMESUPPORT_DEBUG` environment variable is set.
* Changes in `InternalClientAdapted`
  * Use source generator JSON serialization when targeting .NET 6
  * Remove code treating HTTP status code as strings
  * Skipped deserialize response when sending response since all we need is the HTTP status code
  *  Got rid of the `Error2Async` method since this was replaced in previous PR with `ErrorWithXRayCauseAsync`
  * Consolidated the InternalClientAdapted.cs and InternalClientAdapted.custom.cs since are not longer depending on this file being generated.
* Added `AWS_LAMBDA_DOTNET_DEBUG_RUN_ONCE` environment variable for local profiling so process terminates cleanly after processing a single event. This environment variable should never be set when deployed to Lambda.
* For .NET 6 directly create the `SocketsHttpHandler` for the HttpClient to avoid the HttpClient creating an additional wrapper handler `HttpClientHandler`.
* Remove lingering ties to NSwag generator since we have customized `InternalClientAdapted`.


Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/LambdaBootstrap.cs
```diff
@@ -115,20 +116,29 @@ public async Task RunAsync(CancellationToken cancellationToken = default(Cancell
...
             bool doStartInvokeLoop = _initializer == null || await InitializeAsync();
 
             while (doStartInvokeLoop && !cancellationToken.IsCancellationRequested)
             {
                 try
                 {
                     await InvokeOnceAsync(cancellationToken);
+                    if(runOnce)
+                    {
+                        return;
```
ganeshnj, line 140: We should do some logging that Lambda is configured for run once and process is exiting for diagnostics.

normj: Good idea, I have added the messaging.

ganeshnj (approved):

carlzogh (approved):

PR #1065, opened 2022-02-02
Add new page for testing executable assembly Lambda functions

normj:
*Description of changes:*

## What is an executable assembly?
.NET Lambda functions deployed using Custom Runtimes or in .NET 6 using Top Level statements are deployed as executable assemblies. That means the Lambda function handler string is not used by the Lambda runtime client to reflect on the method to call for each event. Instead the provided assembly is executed and during the startup of the assembly it initializes `Amazon.Lambda.RuntimeSupport` as the Lambda runtime client and indicates what method to call for each method. In this programming model the .NET Lambda Test Tool can not locally debug .NET Lambda functions in the test page.

## What is being changed?
To support .NET Lambda functions as executable assemblies the Test Tool will mock the Lambda Runtime API that the Lambda runtime client running in the executable assembly will connect to. Developers can locally launch their executable assemblies to point at the test tool. In the test tool itself users can queue events that will be sent the Lambda function via the Lambda Runtime API.

## Code changes summary:
There are 3 main code file changes
* **Controllers\RuntimeApiController** - Emulates the Lambda Runtime API as defined here https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html
* **Pages\Runtime.razor** - UI for queuing events that will dispatched through Controllers\RuntimeApiController and can view the status of events.
* **Services\RuntimeApiDataStore.cs** - Used as the datastore for events sharing information between the UI and the controller.

There was also a some rework for how the tool is started up to make it easier for unit testing.

## Walkthrough:

After launching the test tool navigate to the **Executable Assembly** page. On this page you can queue events to be processed.

image

Here is an example of a .NET 6 Lambda function using top level statements. The `LambdaBootstrapBuilder` is used to create the Lambda runtime client which will reach out to the Lambda runtime API for events to process.

```csharp
using Amazon.Lambda.Core;
using Amazon.Lambda.RuntimeSupport;
using Amazon.Lambda.Serialization.SystemTextJson;

var handler = (string input, ILambdaContext context) =>
{
    return new Casing(input?.ToLower(), input?.ToUpper());
};

await LambdaBootstrapBuilder.Create<string, Casing>(handler, new DefaultLambdaJsonSerializer())
    .Build()
    .RunAsync();

public record Casing(string Lower, string Upper);
```

To have the Lambda function connect to the test tool's implementation of the Lambda Runtime API set the **AWS_LAMBDA_RUNTIME_API** environment variable to **localhost:<port-of-testtool>**. In Visual Studio this is done by editing the launchSettings.json file. Here is a full launchSettings.json example.
```
{
  "profiles": {
    "Runtime API": {
      "commandName": "Project",
      "environmentVariables": {
        "AWS_LAMBDA_RUNTIME_API": "localhost:5050"
      }
    }
  }
}
````

With the launchSettings.json file configured push **F5** in Visual Studio to debug the Lambda function and the events that we have queued up will start to be processed.

image

Here you can see one event was successfully processed and one event is was failed to processed. The links can be clicked to view the full text of the fields.

Note the Lambda function will still be running polling for events till it is forcibly shutdown. More events can be queued in the test tool and existing events can be requeued.

### DIFF
--- a/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Services/RuntimeApiDataStore.cs
+++ b/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Services/RuntimeApiDataStore.cs
@@ -97,7 +97,7 @@ namespace Amazon.Lambda.TestTool.BlazorTester.Services
                 Monitor.Enter(_lock);
                 try
                 {
-                    return new ReadOnlyCollection<EventContainer>(_queuedEvents); 
+                    return new ReadOnlyCollection<EventContainer>(_queuedEvents.ToArray()); 
                 }
                 finally
                 {
@@ -113,7 +113,7 @@ namespace Amazon.Lambda.TestTool.BlazorTester.Services
                 Monitor.Enter(_lock);
                 try
                 {
-                    return new ReadOnlyCollection<EventContainer>(_executedEvents);
+                    return new ReadOnlyCollection<EventContainer>(_executedEvents.ToArray());
                 }
                 finally
                 {
--- a/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Services/RuntimeApiDataStore.cs
+++ b/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Services/RuntimeApiDataStore.cs
@@ -43,15 +43,9 @@ namespace Amazon.Lambda.TestTool.BlazorTester.Services
         public EventContainer QueueEvent(string eventBody)
         {
             var evnt = new EventContainer(this, _eventCounter++, eventBody);
-            Monitor.Enter(_lock);
-            try
+            lock (_lock)
             {
                 _queuedEvents.Add(evnt);
-                Monitor.PulseAll(_lock);
-            }
-            finally
-            {
-                Monitor.Exit(_lock);
             }
             
             RaiseStateChanged();
@@ -61,8 +55,7 @@ namespace Amazon.Lambda.TestTool.BlazorTester.Services
         public bool TryActivateEvent(out IEventContainer activeEvent)
         {
             activeEvent = null;
-            Monitor.Enter(_lock);
-            try
+            lock(_lock)
             {
                 if (!_queuedEvents.Any())
                 {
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Services/RuntimeApiDataStore.cs [hunk 1]
```diff
@@ -0,0 +1,322 @@
...
+        public IEventContainer ActiveEvent { get; private set; }
+
+        public IReadOnlyList<IEventContainer> QueuedEvents
+        {
+            get
+            {
+                Monitor.Enter(_lock);
+                try
+                {
+                    return new ReadOnlyCollection<EventContainer>(_queuedEvents);
```
boblodgett, line 100: All the places where you are creating a ReadOnlyCollection protected by the lock doesn't seem to be doing what you intend because ReadOnlyCollection constructor doesn't copy the passed collection. So as _queuedEvents is modified so will be the collection the ReadOnlyCollection is wrapping. If you didn't care about the events being added to _queuedEvents under the hood then you wouldn't need the lock.

Did you intend something like this?

`return new ReadOnlyCollection<EventContainer>(_queuedEvents.ToArray());`

normj: Added the `ToArray()`

boblodgett (approved):

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Services/RuntimeApiDataStore.cs
```diff
@@ -0,0 +1,322 @@
...
+        private IList<EventContainer> _executedEvents = new List<EventContainer>();
+        private int _eventCounter = 1;
+        private object _lock = new object();
+        
+        public event EventHandler StateChange;
+        
+        public EventContainer QueueEvent(string eventBody)
+        {
+            var evnt = new EventContainer(this, _eventCounter++, eventBody);
+            Monitor.Enter(_lock);
```
ganeshnj, line 46: Curious on knowing usage of `Monitor` over `lock`. 

I always though Monitor is low level construct while `lock` is abstraction which handles the cirtical section release based on the code block which is better as developer don't have to worry about release/exit.
```csharp
lock(_obc)
{
}
```

normj: Switched to using `lock`. I believe an earlier iteration of this class involved async operations used within the lock so I couldn't use `lock` in that case.

ganeshnj (approved):

napalm684:
Can anyone confirm a successful setup with this new executable assembly approach in JetBrains Rider? I have been unable to even by following the setup steps which appear to be dated now.

PR #1069, opened 2022-02-02
Update README.md

coultonluke:
Corrected Lambda Entrypoint. Should be the same as the class name.

*Issue #, if available:*

*Description of changes:*
I believe the documentation was incorrect. The Lambda Handler mentioned in the readme should match the class name which is currently does not


normj (approved):
Thanks for the correction!

PR #1073, opened 2022-02-03
feat: Add POCOs for S3 object events

### LINKED ISSUE #1072: S3 Events with EventBridge
I have configured my S3 bucket to send events using EventBridge as described in [this recently released feature](https://aws.amazon.com/blogs/aws/new-use-amazon-s3-event-notifications-with-amazon-eventbridge/).  Below is an example JSON payload that a Lambda Function will recieve.

```json
{
  "version": "0",
  "id": "17793124-05d4-b198-2fde-7ededc63b103",
  "detail-type": "Object Created",
  "source": "aws.s3",
  "account": "123456789012",
  "time": "2021-11-12T00:00:00Z",
  "region": "ca-central-1",
  "resources": ["arn:aws:s3:::example-bucket"],
  "detail": {
    "version": "0",
    "bucket": {
      "name": "example-bucket"
    },
    "object": {
      "key": "example-key",
      "size": 5,
      "etag": "b1946ac92492d2347c6235b4d2611184",
      "version-id": "IYV3p45BT0ac8hjHg1houSdS1a.Mro8e",
      "sequencer": "00617F08299329D189"
    },
    "request-id": "N4N7GDK58NMKJ12R",
    "requester": "123456789012",
    "source-ip-address": "1.2.3.4",
    "reason": "PutObject"
  }
}
```

I was looking into using the [S3Events package](https://github.com/aws/aws-lambda-dotnet/tree/master/Libraries/src/Amazon.Lambda.S3Events) but classes comprising the [S3EventNotification](https://github.com/aws/aws-sdk-net/blob/17c412b750f36c5659a86141f8764352114f1451/sdk/src/Services/S3/Custom/Util/S3EventNotification.cs) model do not _quite_ line up with this JSON object; specifically the nested `Records` array.

Am I overlooking something or does there need to be a different Lambda event model for this new event type?

rpf3:
This commit adds POCOs to model S3 object events routed via EventBridge.

closes #1072

### DIFF
--- a/Libraries/src/Amazon.Lambda.CloudWatchEvents/S3Events/Bucket.cs
+++ b/Libraries/src/Amazon.Lambda.CloudWatchEvents/S3Events/Bucket.cs
@@ -1,0 +1,13 @@
+namespace Amazon.Lambda.CloudWatchEvents.S3Events
+{
+    /// <summary>
+    /// This class represents an S3 bucket.
+    /// </summary>
+    public class Bucket
+    {
+        /// <summary>
+        /// The name of the bucket.
+        /// </summary>
+        public string Name { get; set; }
+    }
+}
--- a/Libraries/src/Amazon.Lambda.CloudWatchEvents/S3Events/S3Object.cs
+++ b/Libraries/src/Amazon.Lambda.CloudWatchEvents/S3Events/S3Object.cs
@@ -1,5 +1,3 @@
-using System.Text.Json.Serialization;
-
 namespace Amazon.Lambda.CloudWatchEvents.S3Events
 {
     /// <summary>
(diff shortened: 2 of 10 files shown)


ashishdhingra:
@rpf3 It would be nice to add some test cases as well. Please refer existing ones as a reference.

ashishdhingra:
@rpf3 Also refer https://docs.aws.amazon.com/AmazonS3/latest/userguide//ev-events.html, `detail` could have additional details based on operation. Few examples:
- `deletion-type` for delete operation
- `restore-expiry-time` and `source-storage-class` for object restore completed.

`S3ObjectEventDetails` class perhaps needs to be modeled accordingly, perhaps might want to create subclasses for each scenario.

ashishdhingra:
Stale pr message

rpf3:
@ashishdhingra I've pushed some new models as you requested.  Is this more in line with what you were thinking?

ashishdhingra:
> @ashishdhingra I've pushed some new models as you requested. Is this more in line with what you were thinking?

@rpf3 Thanks. Could you also add some unit tests as well as requested in https://github.com/aws/aws-lambda-dotnet/pull/1073#issuecomment-1050236089.

normj:
I manually added unit tests, change size to be a long from an int and added `DataContract` attributes to fix serialization with Newtonsoft. I'll work on getting this PR released soon.

normj:
This PR has been released as part of version 4.1.0 of Amazon.Lambda.CloudWatchEvents. Thanks for the PR!

Issue #1087, opened 2022-02-24
Support async function initialization with the .NET 6 managed runtime


martincostello:
## Describe the Feature

Add support to `Amazon.Lambda.RuntimeSupport` for asynchronous initialization of .NET functions using the `dotnetcore3.1` or `dotnet6` managed runtimes.

### Is your Feature Request related to a problem?

While investigating moving a .NET Lambda using a custom runtime with Amazon.Lambda.RuntimeSupport to use .NET 6 over to the new built-in managed runtime for .NET 6, I noticed that the managed until uses the same infrastructure to bootstrap the user code from the handler type given in the function configuration.

However, there is no support for _asynchronously_ initializing such a function using the runtime in this way, which is possible with a custom runtime.

https://github.com/aws/aws-lambda-dotnet/blob/e29ba747e4f912be57811f20511e5a57e195fda7/Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/UserCodeInitializer.cs#L44-L48

This means that user code wishing to perform asynchronous initialization must perform it on the path of the first request, moving the "penalty" from the initialization code path to the handler code path. This also means that if initialization fails, the handler will attempt to process requests (and fail), rather than hard-fail due to initialization of the lambda itself failing.

## Proposed Solution

Possible ways this could be achieved:

- An attribute that specifies a bootstrap type/method that is invoked if found.
- An interface that defines an async initialization method that is invoked if the user-specified handler function implements the interface.

If I recall correctly, the in-development annotations library has similar support planned for it.

### Describe alternatives you've considered

None, other than staying on a custom runtime.

## Additional Context

This was found in the process of migrating a .NET 6 Lambda function with a custom runtime to test out the new managed runtime for .NET 6 (`dotnet6`).

### Environment

* [x] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change  

---

This is a :rocket: Feature Request

ashishdhingra:
@martincostello Thanks for submitting feature request. Feel free to contribute in the form of PR that could be reviewed by the team.

martincostello:
@ashishdhingra Any preference to either of the two suggested solutions (or something different entirely) before I start on something?

normj:
For now I would do the async work in the constructor of the type and block on the async work. At that point there is only a single thread going through so there should be no issues blocking on the async work. Plus you should take advantage of the init stage which is free although the init stage does have to complete in some time limit.

[...]

PR #1091, opened 2022-02-25
Support async function initialization with the .NET 6 managed runtime

martincostello:
*Issue #1087*

*Description of changes:*

Adds support for asynchronous initialization of classes that implement handlers for lambdas by introducing a new `IHandlerInitializer` interface (the specific name is open for discussion).

If an instance-based class (i.e. not a static class) used for a handler implements the `IHandlerInitializer` interface, it is invoked during the initialization of user code. If the method throws an exception or returns `false` (the need to return a boolean is also open for discussion), then an exception is thrown and the Lambda runtime would report the initialization failure. Otherwise if the method returns `true`, then the code is considered to be initialized and ready to process requests.

If the handler class is static or does not implement this interface, execution is the same as it is today.

Using the interface allows the user to opt-in to asynchronous initialization in an idiomatic way that integrates with the existing asynchronous initialization support for custom runtimes without the need to block on tasks in a class constructor.


normj:
The challenge for this is a Lambda function using managed runtime doesn't have a direct dependency on `Amazon.Lambda.RuntimeSupport`. The alternative would be to put `IHandlerInitializer` in `Amazon.Lambda.Core` but that also has problems because the user brings the version of `Amazon.Lambda.Core`. If a user brings in an old version that doesn't have `IHandlerInitializer` then `Amazon.Lambda.RuntimeSupport` blows up with a missing type exception. That is why the new logging methods added for .NET 6 were added to the logger interface with a default implementation.

I think the safe way to add this is lookup the `InitializeAsync` method by convention just like we look for the handler method instead of relying on an interface. 

Another concern is we would be making one more reflection call during cold start, so we need to make sure that does not add any significant performance hit when there isn't a `InitializeAsync` at the lowest memory setting.

Also the boolean seems unnecessary. If the task returned has a failed status we know initialization failed.

martincostello:
I've changed this to remove the interface, and instead invoke a method with either of the following two signatures if found on the handler's type:

* `public static Task InitializeAsync()`
* `public Task InitializeAsync()`

There's a few open questions marked with `// TODO` comments.

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/UserCodeLoader.cs
```diff
@@ -406,5 +428,35 @@ private object GetCustomerObject(Type customerType)
 
             return Activator.CreateInstance(customerType);
         }
+
+        private MethodInfo GetCustomerInitializer(Type customerType)
+        {
+            // TODO Should the method require an environment variable to opt-in/out?
```
normj, line 434: I'm very torn on this. The environment variable would avoid any potential conflict if somebody already has an `InitializeAsync` plus you could configure your own name. But the discoverability would be really low.

Wonder if we made the name more unique like `LambdaInitializeAsync` that would avoid the possible naming collision.

@ganeshnj Do you have any thoughts on this?

ganeshnj: There has to a way to avoid calling any customer method that might have same signature. I can't think of anything other than env var as of now.

I would prefer `InitializeLambdaAsync` or `LambdaInitializeAsync` which provide better context that this method not meant to be called in customer code (customer can if they want but not recommended) and will be executed as part of init phase.

normj:
I'm going to close for lack of activity.

Rayzbam:
InitializeAsync would still be a nice shot ..

PR #1096, opened 2022-03-02
Fix error message when .NET SDK found that is below .NET 6 requirement

Otimie:
*Issue #, if available:*
Incorrectly displays .NET Core 3.1 SDK as minimum supported version when it is actually .NET 6 SDK

*Description of changes:*
Change error message to reflect .NET 6 SDK is minimum supported version instead of .NET Core 3.1 SDK


normj (approved):
Thanks you!

Issue #1124, opened 2022-03-26
Amazon.Lambda.TestTool-6.0 serialization exception on Lambda's return value when using JSON serializer source generators


jksware:
## Description

An exception is thrown when trying to convert the result object of the function handler back to a stream in the Mock Lambda Test Tool on .Net 6 when using JSON source generators to serialize the result object.

This happens with types `T` for both `T` and `Task<T>` result types of the function handler, such as `APIGatewayHttpApiV2ProxyResponse` and `Task<APIGatewayHttpApiV2ProxyResponse>`, respectively.

An example of an exception thrown when the function handler's result type is `APIGatewayHttpApiV2ProxyResponse`:
 ```
Amazon.Lambda.Serialization.SystemTextJson.JsonSerializerException: Error converting the response object of type System.Object from the Lambda function to JSON: No JsonTypeInfo registered in AWSLambdaTest.HttpApiJsonSerializerContext for type System.Object.
```

## Reproduction Steps

From Visual Studio 2022, run the AWS .NET Core 6.0 Mock Lambda Test Tool with the following example code (a failing case for a non-async function handler). In Test Function select example request `API Gateway V2 HTTP API` and click Execute Function.

```
using Amazon.Lambda.APIGatewayEvents;
using Amazon.Lambda.Core;
using Amazon.Lambda.Serialization.SystemTextJson;
using System.Text.Json.Serialization;

// Assembly attribute to enable the Lambda function's JSON input to be converted into a .NET class.
[assembly: LambdaSerializer(typeof(SourceGeneratorLambdaJsonSerializer<AWSLambdaTest.HttpApiJsonSerializerContext>))]

namespace AWSLambdaTest;

[JsonSerializable(typeof(APIGatewayHttpApiV2ProxyRequest))]
[JsonSerializable(typeof(APIGatewayHttpApiV2ProxyResponse))]
public partial class HttpApiJsonSerializerContext : JsonSerializerContext
{
}

public partial class Function
{
    public APIGatewayHttpApiV2ProxyResponse Get(APIGatewayHttpApiV2ProxyRequest request, ILambdaContext context)
    {
        return new APIGatewayHttpApiV2ProxyResponse
        {
            Body = request.Body.ToUpper(),
            StatusCode = 200
        };
    }
}
```

Expected response:
```
{"statusCode":200,"body":"HELLO FROM LAMBDA","isBase64Encoded":false}
```

Actual response:
```
[...]

PR #1125, opened 2022-03-26
Fixes serializer exception when using source generators in JSON serializers

jksware:
in Amazon.Lambda.TestTool-6.0

*Fixes Issue #1124*

*Description of changes:*
Removes two direct invocations of the serializer with `System.Object` as the type parameter (for both async and non-async result type cases). Invocations are replaced by calling the correct specialization of the generic serializer via reflection, passing the runtime type of the result object as the first type parameter of the serializer method.

### DIFF
--- a/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LambdaExecutor.cs
+++ b/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LambdaExecutor.cs
@@ -151,7 +151,7 @@ namespace Amazon.Lambda.TestTool.Runtime
                         else
                         {
                             lambdaReturnStream = new MemoryStream();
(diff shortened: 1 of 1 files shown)


ashishdhingra:
@jksware Thanks for submitting the PR and apologies for delay. The fix implemented in the PR is already implemented in the following PR(s):
- https://github.com/aws/aws-lambda-dotnet/pull/1326
- https://github.com/aws/aws-lambda-dotnet/pull/1331

Could you please verify this in the latest version of Lambda Test Tool and confirm if this PR could be closed?

Thanks,
Ashish

jksware:
Great to see this being taken care of after so long. I haven't had time to thoroughly test the merged PRs, but seems so far ok to me as one of them is so similar to the already provided solution in this PR. Thanks to all involved.

PR #1139, opened 2022-04-08
chore: Update stale issues GitHub action to use the latest version.

ashishdhingra:
*Issue #, if available:*

*Description of changes:*
chore: Update stale issues GitHub action to use the latest version.

*NOTE:* Pointing to explicit tag or commit enables to avoid any breaking changes, specially when the repository is maintained by 3rd party. Since `aws-actions` is maintained by us, to avoid changing the GitHub action reference every time new version is released, it was discussed to set it to `main`.

___

### DIFF
--- a/.github/workflows/stale_issues.yml
+++ b/.github/workflows/stale_issues.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Stale issue job
     steps:
-    - uses: aws-actions/stale-issue-cleanup@v3
+    - uses: aws-actions/stale-issue-cleanup@main
       with:
         # Setting messages to an empty string will cause the automation to skip
         # that category


.github/workflows/stale_issues.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Stale issue job
     steps:
-    - uses: aws-actions/stale-issue-cleanup@v3
+    - uses: aws-actions/stale-issue-cleanup@main
```
ashovlin, line 13: https://github.com/aws-actions/stale-issue-cleanup#readme recommends pinning an explicit tag or commit
> NOTE: For stability, you should use the action with either an explicit tag, or commit SHA

Should we follow our own advice here? If not, why diverge?

klementi (approved):

.github/workflows/stale_issues.yml
normj: If that is the case we should update the guidance in the README.

normj (approved):

PR #1165, opened 2022-04-30
Fixed issue with log multiline log message being broken for multiple CloudWatch Log records

normj:
*Description of changes:*
If log messages contain newlines each line gets broken up over separate CloudWatch Logs. This is particular bad for stack traces. This is due to the runtime using STDOUT and STDERR for logging and the Lambda service doesn't know the start and stop of messages other then using newlines.

To fix this problem the .NET runtime will write log messages to the Lambda telemetry file descriptor if it is available. Otherwise fallback to STDOUT and STDERR. This is similar to the [Java Lambda runtime](https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/logging/FramedTelemetryLogSink.java#L11).

Besides confirming all of the existing log tests a new stress test was added to confirm no threading issues and there was no affect to performance.

### DIFF
--- a/Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/ConsoleLoggerWriter.cs
+++ b/Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/ConsoleLoggerWriter.cs
@@ -38,19 +38,45 @@ namespace Amazon.Lambda.RuntimeSupport.Helpers
     /// </summary>
     public class SimpleLoggerWriter : IConsoleLoggerWriter
     {
+        TextWriter _writer;
+
+        public SimpleLoggerWriter()
+        {
+            // Look to see if Lambda's telemetry log file descriptor is available. If so use that for logging.
+            // This will make sure multiline log messages use a single CloudWatch Logs record.
+            var fileDescriptorLogId = Environment.GetEnvironmentVariable(Constants.ENVIRONMENT_VARIABLE_TELEMETRY_LOG_FD);
+            if (fileDescriptorLogId != null)
+            {
+                try
+                {
+                    _writer = FileDescriptorLogFactory.GetWriter(fileDescriptorLogId);
+                    InternalLogger.GetDefaultLogger().LogInformation("Using file descriptor stream writer for logging");
+                }
+                catch (Exception ex)
+                {
+                    _writer = Console.Out;
+                    InternalLogger.GetDefaultLogger().LogError(ex, "Error creating file descriptor log stream writer. Fallback to stdout.");
+                }
+            }
+            else
+            {
+                _writer = Console.Out;
+                InternalLogger.GetDefaultLogger().LogInformation("Using stdout for logging");
+            }
+        }
+
         public void SetCurrentAwsRequestId(string awsRequestId)
         {
-
         }
 
         public void FormattedWriteLine(string message)
         {
-            Console.WriteLine(message);
+            _writer.WriteLine(message);
         }
 
         public void FormattedWriteLine(string level, string message)
         {
-            Console.WriteLine(message);
+            _writer.WriteLine(message);
         }
     }
 
@@ -98,9 +124,6 @@ namespace Amazon.Lambda.RuntimeSupport.Helpers
             Critical = 5
         }
 
-        TextWriter _consoleStdOutWriter;
-        TextWriter _consoleErrorWriter;
-
         WrapperTextWriter _wrappedStdOutWriter;
         WrapperTextWriter _wrappedStdErrorWriter;
 
@@ -112,8 +135,31 @@ namespace Amazon.Lambda.RuntimeSupport.Helpers
         /// Stderror will default log messages to be Error
         /// </summary>
         public LogLevelLoggerWriter()
-            : this(Console.Out, Console.Error)
         {
+            // Look to see if Lambda's telemetry log file descriptor is available. If so use that for logging.
+            // This will make sure multiline log messages use a single CloudWatch Logs record.
+            var fileDescriptorLogId = Environment.GetEnvironmentVariable(Constants.ENVIRONMENT_VARIABLE_TELEMETRY_LOG_FD);
+            if (fileDescriptorLogId != null)
+            {
+                try
+                {
+                    var stdOutWriter = FileDescriptorLogFactory.GetWriter(fileDescriptorLogId);
+                    var stdErrorWriter = FileDescriptorLogFactory.GetWriter(fileDescriptorLogId);
+                    Initialize(stdOutWriter, stdErrorWriter);
+                    InternalLogger.GetDefaultLogger().LogInformation("Using file descriptor stream writer for logging.");
+                }
+                catch(Exception ex)
+                {
+                    InternalLogger.GetDefaultLogger().LogError(ex, "Error creating file descriptor log stream writer. Fallback to stdout and stderr.");
+                    Initialize(Console.Out, Console.Error);
+                }
+            }
+            else
+            {
+                Initialize(Console.Out, Console.Error);
+                InternalLogger.GetDefaultLogger().LogInformation("Using stdout and stderr for logging.");
+            }
+
             // SetOut will wrap our WrapperTextWriter with a synchronized TextWriter. Pass in the new synchronized
             // TextWriter into our writer to make sure we obtain a lock on that instance before writing to the stdout.
             Console.SetOut(_wrappedStdOutWriter);
@@ -127,11 +173,13 @@ namespace Amazon.Lambda.RuntimeSupport.Helpers
 
         public LogLevelLoggerWriter(TextWriter stdOutWriter, TextWriter stdErrorWriter)
         {
-            _consoleStdOutWriter = stdOutWriter;
-            _wrappedStdOutWriter = new WrapperTextWriter(_consoleStdOutWriter, LogLevel.Information.ToString());
+            Initialize(stdOutWriter, stdErrorWriter);
+        }
 
-            _consoleErrorWriter = stdErrorWriter;
-            _wrappedStdErrorWriter = new WrapperTextWriter(_consoleErrorWriter, LogLevel.Error.ToString());
+        private void Initialize(TextWriter stdOutWriter, TextWriter stdErrorWriter)
+        {
+            _wrappedStdOutWriter = new WrapperTextWriter(stdOutWriter, LogLevel.Information.ToString());
+            _wrappedStdErrorWriter = new WrapperTextWriter(stdErrorWriter, LogLevel.Error.ToString());
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


carlzogh (approved):

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/ConsoleLoggerWriter.cs
```diff
@@ -112,8 +135,31 @@ enum LogLevel
...
+            // This will make sure multiline log messages use a single CloudWatch Logs record.
+            var fileDescriptorLogId = Environment.GetEnvironmentVariable(Constants.ENVIRONMENT_VARIABLE_TELEMETRY_LOG_FD);
+            if (fileDescriptorLogId != null)
+            {
+                try
+                {
+                    var stdOutWriter = FileDescriptorLogFactory.GetWriter(fileDescriptorLogId);
+                    var stdErrorWriter = FileDescriptorLogFactory.GetWriter(fileDescriptorLogId);
+                    Initialize(stdOutWriter, stdErrorWriter);
+                    InternalLogger.GetDefaultLogger().LogInformation("Using file descriptor stream writer for logging.");
```
ganeshnj, line 149: Do this and ConsoleLoggerWriter.cs L53 log line appear same when logged? I believe there should be some prefix to differentiate the log lines emitted from different classes to make debugging easier.

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
```diff
@@ -0,0 +1,114 @@
...
+
+namespace Amazon.Lambda.RuntimeSupport.Helpers
+{
+    /// <summary>
+    /// This class wraps the utility of writing to the Lambda telemetry file descriptor for logging into a standard .NET Stream.
+    /// The message
+    /// </summary>
+    public static class FileDescriptorLogFactory
+    {
+        static ConcurrentDictionary<string, StreamWriter> _writers = new ConcurrentDictionary<string, StreamWriter>();
```
ganeshnj, line 15: can have restricted access
```suggestion
        private static readonly ConcurrentDictionary<string, StreamWriter> Writers = new();
```

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// +----------------------+------------------------+-----------------------+
+        /// The first 4 bytes are the frame type. For logs this is always 0xa55a0001.
+        /// The second 4 bytes are the length of the message.
+        /// The remaining bytes are the message itself. Byte order is big-endian.
+        /// </summary>
+        private class FileDescriptorLogStream : Stream
+        {
+            const uint FRAME_TYPE = 0xa55a0001;
+            Stream _fileDescriptorStream;
+            byte[] _frameTypeBytes;
```
ganeshnj, line 43:
```suggestion
            private const uint FRAME_TYPE = 0xa55a0001;
            private readonly Stream _fileDescriptorStream;
            private readonly byte[] _frameTypeBytes;
```

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            public FileDescriptorLogStream(string fileDescriptorId)
+            {
+                SafeFileHandle handle = new SafeFileHandle(new IntPtr(int.Parse(fileDescriptorId)), false);
+                _fileDescriptorStream = new FileStream(handle, FileAccess.Write);
```
ganeshnj, line 48: is it possible two lambda executions trying to write same file or `fileDescriptorId` is always unique per lambda execution?

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            #region Not implemented read and seek operations
+            public override bool CanRead => false;
+
+            public override bool CanSeek => false;
+
+            public override bool CanWrite => true;
+
+            public override long Length => throw new NotImplementedException();
```
ganeshnj, line 93:
```suggestion
            public override long Length => throw new NotSupportedException();
```

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            public override long Position { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
```
ganeshnj, line 95:
```suggestion
            public override long Position { get => throw new NotSupportedException(); set => throw new NotSupportedException(); }
```

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            public override int Read(byte[] buffer, int offset, int count)
+            {
+                throw new NotImplementedException();
```
ganeshnj, line 99:
```suggestion
                throw new NotSupportedException();
```

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            public override long Seek(long offset, SeekOrigin origin)
+            {
+                throw new NotImplementedException();
```
ganeshnj, line 104:
```suggestion
                throw new NotSupportedException();
```

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            public override void SetLength(long value)
+            {
+                throw new NotImplementedException();
```
ganeshnj, line 109:
```suggestion
                throw new NotSupportedException();
```

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            public override void Write(byte[] buffer, int offset, int count)
+            {
+                var messageLengthBytes = BitConverter.GetBytes(count - offset);
+                if (BitConverter.IsLittleEndian)
+                {
+                    Array.Reverse(messageLengthBytes);
+                }
+
+                var typeAndLength = ArrayPool<byte>.Shared.Rent(8);
```
ganeshnj, line 70: ❤️

ganeshnj (changes requested):

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/ConsoleLoggerWriter.cs
normj: There is no precedence for adding a prefix when using InternalLogger.GetDefaultLogger(). If we really wanted to do something like that it would be implemented inside `InternalLogger`.

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
normj: Added the `private` and `readonly`. I didn't use `new()` because it isn't supported in the version of C# this library targets which is 7.3.

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
normj: Done

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
normj: There is one `fileDescriptorId` per Lambda execution environment. The parent classes `GetWriter` makes sure we only use a single writer for a unique `fileDescriptorId`.

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
normj: Done

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
normj: Done

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
normj: Done

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
normj: Done

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
normj: Done

ganeshnj (approved):

normj:
It will take awhile for it to roll out when the next .NET patch is applied to the managed runtime. I don't know the exact Lambda rollout schedule but you can imagine code changes into Lambda are done very slowly and carefully.

If you need to this fix today you can follow the pattern of our top-level statement project templates where the version of Amazon.Lambda.RuntimeSupport is included in the deployment bundle.

svoychik:
Opened a bug to fix that https://github.com/aws/aws-lambda-dotnet/issues/1265

empty-space:
Issue still exists

PR #1171, opened 2022-05-06
New command line parameter "lambdas-file" to force lambdas assembly context

vancleimatheus:
*Issue #, if available:*
I didn't find any issue open for this problem. 

When the project that contains the lambdas references other projects in the same solution sometimes it fails to load the `FunctionHandler` because the first `deps.json` file returned is not always the one for the project itself, since there are other deps.json files for other projects in the same directory.

`Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LocalLambdaRuntime.cs, 56` 

So, if we run the test tool it throws a `FileNotFoundException` even though the assembly described in `serverless.template`  exists in the working directory.

*Description of changes:*

- Introduced a new `lambdas-file` parameter so we can specify exactly which DLL file we want the test tool to point at to extract the function handlers
- Included an additional logic to ignore `deps.json` files if a specific file was provided
- Two new tests to cover the new logic

### DIFF
--- a/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/CommandLineOptions.cs
+++ b/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/CommandLineOptions.cs
@@ -22,6 +22,8 @@ namespace Amazon.Lambda.TestTool
         public string AWSProfile { get; set; }
         
         public string AWSRegion { get; set; }
+
+        public string LambdasFile { get; set; }
         
         public bool ShowHelp { get; set; }
 
@@ -83,6 +85,10 @@ namespace Amazon.Lambda.TestTool
                         break;
                     case "--payload":
                         options.Payload = GetNextStringValue(i);
+                        i++;
+                        break;
+                    case "--lambdas-file":
+                        options.LambdasFile = GetNextStringValue(i);
                         i++;
                         break;
                     case "--pause-exit":
--- a/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LocalLambdaRuntime.cs
+++ b/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LocalLambdaRuntime.cs
@@ -41,25 +41,20 @@ namespace Amazon.Lambda.TestTool.Runtime
             this.AWSService = awsService;
         }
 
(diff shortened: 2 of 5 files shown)


normj:
I'm wondering if instead of a new parameter if we should first look for a `*.runtimeconfig.json` file which should only have in the directory. Once we find that use the prefix of the file as the assembly name and then get that's corresponding deps.json file. What do you think?

vancleimatheus:
> I'm wondering if instead of a new parameter if we should first look for a `*.runtimeconfig.json` file which should only have in the directory. Once we find that use the prefix of the file as the assembly name and then get that's corresponding deps.json file. What do you think?

What I'm seeing here is that I have more than one file both for `*.deps.json` and `*.runtimeconfig.json`, although these files are generated only for another api project referenced and the project itself. I will try to provide a minimum working example providing this multiple files error.

I agree 100% percent that it would be better to have only a fix instead of a whole new parameter

normj:
That is really surprising to have more then one `*.runtimeconfig.json` file. I would like to understand the circumstances that trigger that behavior.

ashishdhingra:
> > I'm wondering if instead of a new parameter if we should first look for a `*.runtimeconfig.json` file which should only have in the directory. Once we find that use the prefix of the file as the assembly name and then get that's corresponding deps.json file. What do you think?
> 
> What I'm seeing here is that I have more than one file both for `*.deps.json` and `*.runtimeconfig.json`, although these files are generated only for another api project referenced and the project itself. I will try to provide a minimum working example ilustrating this multiple json files generation.
> 
> I agree 100% percent that it would be better to have only a fix instead of a whole new parameter

@vancleimatheus Good afternoon. I was going through the list of open PRs and stumbled across this one. Please advise if you managed to have a minimum working example demonstrating multiple JSON file generation per your response above.

Thanks,
Ashish

ashishdhingra:
@vancleimatheus Could you please check if the issue addressed by this PR is fixed in latest version of Test tool? We had done some work in the commit https://github.com/aws/aws-lambda-dotnet/commit/8d43ff9b23af51624307cdb33f02384487e72468.

Thanks,
Ashish

normj:
Closing for lack of activity and I'm pretty sure PR https://github.com/aws/aws-lambda-dotnet/pull/1542 solve this issue.

Issue #1148, opened 2022-04-19
Enhance MarshalRequest/InvokeFeatures to setup the IHttpRequestIdentifierFeature


brendonparker:
### Describe the feature

Enhance the Lambda to AspNetCore handoff to include either (or both) the `AwsRequestId` from the `ILambdaContext` and the `TraceId` from the `LambdaContext`.

### Use Case

This will help log correlation with things like AWS X-Ray, or other AWS Logs that are capturing the RequestId.

### Proposed Solution

1. Enhance `Amazon.Lambda.AspNetCoreServer.Internal.InvokeFeatures` to implement `Microsoft.AspNetCore.Http.Features.IHttpRequestIdentifierFeature`.
2. Enhance the various implementations of `MarshallRequest` to setup the `IHttpRequestIdentifierFeature`, specifying using the `ILambdaContext.AwsRequestId`.

### Other Information

See linked discussion for the path I took to get here. I'm not certain which makes more sense to use, the AwsRequestId or the TraceId. May depend on if you are using AWS X-Ray. Perhaps configurable?

This _may_ be considered a breaking change, as it will no longer use the dotnet-generated RequestId. However, I feel using the AwsRequestId is more correct.

### Acknowledgements

- [X] I may be able to implement this feature request
- [X] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

Amazon.Lambda.AspNetCoreServer.Hosting 1.0.0

### Targeted .NET Platform

.NET Core 3.1, .NET 6

### Operating System and version

Windows 10, OSX Mojave, Ubuntu, AmazonLinux

brendonparker:
Here is the discussion: https://github.com/aws/aws-lambda-dotnet/discussions/1147

normj:
Thanks for the great feature request. I like the idea of having the `HttpContext.TraceIdentifier` match the X-Ray trace ID. I think we should aim to do that.

As far as customization goes in the older style of writing ASP.NET Core Lambda functions you used to be able to override methods from a base class to customize the marshalling between API Gateway and ASP.NET Core. That ability got lost in the new approach using the `AddAWSLambdaHosting` extension method to initialize Lambda. My plan is to allow `Action` callbacks to be added during the `AddAWSLambdaHosting` method to provide back the old customization features.

brendonparker:
Yes, some config or extensibility hooks may solve parts of this. 

Let me know if I can be of more help, create a PR, etc.

[...]

PR #1172, opened 2022-05-07
Set the HttpContext.TraceIdentifier to the trace id for the Lambda invocation

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/1148

*Description of changes:*
For ASP.NET Core Lambda function use Lambda's trace id that can be found in the `_X_AMZN_TRACE_ID` environment variable and set it as the `HttpContext.TraceIdentifier`.

### DIFF
--- a/Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/InvokeFeatures.cs
+++ b/Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/InvokeFeatures.cs
@@ -25,9 +25,10 @@ namespace Amazon.Lambda.AspNetCoreServer.Internal
                              IHttpResponseFeature,
                              IHttpConnectionFeature,
                              IServiceProvidersFeature,
-                             ITlsConnectionFeature
+                             ITlsConnectionFeature,
+                             IHttpRequestIdentifierFeature,
 
-                             ,IHttpResponseBodyFeature
+                             IHttpResponseBodyFeature
 
 #if NET6_0_OR_GREATER
                             ,IHttpRequestBodyDetectionFeature
@@ -50,6 +51,7 @@ namespace Amazon.Lambda.AspNetCoreServer.Internal
             this[typeof(IServiceProvidersFeature)] = this;
             this[typeof(ITlsConnectionFeature)] = this;
             this[typeof(IHttpResponseBodyFeature)] = this;
+            this[typeof(IHttpRequestIdentifierFeature)] = this;
 
 #if NET6_0_OR_GREATER
             this[typeof(IHttpRequestBodyDetectionFeature)] = this;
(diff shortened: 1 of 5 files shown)


boblodgett (approved):

Libraries/test/Amazon.Lambda.AspNetCoreServer.Test/TestApiGatewayHttpApiV2Calls.cs
```diff
@@ -235,6 +235,22 @@ public async Task TestMultipleCookie()
             Assert.Equal("TestValue3", response.Body);
         }
 
+        [Fact]
+        public async Task TestTraceIdSetFromLambdaContext()
+        {
+            Environment.SetEnvironmentVariable("_X_AMZN_TRACE_ID", "MyTraceId-1");
```
ganeshnj, line 241: 1. shouldn't manipulate the real environment variable ideally, creating a wrapper around will solve this
2. If we do manipulate real env var, it makes sense to clear them out to be on safe side

normj: I agree it would be good to have environment variables be mockable in this library but it just isn't setup that way right now to pass around an environment variable abstractions through out the code. I'm not going to do a major refactor like that right now.

I did add an explicit clear of the environment variable in the test.

ganeshnj (approved):

Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/InvokeFeatures.cs
```diff
@@ -354,6 +356,33 @@ public Task<X509Certificate2> GetClientCertificateAsync(CancellationToken cancel
...
+
+                var lambdaTraceId = Environment.GetEnvironmentVariable("_X_AMZN_TRACE_ID");
+                if (!string.IsNullOrEmpty(lambdaTraceId))
+                {
+                    return lambdaTraceId;
+                }
+
+                // If there is no Lambda trace id then fallback to the trace id that ASP.NET Core would have generated.
+                var baseFeature = new Microsoft.AspNetCore.Http.Features.HttpRequestIdentifierFeature();
+                return baseFeature.TraceIdentifier;
```
martincostello, line 379: Would new-ing up a new instance every time cause the TraceIdentifier value to change on each invocation if the environment variable isn't set?

brendonparker: Oh, yes. Need to capture that in `_traceIdentifier`, so:
```
_traceIdentifier = baseFeature.TraceIdentifier;
return _traceIdentifier;
```

normj: It would change for every invocation which is the desired behavior. In the Lambda case the Lambda environment is changing the `_X_AMZN_TRACE_ID` environment variable for every invocation.

martincostello: What I mean is would it change between different accesses within the lifetime of a single "HTTP request"? I would expect it to be stable while dealing with a single invocation.

brendonparker: @normj Yes, if the environment variable is set. If it is not, it will get a new string on each "get" of the `TraceIdentifier` property. I imagine that should be constant for the lifecycle of the request?

normj: I understand now. I was being naive thinking the getter was only called once to construct the `HttpContext` but that is probably not true and `HttpContext` forwards down into the feature set. I update the code to hold on to the identifier.

ganeshnj (approved):

PR #1201, opened 2022-06-12
Do not log preamble Unicode byte order mark

carlzogh:
The log entry will be ignored by the Lambda Platform as UTF-8 is assumed for the log entry content and decoding this entry results in an empty log entry being written.

Added unit tests to validate logger behavior.

CloudWatch limit references:
- Event size: 256kb (ref. https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/cloudwatch_limits_cwl.html)
- Overhead per log event: 26 bytes (ref. BatchSize in https://docs.aws.amazon.com/opsworks/latest/APIReference/API_CloudWatchLogsLogStream.html)

### DIFF
--- a/Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
+++ b/Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
@@ -15,6 +15,11 @@ namespace Amazon.Lambda.RuntimeSupport.Helpers
     {
         private readonly static ConcurrentDictionary<string, StreamWriter> _writers = new ConcurrentDictionary<string, StreamWriter>();
 
+        // max cloudwatch log event size, 256k - 26 bytes of overhead.
+        internal const int MaxCloudWatchLogEventSize = 256 * 1024 - 26;
+        internal const int LambdaTelemetryLogHeaderLength = 8;
+        internal const uint LambdaTelemetryLogHeaderFrameType = 0xa55a0001;
+
         /// <summary>
         /// Get the StreamWriter for the particular file descriptor ID. If the same ID is passed the same StreamWriter instance is returned.
         /// </summary>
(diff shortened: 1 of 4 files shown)


luben:
LGTM.

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
```diff
@@ -15,17 +15,39 @@ public static class FileDescriptorLogFactory
...
+        internal const uint LambdaTelemetryLogHeaderFrameType = 0xa55a0001;
+
         /// <summary>
         /// Get the StreamWriter for the particular file descriptor ID. If the same ID is passed the same StreamWriter instance is returned.
         /// </summary>
         /// <param name="fileDescriptorId"></param>
         /// <returns></returns>
         public static StreamWriter GetWriter(string fileDescriptorId)
+        {
+            SafeFileHandle handle = new SafeFileHandle(new IntPtr(int.Parse(fileDescriptorId)), false);
```
normj, line 30: Add the creation of the `SafeFileHandle` into Lambda expression for calling `InitializeWriter`. Otherwise we are creating an instance of `SafeFileHandle` every time `GetWriter` is called even if we are returning back a cached `StreamWriter`. This could become problematic given `SafeFileHandle` implements `IDisposable`. 

Something like 

```
            var writer = _writers.GetOrAdd(fileDescriptorId, 
                (x) => {
                    SafeFileHandle handle = new SafeFileHandle(new IntPtr(int.Parse(fileDescriptorId)), false);
                    return InitializeWriter(new FileStream(handle, FileAccess.Write));
                });
            return writer;
```

normj (changes requested):

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
carlzogh: Thanks Norm, fixed!

normj (approved):

ganeshnj (approved):

PR #1202, opened 2022-06-14
readme.md: Fix example and document highly useful, undocumented feature

hounddog22030:
*Description of changes:*
The example HttpApi attribute decoration in the readme.md is outdated and causes a compile error.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1207, opened 2022-06-17
Fix/error message type name

hounddog22030:
*Description of changes:*

If the LambdaSerializerAttribute is missing, during invocation of function, the error message of "Could not find the LambdaSerializerAttribute on the assembly 'XXX' or method 'XXX' while attempting to deserialize input data of type 'XXX'. To use types other than System.IO.Stream as input/output parameters, the assembly or Lambda function should be annotated with Amazon.Lambda.LambdaSerializerAttribute." incorrectly names "Amazon.Lambda.LambdaSerializerAttribute".  Should be "Amazon.Lambda..Core.LambdaSerializerAttribute"

Fixed by using `typeof(LambdaSerializerAttribute).FullName`.

I see below that my fork master also has the other PR (https://github.com/aws/aws-lambda-dotnet/pull/1202) in it, so maybe review it first.  I guess I need to not merge my changes into my fork/master anymore.  Sorry.


normj (approved):

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/InvokeDelegateBuilder.cs
```diff
@@ -320,7 +320,7 @@ private Expression CreateSerializeExpression(object customerSerializerInstance,
                 if (customerSerializerInstance == null)
                 {
                     throw LambdaExceptions.ValidationException(Errors.UserCodeLoader.SerializeMissingAttribute,
-                        _handler.AssemblyName, _handler.MethodName, dataType.FullName);
```
ganeshnj, line 323: slight perf, we should cache the `typeof(LambdaSerializerAttribute)` in a member to avoid performing same operation. Not ship blocking.

ganeshnj (approved):

normj:
Thanks for the PR

PR #1208, opened 2022-06-17
fix: source for annotations assembly in nuspec

attilah:
*Issue #, if available:*

N/A

*Description of changes:*

Change the source path for `Amazon.Lambda.Annotations.dll` to make sure that the code signed version will be added to the package similar as it is in line 23.


ganeshnj:
Could you compare the published package and new package that we have same artifacts?

normj (approved):

attilah:
@ganeshnj I've verified it, is the same.

ganeshnj (approved):

Issue #1232, opened 2022-06-30
Microsoft.Extensions.Logging.ILogger interface support for Lambda


dhhoang:
### Describe the feature

The `dotnet6` Lambda runtime should offer a [log provider](https://docs.microsoft.com/en-us/dotnet/core/extensions/custom-logging-provider) that supports the [`Microsoft.Extensions.Logging.ILogger`](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/) logging interface.

### Use Case

We have lots of Lambdas written in C# that makes heavy uses of .NET built-in Dependency injection (aka `IServiceProvider`). The typical usage is to set up the DI container in the function's constructor, and uses it to resolve the classes in the function handler method.  We believe this is the "right" way, as creating a new DI container per invocation is too expensive.  
The provided `ILambdaLogger` interface used for logging however is wrapped inside `ILambdaContext` which is provided per-invocation, making it impossible to set up in the with the DI pipeline (without some ugly work-around). We believe using `IServiceProvider` is very common among .NET customers.

On top of that `ILamdaLogger` misses some features compared to `ILogger`:
* No parameterized logging based on log level. This means that the log message is formed **regardless** of the configured log level. For example, even when we configure `AWS_LAMBDA_HANDLER_LOG_LEVEL` to `Warning`, our `Debug` and `Info` messages are stilled formed. This is not very efficient.
* No support for templatized/structured logging or custom log formatting (e.g. JSON). Being able to output JSON logs will allow us to have flexible queries with CloudWatch Logs insight. This has also been raised in https://github.com/aws/aws-lambda-dotnet/issues/1000 and https://github.com/aws/aws-lambda-dotnet/issues/1144 .

### Proposed Solution

We propose that the runtime offers an implementation of `ILoggerProvider` (aka `LambdaLogProvider`, see [Microsoft docs](https://docs.microsoft.com/en-us/dotnet/core/extensions/custom-logging-provider)).    
There is currently an existing implementation in `Amazon.Lambda.Logging.AspNetCore` which delegates back to `LambdaLogger` but we would like to see it implemented in the core runtime so that other app types can benefit from it.     
We have a PoC implementation here https://github.com/dhhoang/LambdaCustomLogger. Compared to the implementation in `Amazon.Lambda.Logging.AspNetCore`, it has some added features:
* Allows users to extend and support custom formats by registering `ILogHandler`.   
[...]

PR #1238, opened 2022-07-11
Add raw data logging support for LambdaLogger

dhhoang:
*Issue #, if available:* : https://github.com/aws/aws-lambda-dotnet/issues/1232

*Description of changes:* Add an overload to `LambdaLogger.Log()` to allow logging using raw data.

## Overview 
In https://github.com/aws/aws-lambda-dotnet/issues/1232 we brought up our use case of `ILogger` interface. Before we can implement such a logger we need to have a way to forward raw UTF8 data to the output. In this PR we propose that an overload is  added to `LambdaLogger`  (for .NET 6 runtime only)
```csharp
#if NET6_0_OR_GREATER
/// <summary>
/// Logs a message to AWS CloudWatch Logs. <br/>
/// Logging will not be done:
/// If the role provided to the function does not have sufficient permissions.
/// </summary>
/// <param name="utf8Message">The message as UTF-8 encoded data.</param>
public static void Log(ReadOnlySpan<byte> utf8Message)
#endif
```
This will be implemented in a similar fashion to `Log(string)`: it's delegated to a default method but `LogLevelLoggerWriter` will replace that method with the formatted version using reflection.  
To do this the output `Stream` is passed to `WrapperTextWriter`. If the Telemetry logging file descriptor is available this will be the `FileStream` wrapper for that FD, otherwise the stream is set to `NULL` and the logging action is delegated back to `FormattedWriteLine()`

## Testing
Because this new function is only supported on .NET 6 and higher we added `net6.0` target to `Amazon.Lambda.RuntimeSupport.UnitTests.csproj`.   
We added equivalent tests to existing tests for `FileDescriptorLogFactory.FileDescriptorLogStream` (raw data writing variant). We also added several unit tests to `LogLevelLoggerWriter`.  
We also **fixed a failing test** : `HandlerTests.PositiveHandlerTestsAsync`. This test is failing for .NET 6 because by default the `LambdaBootstrap` always initializes the `LogLevelLoggerWriter` instance, which in turn redirects the `LambdaLogger` output, causing the logs to not be redirected to the test. This is fixed by creating another constructor for `LambdaBootstrap` to accept a `IConsoleLoggerWriter` instance which the test itself provides.

## Remarks
There are two places whether the logging action of `LambdaLogger` is set. The first place is `LogLevelLoggerWriter` which we include in this PR. The second place is in `UserCodeLoader.SetCustomerLoggerLogAction()`. We are still not sure what the usage of this method. I'm happy to hear about this usage and amend the PR.  

### DIFF
--- a/Libraries/src/Amazon.Lambda.Core/LambdaLogger.cs
+++ b/Libraries/src/Amazon.Lambda.Core/LambdaLogger.cs
@@ -12,11 +12,40 @@ namespace Amazon.Lambda.Core
         // Logging action, logs to Console by default
         private static Action<string> _loggingAction = LogToConsole;
 
+#if NET6_0_OR_GREATER
+        private static System.Buffers.ReadOnlySpanAction<byte, object> _dataLoggingAction = LogUtf8BytesToConsole;
+#endif
+
         // Logs message to console
         private static void LogToConsole(string message)
         {
             Console.WriteLine(message);
         }
+
+#if NET6_0_OR_GREATER
+        private static void LogUtf8BytesToConsole(ReadOnlySpan<byte> utf8Message, object state)
+        {
+            try
+            {
+                Console.WriteLine(Console.OutputEncoding.GetString(utf8Message));
+            }
+            catch
+            {
+                // ignore any encoding error
+            }
+        }
+
+        /// <summary>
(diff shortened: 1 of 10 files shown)


dhhoang:
Closing in favor or https://github.com/aws/aws-lambda-dotnet/issues/1232

Issue #1200, opened 2022-06-11
Decorating async Task function with [LambdaFunction] generates compile error


hounddog22030:
### Describe the bug

The generated code is awaiting the Task and attempting to return the Task, but because the original method is `async`, the original method is really a `void`, so it cannot be returned.

### Expected Behavior

No compile errors.

### Current Behavior

Receive Compile Error:

```
Severity	Code	Description	Project	File	Line
Error	CS1997	Since 'Functions_SqsMessageHandler_Generated.SqsMessageHandler(SQSEvent.SQSMessage, ILambdaContext)' is an async method that returns 'Task', a return keyword must not be followed by an object expression. Did you intend to return 'Task<T>'?	ServerlessWithSqs	C:\Users\17034\source\repos\ServerlessWithSqs\ServerlessWithSqs\Amazon.Lambda.Annotations.SourceGenerator\Amazon.Lambda.Annotations.SourceGenerator.Generator\Functions_SqsMessageHandler_Generated.g.cs	35

```

### Reproduction Steps

Steps to reproduce:
1. Create a new serverless project
2. Choose the Annotations blueprint
3. In Function.cs add the below code:

```
    [LambdaFunction]
    public async Task SqsMessageHandler(SQSEvent.SQSMessage sqsMessage, ILambdaContext context)
    {
        await Task.Delay(1);
    }

```

4. Build the project

### Possible Solution

remove the return?
return without the await?

### AWS .NET SDK and/or Package version used

    <PackageReference Include="Amazon.Lambda.Annotations" Version="0.5.0-preview" />

### Targeted .NET Platform

.NET 6

### Operating System and version

Windows 11, AmazonLinux

ashishdhingra:
Reproducible. Needs review with the team.

ashishdhingra:
@hounddog22030 Looks like the linked PR https://github.com/aws/aws-lambda-dotnet/pull/1257 is merged. Could you please test your scenario and confirm if it works now?

Thanks,
Ashish

PR #1257, opened 2022-07-31
Fix issue with Source generator for Lambda functions that return void

normj:
*Description of changes:*
For Amazon.Lambda.Annotations if a method with the `LambdaMethod` attribute has its return type set to `void` the source generator would generate invalid code.

For example if you had a Lambda function like this
```
public class VoidExample
{
    [LambdaFunction()]
    public void VoidReturn(string text, ILambdaContext context)
    {
        context.Logger.LogLine(text);
    }
}
```

The generated code that wraps this call would previously look like this with an invalid `return` statement when calling this void method.
```
public class VoidExample_VoidReturn_Generated
{
    private readonly VoidExample voidExample;

...

    public void VoidReturn(string text, Amazon.Lambda.Core.ILambdaContext __context__)
    {
        return voidExample.VoidReturn(text, __context__);
    }

...
}
```

This fix removes the invalid `return` statement. The reason this code was happening is the code was treating methods that return `void` and `Task` as the same via the `ReturnVoidOrTask` property. Since it is not safe to treat void and task the same I replaced this property with `ReturnVoid` and `ReturnTask` forcing users to think what is the right thing to do in the context.

Ignore the `LambdaFunctionTemplate.cs` file because that is generated from the `LambdaFunctionTemplate.tt` file.

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/LambdaMethodModel.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/LambdaMethodModel.cs
@@ -18,9 +18,14 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Models
         public bool IsAsync { get; set; }
 
         /// <summary>
-        /// Returns true if original method returns void or <see cref="System.Threading.Tasks.Task"/>
+        /// Returns true if original method returns void
         /// </summary>
-        public bool ReturnsVoidOrTask { get; set; }
+        public bool ReturnsVoid { get; set; }
+
+        /// <summary>
+        /// Returns true if original method returns <see cref="System.Threading.Tasks.Task"/>
+        /// </summary>
+        public bool ReturnsTask { get; set; }
 
         /// <summary>
         /// Gets or sets the return type of the method.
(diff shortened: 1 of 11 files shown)


philasmar (approved):

Libraries/test/TestServerlessApp/output.template
```diff
@@ -0,0 +1 @@
...otations","SyncedEvents":["RootGet"]},"Properties":{"MemorySize":256,"Timeout":50,"Policies":["AWSLambdaBasicExecutionRole"],"PackageType":"Image","Events":{"RootGet":{"Type":"HttpApi","Properties":{"Path":"/Greeter/SayHelloAsync","Method":"GET","PayloadFormatVersion":"1.0"}}},"ImageUri":"626492997873.dkr.ecr.us-west-2.amazonaws.com/testserverlessapp:greetersayhello-c013811cde6a-latest","ImageConfig":{"Command":["TestServerlessApp::TestServerlessApp.Greeter_SayHelloAsync_Generated::SayHelloAsync"]}}},"ToUpper":{"Type":"AWS::Serverless::Function","Metadata":{"Tool":"Amazon.Lambda.Annotations"},"Properties":{"Runtime":"dotnet6","CodeUri":"s3://normj-west2/","MemorySize":256,"Timeout":30,"Policies":["AWSLambdaBasicExecutionRole"],"PackageType":"Zip","Handler":"TestServerlessApp::TestServerlessApp.Sub1.Functions_ToUpper_Generated::ToUpper"}}},"Outputs":{"RestApiURL":{"Description":"Rest API endpoint URL for Prod environment","Value":{"Fn::Sub":"https://${ServerlessRestApi}.execute-api.${AWS::Region}.amazonaws.com/Prod"}},"HttpApiURL":{"Description":"HTTP API endpoint URL for Prod environment","Value":{"Fn::Sub":"https://${ServerlessHttpApi}.execute-api.${AWS::Region}.amazonaws.com"}}}}
```
ganeshnj, line 1: Should this exist?

serverless.template must have the new function.

normj: Removed

ganeshnj (approved):

hounddog22030:
@normj - is this related:  https://github.com/aws/aws-lambda-dotnet/issues/1200

thanks

PR #1258, opened 2022-07-31
Fix issue of syncing CF template when all LambdaFunctions are removed.

normj:
*Description of changes:*
Fix an issue when all of the remaining `LambdaFunction` are removed from the project the CF template is not synced leaving the Lambda functions defined in the template.

The fix has the following changes:
* Move the computation of the project directory. It used to be computed from the first `LambdaFunction` but since there might not be any `LambdaFunction` it was moved to the syntax visitor and where it can be computed on the first class that was found.
* Change the Generator's short circuit from there being no `LambdaFunction` which caused the bug to only short circuit if the project directory was not calculated. That would happen if there are no classes in the project.

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/CloudFormationTemplateFinder.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/CloudFormationTemplateFinder.cs
@@ -35,11 +35,27 @@ namespace Amazon.Lambda.Annotations.SourceGenerator
 
         public string FindCloudFormationTemplate(string projectRootDirectory)
         {
+            var templateAbsolutePath = DetermineCloudFormationTemplatePath(projectRootDirectory);
+
+            if (!_fileManager.Exists(templateAbsolutePath))
+                _fileManager.Create(templateAbsolutePath).Close();
+            
+            return templateAbsolutePath;
+        }
+
+        public bool DoesCloudFormationTemplateExist(string projectRootDirectory)
+        {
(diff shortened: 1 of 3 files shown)


Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/CloudFormationTemplateFinder.cs
```diff
@@ -34,12 +34,28 @@ public string DetermineProjectRootDirectory(string sourceFilePath)
...
+            if (!_fileManager.Exists(templateAbsolutePath))
+                _fileManager.Create(templateAbsolutePath).Close();
+            
+            return templateAbsolutePath;
+        }
+
+        public bool DoesCloudFormationTemplateExist(string projectRootDirectory)
+        {
+            var templateAbsolutePath = DetermineCloudFormationTemplatePath(projectRootDirectory);
+            return File.Exists(templateAbsolutePath);
```
ganeshnj, line 49: should use IFileManager wrapper here.
```suggestion
            return _fileManager.Exists(templateAbsolutePath);
```

ganeshnj (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/CloudFormationTemplateFinder.cs
normj: Good catch

ganeshnj (approved):

philasmar (approved):

PR #1274, opened 2022-08-16
Fix spelling from substract to subtract

isonaj:
*Description of changes:*
While using the Annotations template, I noticed a spelling mistake in the work subtract. I've fixed it.


ashovlin (approved):
Oops, thanks for the fix! I'm asking around internally why the check is failing, but don't believe it's related to this.

PR #1283, opened 2022-08-25
Fix code generation in Lambda Annotations when return a non generic Task

normj:
*Description of changes:*
If a function was returning a non generic Task for functions that have no event attributes the generated code was incorrectly a return statement after the `await`.

This PR removes the return and adds the missing test case for this scenario.


96malhar (commented):
You need to bump up the count to 14 for the integ tests to pass since a new function has been added - https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/test/TestServerlessApp.IntegrationTests/IntegrationTestContextFixture.cs#L58

normj:
> You need to bump up the count to 14 for the integ tests to pass since a new function has been added - https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/test/TestServerlessApp.IntegrationTests/IntegrationTestContextFixture.cs#L58

Updated the test.

philasmar (approved):

ashovlin (approved):

Issue #1300, opened 2022-09-03
Allow AspNetCoreServer.Hosting to use a custom serializer


jeastham1993:
### Describe the feature

Currently the AspNetCoreServer.Hosting package initializes a DefaultLambdaJsonSerializer when creating the HandlerWrapper for the specific RuntimeSupportServer. Update this to use DefaultLamddaJsonSerializer as a default but allow a custom JsonSerializer to be used if required.

A link to the line in question can be seen here:
https://github.com/aws/aws-lambda-dotnet/blob/0e8e42a0fe818f1adc1f97bb035f2535934c16b4/Libraries/src/Amazon.Lambda.AspNetCoreServer.Hosting/Internal/LambdaRuntimeSupportServer.cs#L73

### Use Case

Allow source generated serialization to be used when running an API on Lambda. If compiling an API with NativeAOT enabled to run on Lambda the DefaultLambdaJsonSerializer will not work and a custom serialization context is required.

### Proposed Solution

Add an optional parameter to the AddAWSLambdaHosting method to allow a custom serializer to be passed in. Store the option in a static variable.

Update the CreateHandlerWrapper method to use this static variable instead of being hardcoded to initialize a new DefaultLambdaJsonSerializer.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

Amazon.Lambda.AspNetCoreServer.Hosting 1.0.0

### Targeted .NET Platform

.NET6 / .NET 7

### Operating System and version

ASL2

ashishdhingra:
This needs review with the team. @jeastham1993 Feel free to contribute PR so that it could be reviewed by the team.

jeastham1993:
@ashishdhingra happy to. I've got a version I've been using for testing locally. I'll create a PR today or tomorrow for review.

jeastham1993:
@ashishdhingra PR raised.

normj:
PR has been released as part of version **1.5.0** of Amazon.Lambda.AspNetCoreServer.Hosting

PR #1311, opened 2022-09-15
#1300 ASP.NET Core Server Custom Serializer

jeastham1993:
#1300 

Allow the AspNetCoreServer.Hosting package to use a user provided serializer.


Libraries/src/Amazon.Lambda.AspNetCoreServer.Hosting/ServiceCollectionExtensions.cs
```diff
@@ -35,8 +37,23 @@ public static class ServiceCollectionExtensions
...
+
+        /// <summary>
+        /// Add the ability to run the ASP.NET Core Lambda function in AWS Lambda. If the project is not running in Lambda 
+        /// this method will do nothing allowing the normal Kestrel webserver to host the application.
+        /// </summary>
+        /// <param name="services"></param>
+        /// <param name="eventSource"></param>
+        /// <param name="serializer"></param>
+        /// <returns></returns>
+        public static IServiceCollection AddAWSLambdaHosting(this IServiceCollection services, LambdaEventSource eventSource, ILambdaSerializer serializer)
```
normj, line 56: Instead of using a static place holder I think it would be more idiomatic to have this method that takes in an action callback that can be used for configuing.

Something like 
```
public static IServiceCollection AddAWSLambdaHosting(this IServiceCollection services, LambdaEventSource eventSource, ILambdaSerializer serializer, Action<HostingOptions> configure = null)
{
    var options = new HostingOptions();
     if(configure != null) configure.Invoke(options);

    ...

    var serializer = options.Serializer ?? new DefaultLambdaJsonSerializer();

   ...
}
```

normj (changes requested):

Libraries/src/Amazon.Lambda.AspNetCoreServer.Hosting/ServiceCollectionExtensions.cs
jeastham1993: @normj code updated, I've added an integration test as well. I think I understood the integration test framework correctly and they pass on my machine. Let me know if you want any more changes.

Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/Utilities.cs
```diff
@@ -22,6 +24,14 @@ namespace Amazon.Lambda.AspNetCoreServer.Internal
     /// </summary>
     public static class Utilities
     {
+        // private static ILambdaSerializer _serializer;
```
normj, line 27: Please remove the commented out code.

Libraries/src/Amazon.Lambda.AspNetCoreServer.Hosting/ServiceCollectionExtensions.cs
```diff
@@ -37,20 +39,42 @@ public static class ServiceCollectionExtensions
...
             // Not running in Lambda so exit and let Kestrel be the web server
             if (string.IsNullOrEmpty(Environment.GetEnvironmentVariable("AWS_LAMBDA_FUNCTION_NAME")))
                 return services;
 
+            var hostingOptions = new HostingOptions();
+            
+            if (configure != null)
+                configure.Invoke(hostingOptions);
+
+            services.AddSingleton<ILambdaSerializer>(hostingOptions.Serializer ?? new DefaultLambdaJsonSerializer());
```
normj, line 66: Use `TryAddSingleton` just in case somebody gets the idea to add the `ILambdaSerializer` directly to the service collection.

Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/HostingOptions.cs
```diff
@@ -0,0 +1,8 @@
+﻿using Amazon.Lambda.Core;
+
+namespace Amazon.Lambda.AspNetCoreServer.Internal;
```
normj, line 3: Since `HostingOptions` is used in the public method `AddAWSLambdaHosting` the type should not be in the `Internal` namespace and needs node.

normj (changes requested):

Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/Utilities.cs
jeastham1993: Done

Libraries/src/Amazon.Lambda.AspNetCoreServer.Hosting/ServiceCollectionExtensions.cs
jeastham1993: Done.

Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/HostingOptions.cs
jeastham1993: Done.

normj (approved):

normj:
I ended up doing some slight tweaks to the PR before finishing the merge. https://github.com/aws/aws-lambda-dotnet/commit/872b5109fe1e266fd86a81cbf6f2f6490cc6a33e

normj:
PR has been released as part of version **1.5.0** of Amazon.Lambda.AspNetCoreServer.Hosting

PR #1320, opened 2022-09-23
Remove useless null checks

bjorg:
*Description of changes:*

Remove `null` checks since they are not useful. The `Content` and `Headers` properties can never be null.

See:
* https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpResponseMessage.cs#L42
* https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpContent.cs#L112

### DIFF
--- a/Libraries/src/Amazon.Lambda.RuntimeSupport/Client/InternalClientAdapted.cs
+++ b/Libraries/src/Amazon.Lambda.RuntimeSupport/Client/InternalClientAdapted.cs
@@ -243,9 +243,6 @@ namespace Amazon.Lambda.RuntimeSupport
                     try
                     {
                         var headers_ = System.Linq.Enumerable.ToDictionary(response_.Headers, h_ => h_.Key, h_ => h_.Value);
-
-                        // NOTE: response_.Content can never be null (See: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpResponseMessage.cs#L42)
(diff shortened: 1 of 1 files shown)


Libraries/src/Amazon.Lambda.RuntimeSupport/Client/InternalClientAdapted.cs
```diff
@@ -243,11 +243,11 @@ public async System.Threading.Tasks.Task<SwaggerResponse<System.IO.Stream>> Next
                     try
                     {
                         var headers_ = System.Linq.Enumerable.ToDictionary(response_.Headers, h_ => h_.Key, h_ => h_.Value);
-                        if (response_.Content != null && response_.Content.Headers != null)
-                        {
-                            foreach (var item_ in response_.Content.Headers)
+
+                        // NOTE: response_.Content can never be null (See: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpResponseMessage.cs#L42)
```
normj, line 247: I'm good with the change but we don't need the comments. Those links will get out of date overtime so I would rather not have them.

normj (changes requested):

Libraries/src/Amazon.Lambda.RuntimeSupport/Client/InternalClientAdapted.cs
bjorg: Done

normj (approved):

normj:
Thanks for the PR!

Issue #1317, opened 2022-09-22
`dotnet new lambda/serverless.* ...` fails with .NET 7.0.100-rc.1.22431.12


bjhogan:
### Describe the bug

With the .NET SDK 7.0.100-rc.1.22431.12, `dotnet new lambda.EmptyFunction ...` and other templates fail. 

```
> dotnet new lambda.EmptyFunction -n MyEmptyFunction
Template "Lambda Empty Function" could not be created.
Failed to create template.
Details: Object reference not set to an instance of an object.

For details on the exit code, refer to https://aka.ms/templating-exit-codes#100
```

```
> dotnet new serverless.AspNetCoreWebApp -n AspNetCoreWebApp
Template "Lambda ASP.NET Core Web Application with Razor Pages" could not be created.
Failed to create template.
Details: Object reference not set to an instance of an object.
```

### Expected Behavior

`dotnet new lambda.EmptyFunction ...` creates the appropriate projects.

### Current Behavior

Command fails with - 

```
Template "Lambda Empty Function" could not be created.
Failed to create template.
Details: Object reference not set to an instance of an object.

For details on the exit code, refer to https://aka.ms/templating-exit-codes#100
```

### Reproduction Steps

Download .NET SDK 7.0.100-rc.1.22431.12 

Run `dotnet new lambda.EmptyFunction -n MyEmptyFunction`

### Possible Solution

Can workaround this problem by creating a `global.json` file for an earlier version of .NET.
 
`dotnet new globaljson --sdk-version 6.0.401`

### AWS .NET SDK and/or Package version used

.NET SDK 7.0.100-rc.1.22431.12

Amazon.Lambda.Templates (version 6.3.0)
amazon.lambda.tools version 5.4.5

### Targeted .NET Platform

.NET SDK 7.0.100-rc.1.22431.12

### Operating System and version

Windows 10

normj:
I opened up an issue against on the Microsoft templating repository to get some help debugging this issue. https://github.com/dotnet/templating/issues/5300

normj:
Version 6.4.0 of the template has been released with the fix. Thanks for reporting the issue Bryan!

PR #1321, opened 2022-09-23
Fixed issue with creating projects from Lambda templates with .NET 7 installed.

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/1317

*Description of changes:*
The Lambda templates were using a "deprecated" feature to handle the things like namespaces and filenames allowed different characters. For example if you create a project like `Test-Test` that is a valid name for the files and directories but due to the `-` it is an invalid namespace name. The original work around was to use a `safe-sourcename` replacement parameter using `safe_namespace` and `safe_name` macros. 

For quite some time the .NET templating engine now handles this safe name logic automatically and the `safe_namespace` and `safe_name` macros no longer due anything. There was an intended change in the .NET 7 RC that broke those macros triggering a NPE. It sounds like the .NET templating engine will revert that change in a future release but since there is no need for our templates to use those macros I have stripped their usages out to unblock users with .NET 7 RC installed.

### DIFF
--- a/Blueprints/BlueprintDefinitions/vs2022/AnnotationsFramework/template/src/BlueprintBaseName.1/serverless.template
+++ b/Blueprints/BlueprintDefinitions/vs2022/AnnotationsFramework/template/src/BlueprintBaseName.1/serverless.template
@@ -20,7 +20,7 @@
           "AWSLambdaBasicExecutionRole"
         ],
         "PackageType": "Zip",
-        "Handler": "BlueprintBaseName._1::BlueprintBaseName._1.Functions_Default_Generated::Default",
+        "Handler": "BlueprintBaseName.1::BlueprintBaseName._1.Functions_Default_Generated::Default",
         "Events": {
           "RootGet": {
             "Type": "HttpApi",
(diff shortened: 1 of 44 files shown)


Blueprints/BlueprintDefinitions/vs2022/AnnotationsFramework/template/src/BlueprintBaseName.1/serverless.template
```diff
@@ -20,7 +20,7 @@
           "AWSLambdaBasicExecutionRole"
         ],
         "PackageType": "Zip",
-        "Handler": "BlueprintBaseName._1::BlueprintBaseName._1.Functions_Default_Generated::Default",
+        "Handler": "BlueprintBaseName.1::BlueprintBaseName._1.Functions_Default_Generated::Default",
```
normj, line 23: This was a bug in the annotations template that it was using `BlueprintBaseName._1` instead of `BlueprintBaseName.1`. What this effect would have been is if the assembly name was `My-Assembly` by using `BlueprintBaseName._1` it would get translated incorrectly here to `My_Assembly`. Luckily for Annotations it doesn't really matter because the annotaton's source generator would have corrected the handler string.

ashovlin (approved):

philasmar (approved):

PR #1325, opened 2022-09-27
In the LambdaTestTool, adding the possibility to order functions by name in an alphabetical …

rsl-pseudocode:
*Description of changes:*

Feature to allow the possibility to order the dropdown of functions by name. Making it easier to navigate and test in repositories with lots of functions where functions isn't sorted by default.

### DIFF
--- a/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Pages/FunctionPickerComponent.razor
+++ b/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Pages/FunctionPickerComponent.razor
@@ -15,7 +15,10 @@
         </div>
     </div>
     <div class="col-sm form-group">
-        <label class="col-xs-3 control-label" for="functions-picker">Function</label>
+        <div class="col-xs-5">
+            <label class="control-label" for="functions-picker">Function</label>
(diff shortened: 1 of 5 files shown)


Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Pages/FunctionPickerComponent.razor
```diff
@@ -15,7 +15,10 @@
         </div>
     </div>
     <div class="col-sm form-group">
-        <label class="col-xs-3 control-label" for="functions-picker">Function</label>
+        <div class="col-xs-5">
+            <label class="control-label" for="functions-picker">Function</label>
+            <button class="float-right btn btn-primary btn-sm" id="functions-picker-sort-order" @onclick="SortFunctionsClicked" >Sort Functions</button>
```
normj, line 20: Should we just always sort the functions? I'm not sure if there is a lot of value for having the option of seeing them list in the order defined in the template. This way you are going to have to click the sort button every time you start the the tool which will probably get old fast.

rsl-pseudocode:
Adjusted slightly so we always sort straight out in the config parser. Button removed.

normj (approved):
I'm good with always sorting. The CloudFormation console sorts all resources in the resources tab.

PR #1326, opened 2022-09-29
Fix issue with test tool and Lambda functions using source generator serializer

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/1124

*Description of changes:*
The test tool uses reflection to call the Lambda function. This means when we get the response back from the Lambda function we get the response back as an object from the reflection API. The tool then sends the object into the ILambdaSerializer to convert the POCO into a stream. The serialize method is a generic method but since we have an object variable the generic parameter gets treated as a object instead of the actual type of the response. Using object as the generic parameter breaks the source generator based serializer because it uses the type information to look up the serialization code.

The fix is to use the reflection API on the ILambdaSerializer to turn the Serialize method into the correct generic method based on the runtime type of the Lambda response object.

The fix is in the `LambdaExecutor.cs` class. The rest of the changes are to add a test case and version bump the test tool. 


boblodgett (approved):

ashishdhingra (approved):
Looks good.

Issue #1279, opened 2022-08-18
Http end point ability in LamdaTestTool


nishant-newton:
### Describe the feature

We can debug the `Lambda function` from the `web page` of this `LamdaTestTool`. This new feature will allow to create the `http` end points so that `lambda functions` can be called through `http` in the local environment for `debugging` purpose

### Use Case

When `Lambda functions` are called through `http` protocol i can't `debug` them in `Visual Studio` and if there are multiple `lambda functions` involved in `microservice architecture pattern` then `debugging` is a nightmare

### Proposed Solution

Creating the `http` mapping for the `request`. The `tool` already parses the `yml` file and map the `handler` and `function` name. `yml` file also contains the `http` end point details for `AWS Api Gateway`. 
Once `http` requests are handled, the logic from `index.razor` file can be put in some `service class`.

### Other Information

The `Lamda functions` are tightly coupled with the infrastructure.  As a Developer i want to have an experience where i could debug solutions locally and if there are multiple projects involved and they make `http` calls to each, i should be able to `debug` them locally

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

LamdaTestTool does not use AWS SDK

### Targeted .NET Platform

.NET 3.1, .NEt 5.0, .NET 6.0

### Operating System and version

Windows, Linux, Mac

ashishdhingra:
Result of https://github.com/aws/aws-lambda-dotnet/issues/1169#issuecomment-1225015482. Needs review with the team. @nishant-newton Feel free to submit the PR which could be reviewed by the team.

nishant-newton:
Thanks @ashishdhingra 
PR submitted https://github.com/aws/aws-lambda-dotnet/pull/1330
@normj FYI

normj:
We have implemented this feature with our new test tool version and the .NET Aspire integration. Please try out the new Lambda .NET Aspire integration: https://github.com/aws/integrations-on-dotnet-aspire-for-aws/issues/17

PR #1330, opened 2022-10-03
Feature request 1279: Added http end point for executing lambda funct…

nishant-newton:
Reg feature request 1279, this PR contains the code which would enable this tool to have the feature where `Lambda functions` can be called through `http` also. So instead of using the web page provided by this tool and clicking on `Execute` button, users can also use `postman` or some other `services` in their `echo system` to call `Lambda functions` using `http`. This would be similar to calling `web api` end point in `.net`

*Issue #, [Feature Request 1279:](https://github.com/aws/aws-lambda-dotnet/issues/1279)*

*Description of changes:*
Ability to accept `http route path` configured in `serverless.yml` file and calling the corresponding `Lambda function`.
Able to read all type of `parameters` like `query`, `path` and `body` and pass it to `Lambda function`.
Able to accept `http verbs` `GET`, `POST`, `PUT` and `DELETE`. Whatever is configured in `serverless.yml` file, this tool can accept that `request`
No change has been done in the existing functionality

### DIFF
--- a/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Startup.cs
+++ b/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Startup.cs
@@ -1,13 +1,18 @@
 using Amazon.Lambda.TestTool.BlazorTester.Services;
+using Amazon.Lambda.TestTool.Services;
 using Blazored.Modal;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.FileProviders;
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
+using System.Net;
+using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -81,7 +86,8 @@ namespace Amazon.Lambda.TestTool.BlazorTester
             services.AddServerSideBlazor()
                     .AddHubOptions(options => options.MaximumReceiveMessageSize = null);
             services.AddHttpContextAccessor();
-
+            services.AddScoped<ILamdaService, LamdaService>();
+            services.AddSingleton<IAwsProfileConfig, AwsProfileConfig>();
             services.AddBlazoredModal();
 
             services.AddOptions<StaticFileOptions>()
@@ -95,7 +101,7 @@ namespace Amazon.Lambda.TestTool.BlazorTester
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILamdaService lamdaService, LocalLambdaOptions localLambdaOptions, IAwsProfileConfig awsProfileConfig)
         {
             app.UseDeveloperExceptionPage();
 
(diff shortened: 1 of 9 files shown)


Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LambdaFunctionInfo.cs
```diff
@@ -11,5 +11,15 @@ public class LambdaFunctionInfo
         /// The Lambda function handler string.
         /// </summary>
         public string Handler { get; set; }
+
+        /// <summary>
+        /// http path for the lambda function
+        /// </summary>
+        public string Path { get; set; }
```
normj, line 18: `Path` could mean a lot in this class's context. What about `HttpResourcePath`?

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LambdaFunctionInfo.cs
```diff
@@ -11,5 +11,15 @@ public class LambdaFunctionInfo
...
+
+        /// <summary>
+        /// The Lambda function method name.
+        /// </summary>
+        public string Method { get; set; }
```
normj, line 23: `Method` could mean a lot in this class's context. What about `HttpMethod`?

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LambdaDefaultsConfigFileParser.cs
```diff
@@ -189,12 +189,47 @@ private static void ProcessYamlServerlessTemplateFunctionBased(LambdaConfigInfo
                 
                 if (handler == null) continue;
                 if (string.IsNullOrEmpty(handler)) continue;
-                
-                
+
+                var events = resourceBody.Children.ContainsKey("events") ?
```
normj, line 193: Looks like you only have this implemented if using Serverless Framework. We also need equivalent versions in the CloudFormation JSON and YAML parse functions.

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Services/LamdaService.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            return response;
+        }
+
+        private string GetPayload(IDictionary<string, string> pathParameters, IDictionary<string, string> queryParameters, string json, IDictionary<string, string> headers, string httpMethod)
+        {
+            //APIGatewayProxyRequest APIGatewayProxyResponse
+            var authorizer = new APIGatewayCustomAuthorizerContext();
+            authorizer.Add("principalId", 3);
+            var request = new APIGatewayProxyRequest
```
normj, line 76: There needs to be some way of figuring out if this is an API Gateway REST API payload or an API Gateway HTTP API payload and then create the respective payload objects.

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
```diff
@@ -9,7 +9,9 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Amazon.Lambda.APIGatewayEvents" Version="2.5.0" />
```
normj, line 12: Bonus challenge in this tool is you can't take dependencies like `Amazon.Lambda.APIGatewayEvents` because it would then interfere with the version the Lambda function is depending on.

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Startup.cs
```diff
@@ -95,16 +101,70 @@ public void ConfigureServices(IServiceCollection services)
...
             app.UseStaticFiles();
 
             app.UseRouting();
 
+            app.UseCors();
+            var configInfo = awsProfileConfig.LambdaConfigInfo();
+
             app.UseEndpoints(endpoints =>
             {
+                if (configInfo != null)
```
normj, line 117: If the fields change in the UI like the chosen config file the mapped endpoints won't get updated. I wonder if it would be better to inject a middleware that could be more dynamic and look to see if the incoming request matches any functions currently available based on the current UI settings. If so the invoke the Lambda function otherwise return continue the middleware pipeline.

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Services/LamdaService.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        public LamdaService(LocalLambdaOptions options, IAwsProfileConfig awsProfileConfig)
+        {
+            _options = options;
+            _awsProfileConfig = awsProfileConfig;
+        }
+
+        public async Task<ExecutionResponse> Execute(string functionName, HttpContext context, IDictionary<string, string> pathParameters)
+        {
+            var awsProfile = "default";
```
normj, line 27: Profile and region need to come from the values configured in the UI.

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Services/LamdaService.cs
```diff
@@ -0,0 +1,93 @@
...
+            };
+            var response = await _options.LambdaRuntime.ExecuteLambdaFunctionAsync(request);
+
+            return response;
+        }
+
+        private string GetPayload(IDictionary<string, string> pathParameters, IDictionary<string, string> queryParameters, string json, IDictionary<string, string> headers, string httpMethod)
+        {
+            //APIGatewayProxyRequest APIGatewayProxyResponse
+            var authorizer = new APIGatewayCustomAuthorizerContext();
```
normj, line 74: I don't think we can just add a random `APIGatewayCustomAuthorizerContext`. That should be configured somewhere.

normj (changes requested):
I like the idea of this feature. I wonder if there should be a separate UI page in the application that allows you to configure this feature being enabled or not. The separate page could also be a good place for documentation to let users know the feature exist as well as configure some of the environment settings like whether an authorizer should be added to the request and what that authorizer should look like.

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LambdaFunctionInfo.cs
nishant-newton: Agreed. Will make the change

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LambdaFunctionInfo.cs
nishant-newton: Agreed. Will make the change

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LambdaDefaultsConfigFileParser.cs
nishant-newton: This solution contains the use case in my team. I think i'll have to look at the bigger picture and consider all the use cases here.

ashishdhingra:
@nishant-newton Please advise if you would be able to review and address review comments by @normj.

ashishdhingra:
Closing this PR since there is no activity.

Issue #1161, opened 2022-04-28
Invalid CodeUri generated by .NET Lambda Annotations Framework (Preview)


jaknor:
### Describe the bug

If you set the template path in `aws-lambda-tools-defaults.json` to a different folder (for example the folder one level above the projects with the lambda code), then the CodeUri path of the lambdas in the generated template is incorrect. The result of this is that commands such as sam build fails with an error message (`No .NET project found in directory`)

The expected result is that 

### Expected Behavior

The CodeUri path should be set relative to the template. 

I.E. if your lambda code exists in a folder named `AwsLambdaAnnotations` and  you set the template path to `../template.yaml` then the CodeUri in the generated template would need to be `./AwsLambdaAnnotations`

### Current Behavior

The CodeUri path is set incorrectly. I've tried 2 different scenarios

- Template path set to one folder above the project with the lambda code
  - CodeUri is set to `..`
- Template path set to a folder inside the project with the lambda code
   - CodeUri is set to `../..`

### Reproduction Steps

https://github.com/jaknor/AwsLambdaAnnotations

### Additional Information/Context

I [mentioned this problem](https://github.com/aws/aws-lambda-dotnet/issues/979#issuecomment-1111462305) in the Annotations design document and was asked to create a new issue for it

### AWS .NET SDK and/or Package version used

    <PackageReference Include="Amazon.Lambda.Core" Version="2.1.0" />
    <PackageReference Include="Amazon.Lambda.APIGatewayEvents" Version="2.4.0" />
    <PackageReference Include="Amazon.Lambda.Serialization.SystemTextJson" Version="2.3.0" />
    <PackageReference Include="Amazon.Lambda.Annotations" Version="0.5.0-preview" />

### Targeted .NET Platform

.NET 6

### Operating System and version

Windows 10

ashishdhingra:
The PR https://github.com/aws/aws-lambda-dotnet/pull/1332 is merged and released. Closing the issue. Feel free to open a new issue in case issue still exists.

PR #1332, opened 2022-10-05
fix: Correctly populate CodeUri when CloudFormation template is above the project root directory

96malhar:
*Issue #, if available:*
* DOTNET-6341
* https://github.com/aws/aws-lambda-dotnet/issues/1161

*Description of changes:*
This PR fixes the issue of an invalid `CodeUri` being generated when the CloudFormation template was above the project root directory.

The main logic is to call `new DirectoryInfo().FullName` on both arguments passed to `_directoryManager.GetRelativePath()` so that paths that end with `..` are resolved correctly.


Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/FileIO/DirectoryManager.cs
```diff
@@ -16,29 +16,34 @@ public class DirectoryManager : IDirectoryManager
...
             var pathDirs = path.Split('/');
+            var len = Math.Min(relativeToDirs.Length, pathDirs.Length);
 
-            int len = relativeToDirs.Length < pathDirs.Length ? relativeToDirs.Length : pathDirs.Length;
-
-            int lastCommonRoot = -1;
+            var lastCommonRoot = -1;
             int index;
 
             for (index = 0; index < len && string.Equals(relativeToDirs[index], pathDirs[index], StringComparison.OrdinalIgnoreCase); index++)
```
normj, line 43: On windows using ignoring case is correct but for all other platforms the string comparison should be case sensitive.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/FileIO/DirectoryManager.cs
```diff
@@ -16,29 +16,34 @@ public class DirectoryManager : IDirectoryManager
...
+            }
+
+            // If there is no common ancestor, just return the path.
+            if (!string.Equals(SanitizePath(relativeTo).Split('/')[0], SanitizePath(path).Split('/')[0]))
+            {
+                return path;
+            }
+
+            // Calculate the full paths by preserving the OS specific directory separator character and then perform the sanitization.
+            relativeTo = SanitizePath(new DirectoryInfo(relativeTo).FullName);
```
normj, line 33: If you move these 2 lines above the if block on line 25 you can remove the `SanitizePath` calls in the if condition.

normj (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/FileIO/DirectoryManager.cs
normj: Per offline conversations I understand this is intentional because we need to do the comparison before calling `new DirectoryInfo().FullName`.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/FileIO/DirectoryManager.cs
96malhar: Fixed

normj (approved):

philasmar (approved):

PR #1344, opened 2022-10-26
Fixed issue when Annotations when incorrect sync CF template when there are compile errors.

normj:
*Description of changes:*
The Lambda Annotations generator collects all of the known information in the project and syncs with the CloudFormation template. Any `LambdaFunction`s that are not found when the source generator runs are removed as orphans from the CloudFormation template.

If there was a compile error in the code then some of the `LambdaFunction`s might not have been parsed and then they would be removed as orphaned from the CloudFormation template. Once the compile error is fixed the `LambdaFunction` would reappear in the CloudFormation template but without any customizations that might have been applied to the lambda function resource in the CloudFormation template.

The fix is at the start of the generator look to see if any of the syntax trees, which equate to a source file, have any error diagnostics in the syntax tree. If there are then skip running the logic in the source generator. For our source generator we only care about generation when the project is complete. There is nothing we generate that affects the intellisense so we don't have need for incremental generation with a project that can't compile.
 

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
@@ -40,6 +40,18 @@ namespace Amazon.Lambda.Annotations.SourceGenerator
                     return;
                 }
 
+                // Check to see if any of the current syntax trees has any error diagnostics. If so
+                // Skip generation. We only want to sync the CloudFormation template if the project
+                // can compile.
+                foreach(var syntaxTree in context.Compilation.SyntaxTrees)
+                {
+                    if(syntaxTree.GetDiagnostics().Any(x => x.Severity == DiagnosticSeverity.Error))
+                    {
+                        return;
+                    }
+                }
+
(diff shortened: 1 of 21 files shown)


Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
```diff
@@ -40,6 +40,18 @@ public void Execute(GeneratorExecutionContext context)
                     return;
                 }
 
+                // Check to see if any of the current syntax trees of any error diagnostics. If so
```
boblodgett, line 43: Non blocking: "of any" should be "has any".

boblodgett (approved):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
normj: Fixed

96malhar (approved):

ganeshnj (approved):

Issue #1343, opened 2022-10-25
Invoke Lambda via Http API inside Lambda Test Tool


ggaller:
### Describe the feature

Adding http API to Lambda Test Tool as another way to call the function, along with web interface and CLI.

### Use Case

In our games we build the server part on the Lambda function. With HTTP endpoint we could write a mock client and connect to the function running from the IDE with the possibility of debugging. It would be a normal local server listening on `localhost`.

### Proposed Solution

I prepared a simple example that works. This is a controller with route to execute endpoint. It used injected `LocalLambdaOptions` and select default `LambdaConfigInfo` from it. Right now only payload is passed, but maybe we need to extend the api to be able to specify the function that we need to execute.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

-

### Targeted .NET Platform

.NET 6

### Operating System and version

Windows, Linux, MacOS

ashishdhingra:
@ggaller Thanks for submitting the feature request. This appears to be duplicate of https://github.com/aws/aws-lambda-dotnet/issues/1279 (which is also reviewed by the team). Please review and confirm if this ticket could be closed as duplicate.

ggaller:
Hi @ashishdhingra!
I looked at the #1279 and it seems to be a similar, but much more complex. 
I know that the AWS team often implements services with partially overlapping functionality to better meet customer needs and maybe this is the case))
My solution serves to support compatibility with Lambda [Invoke API](https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html) no more and no less. 
By adding a controller with these endpoints:
```csharp
[HttpPost("execute")]
public async Task<IActionResult> ExecuteFunction()
{...}

[HttpPost("execute/{functionName}")]
[HttpPost("2015-03-31/functions/{functionName}/invocations")]
public async Task<object> ExecuteFunction(string functionName)
{...}
```
Lambda Test Tool will be support for the standard lambda client from the `AWS SDK` like this:
```csharp
var client = new AmazonLambdaClient(new AmazonLambdaConfig
{
    EndpointProvider = new StaticEndpointProvider("http://localhost:5050/invokeapi/")
});
```
With support of transferring not only payload, but also logs (`LogType.Tail`) and errors similar like we get response from AWS.

What do you think of this approach? 

Code is ready and I could have a PR ready tomorrow.

[...]

PR #1349, opened 2022-10-28
Invoke Lambda via Http API inside Lambda Test Tool

ggaller:
This PR addressed to  #1343 

*Changes:*

1. Added [`InvokeApiController.cs`](Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Controllers/InvokeApiController.cs) to support [Invoke API](https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html) spec;
2. Added support for passing `ClientContext` to Lambda context.

### DIFF
--- a/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Controllers/InvokeApiController.cs
+++ b/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Controllers/InvokeApiController.cs
@@ -1,0 +1,239 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Text;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Amazon.Lambda.Core;
+using Amazon.Lambda.TestTool.Runtime;
+using Amazon.Lambda.TestTool.Runtime.LambdaMocks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Amazon.Lambda.TestTool.BlazorTester.Controllers
+{
+    [Route("[controller]")]
+    public class InvokeApiController : ControllerBase
+    {
+        private readonly LocalLambdaOptions _lambdaOptions;
+        private LambdaConfigInfo _lambdaConfig;
+
+        public InvokeApiController(LocalLambdaOptions lambdaOptions)
+        {
+            _lambdaOptions = lambdaOptions;
+        }
+
+        [HttpPost("execute")]
+        public async Task<IActionResult> ExecuteFunction()
+        {
+            if (!TryGetConfigFile(out var lambdaConfig))
+            {
+                return StatusCode((int)HttpStatusCode.InternalServerError,
+                    new InternalException("ServiceException", "Error while loading function configuration"));
+            }
+
+            if (lambdaConfig.FunctionInfos.Count == 0)
+            {
+                return NotFound(new InternalException("ResourceNotFoundException", "Default function not found"));
+            }
+
+            return Ok(await ExecuteFunctionInternal(lambdaConfig, lambdaConfig.FunctionInfos[0]));
+        }
+
+        [HttpPost("execute/{functionName}")]
+        [HttpPost("2015-03-31/functions/{functionName}/invocations")]
+        public async Task<object> ExecuteFunction(string functionName)
+        {
+            if (!TryGetConfigFile(out var lambdaConfig))
+            {
+                return StatusCode((int)HttpStatusCode.InternalServerError,
+                    new InternalException("ServiceException", "Error while loading function configuration"));
+            }
+
+            var functionInfo = lambdaConfig.FunctionInfos.FirstOrDefault(f => f.Name == functionName);
+            if (functionInfo == null)
+            {
+                return NotFound(new InternalException("ResourceNotFoundException",
+                    $"Function not found: {functionName}"));
+            }
+
+            return Ok(await ExecuteFunctionInternal(lambdaConfig, functionInfo));
+        }
+
+        private bool TryGetConfigFile(out LambdaConfigInfo configInfo)
+        {
+            configInfo = null;
+
+            if (_lambdaConfig != null)
+            {
+                configInfo = _lambdaConfig;
+                return true;
+            }
+
+            if (_lambdaOptions.LambdaConfigFiles.Count == 0)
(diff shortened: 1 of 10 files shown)


jeroenvervaeke (commented):
LGTM, got one small remark regarding readability

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Controllers/InvokeApiController.cs
```diff
@@ -0,0 +1,239 @@
...
+                if (errorLines.Length == 0)
+                {
+                    StackTrace = Array.Empty<string>();
+                    return;
+                }
+
+                StackTrace = errorLines.Skip(1).Select(s => s.Trim()).ToArray();
+
+                var errorMessage = errorLines[0];
+                var errorTypeDelimiterPos = errorMessage.IndexOf(':');
```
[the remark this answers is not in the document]
ggaller: Agree with you, fixed it.

jeroenvervaeke (approved):
LGTM

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Controllers/InvokeApiController.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            if (lambdaConfig.FunctionInfos.Count == 0)
+            {
+                return NotFound(new InternalException("ResourceNotFoundException", "Default function not found"));
+            }
+
+            return Ok(await ExecuteFunctionInternal(lambdaConfig, lambdaConfig.FunctionInfos[0]));
+        }
+
+        [HttpPost("execute/{functionName}")]
```
normj, line 44: Can we just have the `2015-03-31/functions/{functionName}/invocations` path. Somebody I would like this tool to have a sort of API Gateway emulator and I want to avoid any possible name collisions between the user's defined resource path to any of our resource paths.

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Controllers/InvokeApiController.cs
```diff
@@ -0,0 +1,239 @@
...
+        private async Task<object> ExecuteFunctionInternal(LambdaConfigInfo lambdaConfig,
+            LambdaFunctionInfo functionInfo)
+        {
+            var requestReader = new LambdaRequestReader(Request);
+            var function = _lambdaOptions.LoadLambdaFuntion(lambdaConfig, functionInfo.Handler);
+
+            var request = new ExecutionRequest
+            {
+                Function = function,
+                AWSProfile = lambdaConfig.AWSProfile,
```
normj, line 104: We need some way on either a per invoke or per session users can configure the profile and region. This feature won't work for users that don't want profiles and regions in their config file. I think it is fine using the config as default the. Also we might want to use the SDK's credential and region fallback chain if neither the config or a specific profile and region are not specified. Also can we add a log message which profile and region were used?

normj (changes requested):
Cool feature, sorry it took me so long to find time to look at it. We will need some test coverage for the feature. You can look at the `RuntimeApiControllerTests` on how I wrote tests for spinning up the tool and making API calls.

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Controllers/InvokeApiController.cs
ggaller: In [LambdaExecutor](https://github.com/aws/aws-lambda-dotnet/blob/master/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LambdaExecutor.cs#L31) these config parameters are passed to env variables if they are set. If you don't fill it in the config file, you can pass these parameters directly via environment variables when you start debugging. What we actively use, for example, through `launchSettings.json`.

At what level do you want to see `profile` and `region` logging? I propose to show the values of the corresponding environment variables at `LambdaExecutor` level. Because these are the values that will already be used later in the execution of the function

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Controllers/InvokeApiController.cs
```diff
@@ -0,0 +1,239 @@
...
+            if (lambdaConfig.FunctionInfos.Count == 0)
+            {
+                return NotFound(new InternalException("ResourceNotFoundException", "Default function not found"));
+            }
+
+            return Ok(await ExecuteFunctionInternal(lambdaConfig, lambdaConfig.FunctionInfos[0]));
+        }
+
+        [HttpPost("execute/{functionName}")]
+        [HttpPost("2015-03-31/functions/{functionName}/invocations")]
```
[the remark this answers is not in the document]
ggaller: I think the ability to call through a simple `HTTP API` is primary. For example we don't use the `AWS SDK` in `Unity`, instead we have a self-written client to call Lambda functions. The idea was originally born when I saw the `Blazor` client and thought that if a function could be called from the UI, there might be a ready-made API, but it turned out not.
So I think of it as an API for LambdaTestTool in addition to the UI. Compatibility with AWS SDK is a nice bonus 😃

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Controllers/InvokeApiController.cs
ggaller: See my comments below ⬇️

ggaller:
Added few integration tests

normj:
We are working on a new version of the test tool that does support this scenario. We just released the first early preview and admittedly our documentation is light and focused on Aspire integration. But the use case of using the SDK to invoke the Lambda function via the new test tool is supported. In fact part of the new tool is an API Gateway emulator that uses the SDK to invoke Lambda functions to the Lambda emulator side of the tool. https://github.com/aws/aws-lambda-dotnet/blob/master/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs#L159

I'm closing this PR since the focus is on our new tooling which will cover this use case.

Issue #1358, opened 2022-11-07
Allow NativeAOT to be executed as a part of .NET Lambda managed runtime


sagulati:
### Describe the feature

Currently for Executable Assemblies, the Amazon.Lambda.RuntimeSupport package looks for ${Executable_Assembly}.dll when initializing the lambda function. For NativeAOT and [Single File Executable Assemblies](https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/overview?tabs=cli), Update this to also look for "${Executable_Binary}" without ".dll". 

A link to the line in question can be seen here:
[https://github.com/aws/aws-lambda-dotnet/blob/0e8e42a0fe818f1adc1f97bb035f2535934c16b4/Libraries/src/Amazon.Lambda.RuntimeSupport/bootstrap.sh#L44](https://github.com/aws/aws-lambda-dotnet/blob/0e8e42a0fe818f1adc1f97bb035f2535934c16b4/Libraries/src/Amazon.Lambda.RuntimeSupport/bootstrap.sh#L44)

### Use Case

Allow NativeAOT and Single file executables to be run as a part of .NET Lambda managed runtime. Compiling the .NET executable assembly as NativeAOT generates a self-contained single compressed executable file that does not work in .NET managed runtime and the AWS Lambda Custom Runtime is required. 

### Proposed Solution

Update the bootstrap.sh file to add support for single executable file. 

If the Lambda_Handler does not contain ":" and {EXECUTABLE_ASSEMBLY}.dll is not found, then look for Single Executable file at path "${LAMBDA_TASK_ROOT}/${LAMBDA_HANDLER}" and execute it.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

Amazon.Lambda.RuntimeSupport 1.8.2

### Targeted .NET Platform

.NET6 / .NET 7

### Operating System and version

ASL2

ashishdhingra:
Appears to be a valid feature request. Needs review with the team.

sagulati:
@ashishdhingra [PR raised](https://github.com/aws/aws-lambda-dotnet/pull/1365).

ashishdhingra:
[...]

PR #1365, opened 2022-11-11
#1358 NativeAOT to be executed as a part of .NET Lambda managed runtime.

sagulati:
[#1358](#1358)

Allow NativeAOT to be executed as a part of .NET Lambda managed runtime


normj (approved):
Approved but might some slight adjustments to handle a Lambda exec wrapper could be used.

PR #1372, opened 2022-11-23
fix: added support for dynamic type in lambda annotations

### LINKED ISSUE #1304: Annotations - FunctionHander return type of Task<dynamic> Source Generation fails
### Describe the bug

As background we're attempting to wrap this [repo](https://github.com/oliverschenk/cognito-lambda-triggers-dotnet) in the Annotations Setup that we can leverage the DI code generation to get our C# service implementations into each Cognito trigger implementation.

This single handler uses a `Task<dynamic>` return type. This results in the following code generation error:
```Log
Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.Symbols.PublicModel.DynamicTypeSymbol' to type 'Microsoft.CodeAnalysis.INamedTypeSymbol'.   at Amazon.Lambda.Annotations.SourceGenerator.Models.TypeModelBuilder.Build(ITypeSymbol symbol, GeneratorExecutionContext context) [<path to csproj>]
```

The Functions/Handler class then looks like this:

``` C#
public class Functions
{
    private ILambdaLogger Logger;

    [LambdaSerializer(typeof(Amazon.Lambda.Serialization.SystemTextJson.CamelCaseLambdaJsonSerializer))]
    
    [LambdaFunction]
     public async Task<dynamic> FunctionHandler(JsonElement cognitoEvent, ILambdaContext context)
     {
         Logger = context.Logger;
    
         Logger.LogDebug("ENVIRONMENT VARIABLES: " + JsonSerializer.Serialize(System.Environment.GetEnvironmentVariables()));
         Logger.LogDebug("CONTEXT: " + JsonSerializer.Serialize(context));
    
         Logger.LogDebug("EVENT: " + JsonSerializer.Serialize(cognitoEvent));
    
         return await new CognitoEventHandlerFactory(Logger).GetHandler(cognitoEvent).HandleTriggerEventAsync();
     }
}
```

If this is resolvable we can then look to modify the example to pass our DI services though to the Trigger Handles.

Short term the workaround is to have a separate Function Handler with the explicit return types based on the given Trigger Sources combinations, and short circuit the dynamic behavior and deploy multiple Lambdas.

Perhaps in the longer term the Annotation library can use this larger scenario as a use case.

### Expected Behavior

Generate a suitable Handler wrapper to support the dynamic return type.

### Current Behavior

```Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.Symbols.PublicModel.DynamicTypeSymbol' to type 'Microsoft.CodeAnalysis.INamedTypeSymbol'.   at Amazon.Lambda.Annotations.SourceGenerator.Models.TypeModelBuilder.Build(ITypeSymbol symbol, GeneratorExecutionContext context) [<path to csproj>]```

### Reproduction Steps

clone repo as above and modify the Function handler as above.

### Possible Solution

as above

### Additional Information/Context

The value proposition is that, from a CDK perspective, we have a single Lambda Handler that services _all_ the triggers so wet don't have to deploy a separate Lambda `Function` with same code base for each Cognito Trigger for each Handler - for which there can be many.

As such supporting a `dynamic` return type on a single Function Handler has great value.

```
<PackageReference Include="Amazon.Lambda.Core" Version="2.1.0" />
<PackageReference Include="Amazon.Lambda.APIGatewayEvents" Version="2.4.1" />
<PackageReference Include="Amazon.Lambda.Serialization.SystemTextJson" Version="2.3.0" />
<PackageReference Include="Amazon.Lambda.Annotations" Version="0.7.0-preview" />
<PackageReference Include="AWSSDK.CognitoIdentityProvider" Version="3.7.5.10" />
```

### Targeted .NET Platform

NET 6

OSX Monterey, Windows 10, Linux

### PR DESCRIPTION
**Pull request by philasmar**
*Issue #, if available:*
DOTNET-6342
Fixes #1304 

*Description of changes:*
The Lambda Annotations Source Generator was failing with the error
```
Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.Symbols.PublicModel.DynamicTypeSymbol' to type 'Microsoft.CodeAnalysis.INamedTypeSymbol'.   at Amazon.Lambda.Annotations.SourceGenerator.Models.TypeModelBuilder.Build(ITypeSymbol symbol, GeneratorExecutionContext context) [<path to csproj>]
```
This happens when the type `dynamic` is used in a `LambdaFunction` such as
```
[LambdaFunction(PackageType = LambdaPackageType.Image)]
public dynamic DynamicReturn(string text, ILambdaContext context)
{
    context.Logger.LogLine(text);
    return text;
}

[LambdaFunction(PackageType = LambdaPackageType.Image)]
public string DynamicInput(dynamic text, ILambdaContext context)
{
    context.Logger.LogLine(text);
    return text;
}
```
The issue happens due to us casting to an `INamedTypeSymbol` in all cases.
This PR fixes the issue by casting to `INamedTypeSymbol` or `IDynamicTypeSymbol` when applicable.

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/TypeModelBuilder.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/TypeModelBuilder.cs
@@ -11,14 +11,34 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Models
     {
         public static TypeModel Build(ITypeSymbol symbol, GeneratorExecutionContext context)
         {
+            bool isGenericType;
+            if (symbol is INamedTypeSymbol)
+                isGenericType = ((INamedTypeSymbol)symbol)?.IsGenericType ?? false;
+            else
+                isGenericType = false;
+
+            IList<TypeModel> typeArguments;
+            if (symbol is INamedTypeSymbol)
+                typeArguments = ((INamedTypeSymbol)symbol)?.TypeArguments.Select(arg => Build(arg, context)).ToList();
+            else
+                typeArguments = new List<TypeModel>();
+
+            bool isEnumerable;
+            if (symbol is INamedTypeSymbol)
+                isEnumerable = ((INamedTypeSymbol)symbol)?.Interfaces.Any(it => it.Equals(context.Compilation.GetTypeByMetadataName(TypeFullNames.IEnumerable), SymbolEqualityComparer.Default)) ?? false;
+            else if (symbol is IDynamicTypeSymbol)
+                isEnumerable = ((IDynamicTypeSymbol)symbol)?.Interfaces.Any(it => it.Equals(context.Compilation.GetTypeByMetadataName(TypeFullNames.IEnumerable), SymbolEqualityComparer.Default)) ?? false;
+            else
+                isEnumerable = false;
+
             var model = new TypeModel
             {
                 Name = symbol.Name,
                 FullName = symbol.ToDisplayString(),
                 IsValueType = symbol.IsValueType,
-                IsGenericType = ((INamedTypeSymbol)symbol)?.IsGenericType ?? false,
-                TypeArguments = ((INamedTypeSymbol)symbol)?.TypeArguments.Select(arg => Build(arg, context)).ToList(),
-                IsEnumerable = ((INamedTypeSymbol)symbol)?.Interfaces.Any(it => it.Equals(context.Compilation.GetTypeByMetadataName(TypeFullNames.IEnumerable), SymbolEqualityComparer.Default)) ?? false
+                IsGenericType = isGenericType,
+                TypeArguments = typeArguments,
+                IsEnumerable = isEnumerable
             };
 
             return model;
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/TypeModelBuilder.cs [hunk 1]
```diff
@@ -11,14 +11,32 @@ public static class TypeModelBuilder
...
+            IList<TypeModel> typeArguments;
+            if (symbol is INamedTypeSymbol)
+                typeArguments = ((INamedTypeSymbol)symbol)?.TypeArguments.Select(arg => Build(arg, context)).ToList();
+            else
+                typeArguments = new List<TypeModel>();
+
+            bool isEnumerable;
+            if (symbol is INamedTypeSymbol)
+                isEnumerable = ((INamedTypeSymbol)symbol)?.Interfaces.Any(it => it.Equals(context.Compilation.GetTypeByMetadataName(TypeFullNames.IEnumerable), SymbolEqualityComparer.Default)) ?? false;
+            else
```
normj, line 29: Lets be defensive lets make this an `else if(symbol is IDynamicTypeSymbol)` and then I think the final `else` block would set `isEnumerable` to `false`.

normj (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/TypeModelBuilder.cs
philasmar: Fixed

normj (approved):

ashovlin (approved):

PR #1361, opened 2022-11-09
feat: Add a message to the CloudFormation template description when managed by Lambda Annotations

ashovlin:
*Issue #, if available:* DOTNET-6400

*Description of changes:*

When Lambda Annotations syncs the function resources to a CloudFormation template, this now also appends a string to the CloudFormation template description.

This string allows AWS to report on these templates to measure the usage of this framework, which will help AWS with:
1. Investigations and outreach if we find a critical bug.
2. Understanding our version adoption
3. Prioritizing improvements to this library against other .NET projects.

This is targeting a staging branch, I'll explore an opt-out mechanism in a subsequent PR.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1380, opened 2022-12-05
feat: Add opt-out flag to suppress modifying the CloudFormation template description field

ashovlin:
*Issue #, if available:* DOTNET-6400

*Description of changes:* This extends #1361 to add an opt-out mechanism for users who do not want Lambda Annotations to modify the CloudFormation template's description field.

This can be set via the following in the csproj:
```
<PropertyGroup>
  <AWSSuppressCloudFormationDescriptionModification>true</AWSSuppressCloudFormationDescriptionModification>
</PropertyGroup>
```
_Open to feedback on naming. Think @normj had proposed something different at one meeting but I didn't save it off._

This includes README updates explaining the opt-out mechanism and why we are adding the template description.


ik130 (changes requested):

normj (approved):

Libraries/src/Amazon.Lambda.Annotations/README.md
```diff
@@ -284,6 +284,26 @@ public class Functions
...
+  "AWSTemplateFormatVersion": "2010-09-09",
+  "Transform": "AWS::Serverless-2016-10-31",
+  "Description": "An AWS Serverless Application. This template is partially managed by Amazon.Lambda.Annotations (v0.9.0.0).",
+   ...
+```
+
+This version in the template allows AWS to report on these templates to measure the usage of this framework. This aids investigations and outreach if we find a critical bug, 
+helps understanding our version adoption, and allows us to prioritize improvements to this library against other .NET projects.
+
+If you do not want Lambda Annotations to modify the description, you can set the following in your project (csproj) file:
```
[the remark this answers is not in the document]
ashovlin: Merged the old and new sentences together somewhat to:
> If you do not want AWS to track the usage of the library, please set the following in your project (csproj) file:


Pushed up in an amended commit.

Libraries/src/Amazon.Lambda.Annotations/README.md
```diff
@@ -284,6 +284,26 @@ public class Functions
...
+field to include the version of Lambda Annotations that was used to modify the template.
+```
+{
+  "AWSTemplateFormatVersion": "2010-09-09",
+  "Transform": "AWS::Serverless-2016-10-31",
+  "Description": "An AWS Serverless Application. This template is partially managed by Amazon.Lambda.Annotations (v0.9.0.0).",
+   ...
+```
+
+This version in the template allows AWS to report on these templates to measure the usage of this framework. This aids investigations and outreach if we find a critical bug,
```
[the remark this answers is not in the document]
ashovlin: Per internal discussion tweaked slightly to replace "request level" with "CloudFormation stack" since this is unlike our usual useragent metrics.
> This description allow AWS to record the version and the usage of the Lambda Annotations framework in order to improve its quality. We record details at the CloudFormation stack level, and do not identify the application, library or tool that was deployed. Note that we do not record any personal information, such as usernames, email addresses or sensitive project-level information. 

Pushed up in an amended commit.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
```diff
@@ -127,6 +127,9 @@ public void Execute(GeneratorExecutionContext context)
                 {
                     annotationReport.CloudFormationTemplatePath = templateHandler.FindTemplate(receiver.ProjectDirectory);
                     annotationReport.ProjectRootDirectory = receiver.ProjectDirectory;
+                    annotationReport.IsCloudFormationDescriptionModificationSuppressed =
```
philasmar, line 130: nit: I'm fine with the implementation, though I'd prefer if naming referred to the purpose instead of implementation description. So instead of `IsCloudFormationDescriptionModificationSuppressed`, I'd prefer if it was `IsTelemetrySuppressed` or something like that. Same applies to `ProjectFileHandler` function `IsCloudFormationDescriptionModificationSuppressed`.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/AnnotationReport.cs
```diff
@@ -4,8 +4,25 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Models
...
+        /// <summary>
+        /// Path to the directory containing the csproj file for the Lambda project
+        /// </summary>
         public string ProjectRootDirectory { get; set; }
+
+        /// <summary>
+        /// Whether the flag to suppress modifying the CloudFormation template 
+        /// description is set to true in the Lambda project csproj
+        /// </summary>
+        public bool IsCloudFormationDescriptionModificationSuppressed { get; set; } = false;
```
philasmar, line 26: nit: No need for the explicit assignment to `false` as it is false by default.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/ProjectFileHandler.cs
```diff
@@ -0,0 +1,47 @@
+﻿using Amazon.Lambda.Annotations.SourceGenerator.FileIO;
+using System;
+using System.Xml;
+
+namespace Amazon.Lambda.Annotations.SourceGenerator
+{
+    /// <summary>
+    /// Utilities for working with project (.csproj) files, such as allowing projects
```
philasmar, line 8: nit: I'd exclude "such as allowing projects to opt out of the CloudFormation description modification." from the class description as it is implying the class is more specific than it's supposed to be.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/ProjectFileHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// to opt out of the CloudFormation description modification.
+    /// </summary>
+    public class ProjectFileHandler
+    {
+        private const string OptOutNodeXpath = "//PropertyGroup/AWSSuppressCloudFormationDescriptionModification";
```
philasmar, line 13: nit: similar to my above comments, i'd prefer if we make the attribute name shorter and refer to what it's purpose actually is instead of an implementation detail. So maybe name it `AWSSuppressTelemetry`?

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/ProjectFileHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        /// <summary>
+        /// Determines if the project has opted out of modifying the CloudFormation template descriptio
```
philasmar, line 16: nit: you are missing an "n" at the end of "descriptio". I'd prefer if the entire description would change to something like "Determines if the project has opted out of providing telemetry to AWS"

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/ProjectFileHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// </summary>
+        /// <param name="projectFilePath">Path to a .csproj file</param>
+        /// <param name="fileManager">FileManager instance used to read the csproj contents</param>
+        /// <returns>True if opted out of modifying the description, false otherwise</returns>
+        public static bool IsCloudFormationDescriptionModificationSuppressed(string projectFilePath, IFileManager fileManager)
```
philasmar, line 21: nit: same comment about renaming this method

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
ashovlin: Went with `AWSSuppressLambdaAnnotationsTelemetry` after internal discussion. Fixed in amended commit.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/AnnotationReport.cs
ashovlin: Removed in amended commit.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/ProjectFileHandler.cs
ashovlin: Went with `AWSSuppressLambdaAnnotationsTelemetry` after internal discussion. Fixed in amended commit.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/ProjectFileHandler.cs
ashovlin: Fixed in amended commit.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/ProjectFileHandler.cs
ashovlin: Fixed in amended commit.
>         /// Determines if the project has opted out of any Lambda Annotations telemetry

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/ProjectFileHandler.cs
ashovlin: Fixed in amended commit.

philasmar (approved):

PR #1397, opened 2022-12-09
Add the word "Lambda" to templates that don't have it. Issue 1396.

bjhogan:
*Issue #, if available:*
1396
*Description of changes:*

Some templates didn't have the word "Lambda" in them, so they don't show when dotnet new lambda --list is run.

### DIFF
--- a/Blueprints/BlueprintDefinitions/vs2017/DetectImageLabelsServerless-FSharp/template/.template.config/template.json
+++ b/Blueprints/BlueprintDefinitions/vs2017/DetectImageLabelsServerless-FSharp/template/.template.config/template.json
@@ -5,7 +5,7 @@
     "Lambda",
     "Serverless"
   ],
-  "name": "Serverless Detect Image Labels",
+  "name": "Lambda Serverless Detect Image Labels",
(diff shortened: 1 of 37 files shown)


normj:
We updated the [README](https://github.com/aws/aws-lambda-dotnet#blueprints) file to how to properly search for our templates. I don't think we need to prefix every template with the word Lambda so I'm going to close the PR.

bjhogan:
Do we have any feedback on how users discover new templates? If they filter on the word "Lambda", as the instructions used to say, they may not find some of the more recent templates.

normj:
I haven't heard any feedback one way or the other. You following the comment in the README file was the first comment I had heard about this in the years that comment has been in the README.

normj:
Sorry I should have been more explicit. The main reason I don't really want to do this is it makes the names quite long. In Visual Studio the project wizard would likely start to cut off the meaningful part of the names of some of these projects to force the Lambda prefix in the name.

bjhogan:
What if the word "Lambda" was added to the end of the name, that way command line users could easily find the templates.

Issue #1233, opened 2022-07-04
Amazon.Lambda.AspNetCoreServer.APIGatewayProxyFunction Doesn't replace path params when proxy+ is present


erlinghi:
### Describe the bug

Looking at the code of Amazon.Lambda.AspNetCoreServer.APIGatewayProxyFunction 
```
      if (apiGatewayRequest.PathParameters != null && apiGatewayRequest.PathParameters.ContainsKey("proxy") && !string.IsNullOrEmpty(apiGatewayRequest.Resource))
      {
        string pathParameter = apiGatewayRequest.PathParameters["proxy"];
        str1 = apiGatewayRequest.Resource.Replace("{proxy+}", pathParameter);
      }
      if (string.IsNullOrEmpty(str1))
        str1 = apiGatewayRequest.Path;
      if (!str1.StartsWith("/"))
        str1 = "/" + str1;
      aspNetCoreRequestFeature.Path = str1;
```
It doesn't replace other path params if proxy+ is present

### Expected Behavior

I expect path params before the proxy+ to get replaced correctly.
`
HTTP “GET” "/site/2626/status/testEndpoint/123123/asdasdasd"
`

### Current Behavior

You can see that if you have a resource with proxy+, it doesn't replace other path parameters in the resource.
So if you have a resource like this:
`
/site/{siteId}/{proxy+} 
`
And do a call like this
`
HTTP “GET” "/site/2626/status/testEndpoint/123123/asdasdasd"
`
your lambda get hit with this
`
HTTP “GET” “/site/{siteId}/status/testEndpoint/123123/asdasdasd”
`

### Reproduction Steps

Setup a resource in your apigateway with both path params and a greedy path param.
Example
`
/site/{siteId}/{proxy+} 
`
Then APIGatewayProxyFunction won't pass on the value you use for siteId.

### Possible Solution

A fix is to replace all path params in the path, if proxy+ is present, else do like before.

You can see our current workaround in our entrypoint
```
public class LambdaEntryPoint : APIGatewayProxyFunction
{
    static LambdaEntryPoint()
    {
        LogConfigurator.ConfigureSerilog();
    }

    protected override void MarshallRequest(InvokeFeatures features, APIGatewayProxyRequest apiGatewayRequest, ILambdaContext lambdaContext)
    {
        UseFixForProxyReplacementBugInApiGatewayProxyFunction(apiGatewayRequest);

[...]

PR #1403, opened 2022-12-24
Adds non greedy parameter values into the path when {proxy+}

christostatitzikidis:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/1233

A confirmed bug since July, that has not been fixed yet, and has taken me a whole day until I realised what the problem is.
Basically, when {proxy+}, the values of the non-greedy parameters are not picked up from the request path. 

So the following scenario **_fails_**:

Given an API Gateway resource set up as `/path/{foo}/bar/{proxy+}`
When I make a request to path `/path/afoo/bar/subpath`
Then the `foo` variable in my controller action should be equal to `afoo`

What actually happens is that my controller's `foo` variable contains the value `{foo}`

*Description of changes:*
Added code that will actually replace the non-greedy parameters with their values in the path.
In order to keep changes as small as possible, I just added a loop, but I think it would be better optimised if the path was replaced straight away with the incoming request's path - that's already what happens with non-proxy requests ((line 176). I added a comment about that.

The pull request refers to lines `156-177` of `ApiGatewayProxyFunction.cs`

### DIFF
--- a/Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
+++ b/Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
@@ -168,8 +168,7 @@ namespace Amazon.Lambda.AspNetCoreServer
                         path = path.Replace($"{{{pathParameter.Key}}}", pathParameter.Value);
                     }
                 }
-                // As a matter of fact, is there a reason for the if clause above? A simpler/faster solution would be to straight on replace the path as below
-
+ 
                 if (string.IsNullOrEmpty(path))
                 {
                     path = apiGatewayRequest.Path;
--- a/Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
+++ b/Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
@@ -154,12 +154,21 @@ namespace Amazon.Lambda.AspNetCoreServer
                 requestFeatures.Method = apiGatewayRequest.HttpMethod;
 
                 string path = null;
+
+                // Replaces {proxy+} in path, if exists
                 if (apiGatewayRequest.PathParameters != null && apiGatewayRequest.PathParameters.TryGetValue("proxy", out var proxy) &&
                     !string.IsNullOrEmpty(apiGatewayRequest.Resource))
                 {
                     var proxyPath = proxy;
                     path = apiGatewayRequest.Resource.Replace("{proxy+}", proxyPath);
+
+                    // Adds all the rest of non greedy parameters in apiGateway.Resource to the path
+                    foreach (var pathParameter in apiGatewayRequest.PathParameters.Where(pp => pp.Key != "proxy"))
+                    {
+                        path = path.Replace($"{{{pathParameter.Key}}}", pathParameter.Value);
+                    }
                 }
+                // As a matter of fact, is there a reason for the if clause above? A simpler/faster solution would be to straight on replace the path as below
 
                 if (string.IsNullOrEmpty(path))
                 {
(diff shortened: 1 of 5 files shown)


Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
```diff
@@ -154,12 +154,21 @@ protected override void MarshallRequest(InvokeFeatures features, APIGatewayProxy
                 requestFeatures.Method = apiGatewayRequest.HttpMethod;
 
                 string path = null;
+
+                // Replaces {proxy+} in path, if exists
```
normj, line 158: Seems like we should always be doing this logic regardless over the being a `{proxy+}`. It is possible to use this package and not use `{proxy+}` and use path parameters..

normj (changes requested):
This for this bug fix. We also need some tests for this fix. Tests for this project are here and are relatively easy to add. Basically just create your sample event json and a corresponding endpoint in the test project.

Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
```diff
@@ -154,12 +154,21 @@ protected override void MarshallRequest(InvokeFeatures features, APIGatewayProxy
...
                     var proxyPath = proxy;
                     path = apiGatewayRequest.Resource.Replace("{proxy+}", proxyPath);
+
+                    // Adds all the rest of non greedy parameters in apiGateway.Resource to the path
+                    foreach (var pathParameter in apiGatewayRequest.PathParameters.Where(pp => pp.Key != "proxy"))
+                    {
+                        path = path.Replace($"{{{pathParameter.Key}}}", pathParameter.Value);
+                    }
                 }
+                // As a matter of fact, is there a reason for the if clause above? A simpler/faster solution would be to straight on replace the path as below
```
christostatitzikidis, line 172: I think all the logic, between lines 159 and 169 is not needed at all. 
If that logic is removed, both {proxy+} and non proxy paths will be replaced just fine on line 175.

In fact this whole block of code:
`string path = null;
if (apiGatewayRequest.PathParameters != null && apiGatewayRequest.PathParameters.TryGetValue("proxy", out var proxy) &&
    !string.IsNullOrEmpty(apiGatewayRequest.Resource))
{
    var proxyPath = proxy;
    path = apiGatewayRequest.Resource.Replace("{proxy+}", proxyPath);
}

if (string.IsNullOrEmpty(path))
{
    path = apiGatewayRequest.Path;
}`

should just become:
`string path = apiGatewayRequest.Path;`

This would fix the issue, and be a lot cleaner

Libraries/src/Amazon.Lambda.AspNetCoreServer/APIGatewayProxyFunction.cs
christostatitzikidis: Agreed with the above.

christostatitzikidis: I double checked and we are actually also using non-greedy parameters in non-proxy resource paths in our custom deployments that contain this PR.
This works just fine because the path does never get in the {proxy+} logic and gets replaced properly by line 175 
```
if (string.IsNullOrEmpty(path))
{
    path = apiGatewayRequest.Path;
}
```

I have added an additional test to cover this case, and it passes without any modifications to the PR

normj (approved):

normj:
Thanks for the PR. The PR has been released as part of version 8.0.0 of Amazon.Lambda.AspNetCoreServer. This was a major version bump to PR #899 shipped in this release which is a slight breaking change.

PR #1404, opened 2022-12-28
Emit the milti-line logs with timestamps

luben:
*Description of changes:*

Lambda accepts multi-line messages where the timestamp of the log emission is encoded in the log event header. This helps ordering the function emitted messages with regards to platform emitted messages, e.g. START, REPORT.


alexey-katranov (approved):
nit: testing: you can wrap `Write` with two timestamps (before and after) to check that the header contains the timestamp within the range.

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
```diff
@@ -32,8 +32,9 @@ public static class FileDescriptorLogFactory
 
         // max cloudwatch log event size, 256k - 26 bytes of overhead.
         internal const int MaxCloudWatchLogEventSize = 256 * 1024 - 26;
-        internal const int LambdaTelemetryLogHeaderLength = 8;
-        internal const uint LambdaTelemetryLogHeaderFrameType = 0xa55a0001;
+        internal const int LambdaTelemetryLogHeaderLength = 16;
+        internal const uint LambdaTelemetryLogHeaderFrameType = 0xa55a0003;
+        internal const DateTimeOffset UnixEpoch = new DateTimeOffset(1970, 1, 1, 0, 0, 0, TimeSpan.Zero);
```
normj, line 37: This failed to compile because you can't use `const` when newing a new `DateTimeOffset`. Replace `const` here with `static readonly`.

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/FileDescriptorLogStream.cs
```diff
@@ -107,12 +110,14 @@ public override void Write(byte[] buffer, int offset, int count)
                 }
 
                 var typeAndLength = ArrayPool<byte>.Shared.Rent(LambdaTelemetryLogHeaderLength);
+                var now = (DateTimeOffset.UtcNow - UnixEpoch).Ticks / 10; // There are 10 tick per microsecond
                 try
                 {
                     Buffer.BlockCopy(_frameTypeBytes, 0, typeAndLength, 0, 4);
                     Buffer.BlockCopy(messageLengthBytes, 0, typeAndLength, 4, 4);
+                    Buffer.BlockCopy(now, 0, typeAndLength, 8, 8);
```
normj, line 118: This does not compile because `now` is a `long` but the `BlockCopy` method needs a `byte[]`. Below is how I changed the function to compile.

```
            public override void Write(byte[] buffer, int offset, int count)
            {
                var messageLengthBytes = BitConverter.GetBytes(count - offset);
                if (BitConverter.IsLittleEndian)
                {
                    Array.Reverse(messageLengthBytes);
                }

                var now = (DateTimeOffset.UtcNow - UnixEpoch).Ticks / 10; // There are 10 tick per microsecond
                var nowInBytes = BitConverter.GetBytes(now);
                if (BitConverter.IsLittleEndian)
                {
                    Array.Reverse(nowInBytes);
                }

                var typeAndLength = ArrayPool<byte>.Shared.Rent(LambdaTelemetryLogHeaderLength);
                try
                {
                    Buffer.BlockCopy(_frameTypeBytes, 0, typeAndLength, 0, 4);
                    Buffer.BlockCopy(messageLengthBytes, 0, typeAndLength, 4, 4);
                    Buffer.BlockCopy(nowInBytes, 0, typeAndLength, 8, 8);

                    _fileDescriptorStream.Write(typeAndLength, 0, LambdaTelemetryLogHeaderLength);
                    _fileDescriptorStream.Write(buffer, offset, count);
                    _fileDescriptorStream.Flush();
                }
                finally
                {
                    ArrayPool<byte>.Shared.Return(typeAndLength);
                }
            }
```

normj (changes requested):
After I made suggest changes I was able to deploy a function to the .NET 6 managed runtime with the current released runtime and with this version of the runtime and saw the log output in CloudWatch logs was the same between the runtimes. I'm not exactly sure how to verify the new timestamp is being consumed but sending it along with the log message is not causing any problems.

luben:
Thanks for the suggested fixes. I have applied them.

Yes, there should not be any difference in the log messages. The only difference may be in how messages are ordered. For example, without this patch, some the the function logs may appear after the "REPORT" line - the function and platform messages are read from different sockets and there is no inherent ordering between them, so we used the time of arrival (when we read the message). In order to fix this we added timestamps - already done for the platform messages, this change adds timestamps to the runtime generated messages. This will not fix through logs written to stdout/stderr as these are line terminated and we cannot send additional data in the framing, like the timestamp here.

Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.UnitTests/FileDescriptorLogStreamTests.cs
```diff
@@ -16,7 +16,7 @@ public class FileDescriptorLogStreamTests
 
         private static readonly byte[] ExpectedMagicBytes =
         {
-            0xA5, 0x5A, 0x00, 0x01
+            0xA5, 0x5A, 0x00, 0x03
```
ashovlin, line 19: GitHub won't let me comment on it directly, but this test is failing for me locally with this branch checked out:

```
Amazon.Lambda.RuntimeSupport.UnitTests.FileDescriptorLogStreamTests.MultilineLoggingInSingleLogEntryWithTlvFormat
   Source: FileDescriptorLogStreamTests.cs line 23
   Duration: 33 ms

  Message: 
Assert.Equal() Failure
Expected: Byte[] [0, 0, 0, 41]
Actual:   Byte[] [83, 171, 221, 67]

  Stack Trace: 
FileDescriptorLogStreamTests.AssertHeaderBytes(Byte[] buf, Byte[] expectedLengthBytes) line 190
FileDescriptorLogStreamTests.MultilineLoggingInSingleLogEntryWithTlvFormat() line 66
```

ashovlin (changes requested):

normj (approved):
Approving assuming the test that Alex pointed out gets updated.

Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.UnitTests/FileDescriptorLogStreamTests.cs
ashovlin: Moved this to follow-up https://github.com/aws/aws-lambda-dotnet/pull/1409

ashovlin (approved):
Approving, CI failure is expected since this is from a fork.

PR #1409, opened 2023-01-13
Fix file descriptor log stream tests

ashovlin:
**Note:** Builds on #1404 - needs to be rebased and merged after it. Only the top commit is new.

*Description of changes:*
1. Now that we're emitting the timestamp in the log message headers, `AssertHeaderBytes` needed to be updated to extract the _next 4_ bytes for the message length, as opposed to the _last 4_  (which are now the timestamp).
2. Addressed build warnings on this file around collection assertions.


luben:
LGTM

normj (approved):

Issue #979, opened 2021-11-17
Design Doc: Lambda Annotation


normj:
The AWS .NET team is working on a new library for constructing .NET Lambda functions. The design doc can be viewed and commented on in this PR https://github.com/aws/aws-lambda-dotnet/pull/961. 

The high level overview of the new library is to use .NET source generator support to translating from the low level single event object programming model of Lambda to an experience similar to ASP.NET Core but with minimal overhead. The initial work is focused on REST API Lambda functions but the technology is applicable for other Lambda event types.

For example developers will be able to write a Lambda function like the following with the `ICalculator ` service being injected by dependency injection and and the `LambdaFunction` and `HttpApi` attributes directing the source generator to generator the compatible Lambda boiler plate code and sync with the CloudFormation template.
```csharp
public class Functions
{
	ICalculator _calulator;

	public Functions(ICalculator calculator)
	{
		_calulator = calculator;
	}

	[LambdaFunction]
	[HttpApi(HttpMethod.Get, HttpApiVersion.V2, "/add/{x}/{y}")]
	public int Add(int x, int y)
	{
		return _calulator.Add(x, y);
	}
}
```

We would appreciate feedback on the design. What use cases what you like this library to help solve and what boiler plate code can we remove from your applications to make it simple to write Lambda functions.

----------
**Update 12/21/2021**

We are excited to announce that Lambda Annotations first preview is here. Developers using the preview version of [Amazon.Lambda.Annotations NuGet package](https://www.nuget.org/packages/Amazon.Lambda.Annotations) can start using the simplified programming model to write REST and HTTP API Lambda functions on .NET 6. (Only Image package type is supported as of now)

Example ([sample project](https://github.com/ganeshnj/lambda-annotations-sample))

```csharp
[LambdaFunction(Name = "CalculatorAdd", PackageType = LambdaPackageType.Image)]
[RestApi(HttpMethod.Get, "/Calculator/Add/{x}/{y}")]
public int Add(int x, int y, [FromServices]ICalculatorService calculatorService)
{
    return calculatorService.Add(x, y);
}
```
```csharp
[LambdaStartup]
public class Startup
{
    public void ConfigureServices(IServiceCollection services)
    {
        services.AddScoped<ICalculatorService, CalculatorService>();
    }
}
```

[...]

PR #1421, opened 2023-01-27
Lambda Annotations: Add support for returning non 200 responses.

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/979#issuecomment-1218455773
https://github.com/aws/aws-lambda-dotnet/discussions/1379

*Description of changes:*
ASP.NET Core has the [TypesResults](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.typedresults) class which is used to create an `IResult` for the desired HTTP response. Using that for inspiration this PR creates `HttpResults` and `IHttpResults` to allow users using the Annotations library to customize the HTTP response.

The default experience without using `IHttpResults` is users return back their own types like the following:
```
    [LambdaFunction()]
    [HttpApi(LambdaHttpMethod.Get, "/")]
    public IHttpResults SimpleExample()
    {
        return new OrderInfo();
    }
```
Annotations will treat the response as a 200 success and serialize the `OrderInfo` as JSON.

Using `HttpResults` you can do same thing and include headers.
```
    [LambdaFunction()]
    [HttpApi(LambdaHttpMethod.Get, "/")]
    public IHttpResults SimpleExample()
    {
        return HttpResults.Ok(new OrderInfo())
                                      .AddHeader("MyHeader", "Value");
    }
```
The `IHttpResults` has the `AddHeader` method that returns back the instance of `IHttpResults` to follow a fluent building pattern.

There are methods for on `HttpResults` for each of the typical status codes like `TypesResults`. Here is a not found example:

```
    [LambdaFunction()]
    [HttpApi(LambdaHttpMethod.Get, "/{orderId}")]
    public IHttpResults GetOrder(int orderId)
    {
        return HttpResults.NotFound($"Failed to find order for {orderId}");
    }
```

## Other refactoring changes
* Added .NET 6 target framework to `Amazon.Lambda.Annotations`. It still needs to target .NET Standard 2.0 when used inside the source generator. This causes some tricky snipping out of .NET 6 code.
* Found a bug when the Lambda function was returning a `Task` but not using the `async` keyword. I got rid of the logic based on the `IsAsync` property and created new `Task` properties to identify when a `Task` is used and when the task is returning void or a type.
* Updated source generator test project to target .NET 6.

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
@@ -63,7 +63,7 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
 #>
         }
 
-        public <#= _model.LambdaMethod.IsAsync ? "async " : "" #><#= _model.GeneratedMethod.ReturnType.FullName #> <#= _model.LambdaMethod.Name #>(<#= string.Join(", ", _model.GeneratedMethod.Parameters.Select(p => $"{p.Type.FullName} {p.Name}")) #>)
+        public <#= _model.LambdaMethod.ReturnsVoidOrGenericTask ? "async " : "" #><#= _model.GeneratedMethod.ReturnType.FullName #> <#= _model.LambdaMethod.Name #>(<#= string.Join(", ", _model.GeneratedMethod.Parameters.Select(p => $"{p.Type.FullName} {p.Name}")) #>)
         {
 <#
     if (_model.LambdaMethod.UsingDependencyInjection)
@@ -353,7 +353,7 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
             // return 400 Bad Request if there exists a validation error
             if (validationErrors.Any())
             {
-                return new <#= _model.LambdaMethod.IsAsync ? _model.GeneratedMethod.ReturnType.TaskTypeArgument : _model.GeneratedMethod.ReturnType.FullName #>
+                var errorResult = new <#= restApiAttribute != null || httpApiAttribute?.Data.Version == Amazon.Lambda.Annotations.APIGateway.HttpApiVersion.V1 ? "Amazon.Lambda.APIGatewayEvents.APIGatewayProxyResponse" : "Amazon.Lambda.APIGatewayEvents.APIGatewayHttpApiV2ProxyResponse" #>
                 {
                     Body = @$"{{""message"": ""{validationErrors.Count} validation error(s) detected: {string.Join(",", validationErrors)}""}}",
                     Headers = new Dictionary<string, string>
@@ -363,18 +363,44 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
                     },
                     StatusCode = 400
                 };
+<#
+                if(_model.LambdaMethod.ReturnsIHttpResults)
+                {
+#>
+                var errorStream = new System.IO.MemoryStream();
+                System.Text.Json.JsonSerializer.Serialize(errorStream, errorResult);
+                return errorStream;
+<#
+                }
+                else
+                {
+#>
+                return errorResult;
+<#
+                }
+#>
             }
 
 <#
         }
 
-        if (_model.LambdaMethod.ReturnsVoid)
+        if (_model.LambdaMethod.ReturnsIHttpResults)
+        {
+#>
+            var httpResults = <#= _model.LambdaMethod.ReturnsGenericTask ? "await " : "" #><#= _model.LambdaMethod.ContainingType.Name.ToCamelCase() #>.<#= _model.LambdaMethod.Name #>(<#= parameters #>);
+            HttpResultSerializationOptions.ProtocolFormat serializationFormat = <#= restApiAttribute != null ? "HttpResultSerializationOptions.ProtocolFormat.RestApi" : "HttpResultSerializationOptions.ProtocolFormat.HttpApi"#>;
+            HttpResultSerializationOptions.ProtocolVersion serializationVersion = <#= restApiAttribute != null || httpApiAttribute?.Data.Version == Amazon.Lambda.Annotations.APIGateway.HttpApiVersion.V1 ? "HttpResultSerializationOptions.ProtocolVersion.V1" : "HttpResultSerializationOptions.ProtocolVersion.V2"#>;
+            var serializationOptions = new HttpResultSerializationOptions { Format = serializationFormat, Version = serializationVersion };
+            var response = httpResults.Serialize(serializationOptions);
+<#
+        }
+        else if (_model.LambdaMethod.ReturnsVoid)
         {
 #>
             <#= _model.LambdaMethod.ContainingType.Name.ToCamelCase() #>.<#= _model.LambdaMethod.Name #>(<#= parameters #>);
 <#
         }
-        else if (_model.LambdaMethod.ReturnsTask)
+        else if (_model.LambdaMethod.ReturnsVoidTask)
         {
 #>
             await <#= _model.LambdaMethod.ContainingType.Name.ToCamelCase() #>.<#= _model.LambdaMethod.Name #>(<#= parameters #>);
@@ -383,11 +409,11 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
         else
         {
 #>
-            var response = <#= _model.LambdaMethod.IsAsync ? "await " : "" #><#= _model.LambdaMethod.ContainingType.Name.ToCamelCase() #>.<#= _model.LambdaMethod.Name #>(<#= parameters #>);
+            var response = <#= _model.LambdaMethod.ReturnsGenericTask ? "await " : "" #><#= _model.LambdaMethod.ContainingType.Name.ToCamelCase() #>.<#= _model.LambdaMethod.Name #>(<#= parameters #>);
 <#
         }
 
-        if (_model.GeneratedMethod.ReturnType.FullName == _model.LambdaMethod.ReturnType.FullName)
+        if (_model.GeneratedMethod.ReturnType.FullName == _model.LambdaMethod.ReturnType.FullName || _model.LambdaMethod.ReturnsIHttpResults)
         {
 #>
             return response;
@@ -395,7 +421,7 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
         }
         else
         {
-            if (!_model.LambdaMethod.ReturnsVoid && !_model.LambdaMethod.ReturnsTask)
+            if (!_model.LambdaMethod.ReturnsVoid && !_model.LambdaMethod.ReturnsVoidTask)
             {
                 if (_model.LambdaMethod.ReturnType.IsValueType)
                 {
@@ -418,10 +444,10 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
             }
 #>
 
-            return new <#= _model.LambdaMethod.IsAsync ? _model.GeneratedMethod.ReturnType.TaskTypeArgument : _model.GeneratedMethod.ReturnType.FullName #>
+            return new <#= _model.LambdaMethod.ReturnsVoidOrGenericTask ? _model.GeneratedMethod.ReturnType.TaskTypeArgument : _model.GeneratedMethod.ReturnType.FullName #>
             {
 <#
-            if (!_model.LambdaMethod.ReturnsVoid && !_model.LambdaMethod.ReturnsTask)
+            if (!_model.LambdaMethod.ReturnsVoid && !_model.LambdaMethod.ReturnsVoidTask)
             {
 #>
                 Body = <#= _model.LambdaMethod.ReturnType.IsString() ? "response" : "body" #>,
@@ -467,7 +493,7 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
             <#= _model.LambdaMethod.ContainingType.Name.ToCamelCase() #>.<#= _model.LambdaMethod.Name #>(<#= parameters #>);
 <#
         }
-        else if (_model.LambdaMethod.ReturnsTask)
+        else if (_model.LambdaMethod.ReturnsVoidTask)
         {
 #>
             await <#= _model.LambdaMethod.ContainingType.Name.ToCamelCase() #>.<#= _model.LambdaMethod.Name #>(<#= parameters #>);
@@ -476,7 +502,7 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
         else
         {
 #>
-            return <#= _model.LambdaMethod.IsAsync ? "await " : "" #><#= _model.LambdaMethod.ContainingType.Name.ToCamelCase() #>.<#= _model.LambdaMethod.Name #>(<#= parameters #>);
+            return <#= _model.LambdaMethod.ReturnsGenericTask ? "await " : "" #><#= _model.LambdaMethod.ContainingType.Name.ToCamelCase() #>.<#= _model.LambdaMethod.Name #>(<#= parameters #>);
 <#
         }
     }
--- a/Libraries/src/Amazon.Lambda.Annotations.nuspec
+++ b/Libraries/src/Amazon.Lambda.Annotations.nuspec
@@ -21,8 +21,8 @@
         <file src="Amazon.Lambda.Annotations\THIRD_PARTY_LICENSES" target="THIRD_PARTY_LICENSES" />
 
         <!-- Dependencies to make sure attributes are available in consuming csproj, this ensures packaged version of customer code have all the dependencies needed. -->
-        <file src="Amazon.Lambda.Annotations\bin\Release\netstandard2.0\Amazon.Lambda.Annotations.dll" target="lib/net6.0" />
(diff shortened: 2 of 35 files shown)


Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
```diff
@@ -0,0 +1,437 @@
...
+    /// </remarks>
+    /// <example>
+    /// return HttpResults.Ok("All Good")
+    ///                   .AddHeader("Custom-Header", "FooBar");
+    /// 
+    /// </example>
+    public class HttpResults : IHttpResults
+    {
+        private HttpStatusCode _statusCode;
+        private string _body;
```
normj, line 116: These 2 warnings are misleading because the variables are not used in the stub .NET Standard 2.0 version but are used in the real .NET 6 version.

Libraries/src/Amazon.Lambda.Annotations.nuspec
```diff
@@ -21,8 +21,8 @@
         <file src="Amazon.Lambda.Annotations\THIRD_PARTY_LICENSES" target="THIRD_PARTY_LICENSES" />
 
         <!-- Dependencies to make sure attributes are available in consuming csproj, this ensures packaged version of customer code have all the dependencies needed. -->
-        <file src="Amazon.Lambda.Annotations\bin\Release\netstandard2.0\Amazon.Lambda.Annotations.dll" target="lib/net6.0" />
-        <file src="Amazon.Lambda.Annotations\bin\Release\netstandard2.0\Amazon.Lambda.Annotations.xml" target="lib/net6.0" />
+        <file src="Amazon.Lambda.Annotations\bin\Release\net6.0\Amazon.Lambda.Annotations.dll" target="lib/net6.0" />
```
normj, line 24: This was changed to now use the .NET 6 version that was added for this PR.

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
```diff
@@ -0,0 +1,437 @@
...
+
+            return result;
+        }
+
+        /// <summary>
+        /// Creates an IHttpResult for a Forbidden (403) status code.
+        /// </summary>
+        /// <param name="body">Optional response body</param>
+        /// <returns></returns>
+        public static IHttpResults Forbid(object body = null)
```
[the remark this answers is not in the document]
normj: I was surprised by that as well but I followed the same pattern ASP.NET Core does with their version of this class. https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.typedresults.forbid

Libraries/src/Amazon.Lambda.Annotations.nuspec
```diff
@@ -21,8 +21,8 @@
...
+        <file src="Amazon.Lambda.Annotations\bin\Release\net6.0\Amazon.Lambda.Annotations.xml" target="lib/net6.0" />
 
         <!-- Include every dependency manually for analyzer, whenever a new dependency is added, it has to be added here. -->
         <!-- NOTE: Project dependencies should come from their own bin folder to make sure a code signed binary is packed -->
```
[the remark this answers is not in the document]
normj: Not sure understand your question but hopefully below describes what we are trying to do.

The files elements with the `target` attribute set to `lib/net6.0` are for the runtime dependency of the application which we are targeting .NET 6 for applications.

The file elements with the target to `analyzers\dotnet\cs` are used by the source generator and used at compile time inside MSbuild/VS so it must target .NET Standard 2.0 to support .NET Framework.

boblodgett (approved):

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
```diff
@@ -0,0 +1,437 @@
...
+
+                JsonSerializer.Serialize(stream, response);
+            }
+            else
+            {
+                var response = new APIGatewayV2Response
+                {
+                    StatusCode = (int)_statusCode,
+                    Body = _body,
+                    Headers = ConvertToV2MultiValueHeaders(_headers),
```
ashovlin, line 364: I'm not sure exactly where in the code it's happening, but when I try:
```
return HttpResults.BadRequest("Cannot divide by zero")
                    .AddHeader("key", "value1")
                    .AddHeader("key", "value2");
```
it results in `key: value1, value2` as expected.

But this:
```
return HttpResults.BadRequest("Cannot divide by zero")
                    .AddHeader("key", "value1")
                    .AddHeader("key", "value2")
                    .AddHeader("KEY", "VALUE3");
```
results in `key: VALUE3`. I was expecting `key: value1, value2, VALUE3` since the header keys should be case-insensitive, but let me know if my understanding is incorrect. I think separate `key` and `KEY` headers might be okay since multiple values could be comma-separated or on multiple lines (would have to double-check spec), but wasn't expecting the first two values to be wiped out.

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
```diff
@@ -0,0 +1,437 @@
...
+    /// configured event source's protocol format and version.
+    /// 
+    /// Users should use the implementation class HttpResults to construct an instance of IHttpResults with the configured, status code, response body and headers.
+    /// </summary>
+    /// <example>
+    /// return HttpResults.Ok("All Good")
+    ///                   .AddHeader("Custom-Header", "FooBar");
+    /// 
+    /// </example>
+    public interface IHttpResults
```
ashovlin, line 81: Not a big deal/okay if others disagree, but I don't love this as plural since when I look at the Lambda method signature I read it as _returning multiple http results_. 

Plus I think singular would more closely align with ASP.NET's `IResult`.

But I do like `HttpResults` as plural, because I think of `HttpResults.Ok` as I'm choosing `Ok` out of all of the possible results.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
```diff
@@ -476,7 +486,7 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
         else
         {
 #>
-            return <#= _model.LambdaMethod.IsAsync ? "await " : "" #><#= _model.LambdaMethod.ContainingType.Name.ToCamelCase() #>.<#= _model.LambdaMethod.Name #>(<#= parameters #>);
+            return <#= _model.LambdaMethod.ReturnsGenericTask ? "await " : "" #><#= _model.LambdaMethod.ContainingType.Name.ToCamelCase() #>.<#= _model.LambdaMethod.Name #>(<#= parameters #>);
```
ashovlin, line 489: I think this is not supported, since you have to use `IHttpResults` with API Gateway (right?)
```       
        [LambdaFunction()]
        public IHttpResults GetOrderInfo()
        {
            return HttpResults.Ok(new OrderInfo { OrderId = "132", CustomerId = "999" });
        }
```

But it results in this compile error, which I think may be confusing because the `Stream` is hidden from the user's code. Can we give a better error for if `HttpResults` are returned outside of API Gateway?
```
Cannot implicitly convert type 'Amazon.Lambda.Annotations.APIGateway.IHttpResults' to 'System.IO.Stream'. An explicit conversion exists (are you missing a cast?)
```

ashovlin (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
```diff
@@ -353,7 +353,7 @@ namespace <#= _model.LambdaMethod.ContainingNamespace #>
             // return 400 Bad Request if there exists a validation error
             if (validationErrors.Any())
             {
-                return new <#= _model.LambdaMethod.IsAsync ? _model.GeneratedMethod.ReturnType.TaskTypeArgument : _model.GeneratedMethod.ReturnType.FullName #>
+                return new <#= _model.LambdaMethod.ReturnsVoidOrGenericTask ? _model.GeneratedMethod.ReturnType.TaskTypeArgument : _model.GeneratedMethod.ReturnType.FullName #>
```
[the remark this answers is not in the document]
normj: Great catch. Updated the generator and tests to cover validations.

Libraries/src/Amazon.Lambda.Annotations/Amazon.Lambda.Annotations.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <AssemblyVersion>0.10.0.0</AssemblyVersion>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
```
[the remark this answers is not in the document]
normj: Because the new `HttpResults` class uses `System.Text.Json` for JSON serialization that is not available for .NET Standard 2.0 without taking a dependency. Since `System.Text.Json` is only used at runtime, not source generator time, we can avoid the dependency by just using the newer APIs for .NET 6 during runtime.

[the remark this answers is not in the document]
ashovlin: Got it, thanks.

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
```diff
@@ -0,0 +1,437 @@
...
+        /// <returns>The same isntance to allow fluent call pattern.</returns>
+        IHttpResults AddHeader(string name, string value);
+    }
+
+    /// <summary>
+    /// Implemenation class for IHttpResults. Consumers should use one of the static methods to create the a result with the desired status code.
+    /// </summary>
+    /// <remarks>
+    /// If a response body is provided it is format using the following rules:
+    ///     * For string then returned as is.
```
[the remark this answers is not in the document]
ashovlin: And typo in "implementation"

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
```diff
@@ -0,0 +1,437 @@
...
+        /// <summary>
+        /// Creates an IHttpResult for redirect responses.
+        /// </summary>
+        /// <remarks>
+        /// This method uses the same logic for determing the the Http status code as the Microsoft.AspNetCore.Http.TypedResults.Redirect uses.
+        /// https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.typedresults.redirect
+        /// </remarks>
+        /// <param name="uri">The URI to redirect to. The value will be set in the location header.</param>
+        /// <param name="permanent">Whether the redirect should be a permanet (301) or temporary (302) redirect.</param>
+        /// <param name="preserveMethod">Whether the request method should be preserverd. If set to true use 308 for permanent or 307 for temporary redirects.</param>
```
ashovlin, line 231: nit: typo in "preserved"

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
normj: I agree the interface name doesn't make sense to be plural. I renamed it to be `IHttpResult`.

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
normj: Fixed

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
normj: Updated the `AddHeader` method to do a `ToLower` on the name and added your example as a test case.

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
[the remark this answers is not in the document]
normj: Added bullet list and fixed typo.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
normj: I added a new Diagnostic compiler error message that will give a more meaningful error message when the `IHttpResult` is attempted to be used as a return type for a non API Lambda function.

I also updated the generator to not run the CloudFormation sync if our compiler error diagnostics are emitted. No sense mutating the CF template if we now we have a bad build.

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
```diff
@@ -0,0 +1,437 @@
...
+        /// <param name="options"></param>
+        /// <returns></returns>
+        Stream Serialize(HttpResultSerializationOptions options);
+
+        /// <summary>
+        /// Add header to the IHttpResult. The AddHeader method can be called multiple times for the same header to add multi values for a header.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="value"></param>
+        /// <returns>The same isntance to allow fluent call pattern.</returns>
```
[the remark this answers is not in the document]
normj: Fixed

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
```diff
@@ -0,0 +1,448 @@
...
+    /// If a response body is provided it is format using the following rules:
+    /// <list type="bullet">
+    ///     <item>
+    ///         <description>For string then returned as is.</description>
+    ///     </item>
+    ///     <item>
+    ///         <description>For Stream, byte[] or IList&lt;byte&gt; the data is consided binary and base 64 encoded.</description>
+    ///     </item>
+    ///     <item>
+    ///         <description>Anything other type is serialized to JSON.</description>
```
ashovlin, line 113: I don't have a strong opinion, so this isn't a blocker, but I'm wondering if it'd be helpful if we set `Content-Type:  "application/json"`automatically in this case?

Admittedly when testing in Firefox I was initially wondering _"why doesn't this JSON look normal?"_, which was because it wasn't triggering the JSON viewer.

I haven't tested ASP.NET equivalent to see how it behaves.

ashovlin (approved):
Added one new suggestion/comment, but I don't have a strong opinion on it so approving.

I think we should update the annotations-specific README with the new feature, but I'm fine if we do that in a separate PR ahead of releasing this.

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
normj: With some experimenting with ASP.NET Core I added an autofill of content type with the following rules.

* No body - content type not set
* String passed in for body - content type set as `text/plain`
* Stream, Byte[] or IList<byte> - content type set as `application/octet-stream`
* Anything else - content type set to `application/json`

`HttpResults` unit tests updated to reflect this change.

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
```diff
@@ -0,0 +1,465 @@
...
+        /// </summary>
+        /// <param name="options"></param>
+        /// <returns></returns>
+        Stream Serialize(HttpResultSerializationOptions options);
+
+        /// <summary>
+        /// Add header to the IHttpResult. The AddHeader method can be called multiple times for the same header to add multi values for a header.
+        /// HTTP header names are case insensitive and the AddHeader method will normalize header name casing by calling ToLower on them.
+        /// </summary>
+        /// <param name="name">HTTP heade rname</param>
```
ashovlin, line 94: nit: Typo in "header"

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
```diff
@@ -0,0 +1,465 @@
...
+    /// return HttpResults.Ok("All Good")
+    ///                   .AddHeader("Custom-Header", "FooBar");
+    /// 
+    /// </example>
+    public class HttpResults : IHttpResult
+    {
+        private const string HEADER_NAME_CONTENT_TYPE = "content-type";
+        private const string CONTENT_TYPE_APPLICATION_JSON = "application/json";
+        private const string CONTENT_TYPE_TEXT_PLAIN = "text/plain";
+        private const string CONTENT_TYPE_APPLICATION_OCTECT_STREAM = "application/octet-stream";
```
ashovlin, line 127: nit: Typo in "OCTET"

ashovlin (changes requested):

Libraries/src/Amazon.Lambda.Annotations/APIGateway/HttpResults.cs
ashovlin: Nice, thanks!

ashovlin (approved):

PR #1440, opened 2023-02-25
fix: Report diagnostic error for invalid CloudFormation resource names

96malhar:
*Issue #, if available:*
DOTNET-6738

*Description of changes:*
* Renames the `Name` property of `ILambdaFunctionSerializable` and `LambdaFunctionAttribute` to `ResourceName` so that it is better aligned with its use-case.
* Report diagnostic errrors when customers specify an invalid `ResourceName` as part of the `LambdaFunctionAttribute`. 

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Diagnostics/DiagnosticDescriptors.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Diagnostics/DiagnosticDescriptors.cs
@@ -37,7 +37,7 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Diagnostics
             category: "AWSLambdaCSharpGenerator",
             DiagnosticSeverity.Info,
             isEnabledByDefault: true);
-        
+
         public static readonly DiagnosticDescriptor MissingDependencies = new DiagnosticDescriptor(id: "AWSLambda0104",
             title: "Missing reference to a required dependency",
             messageFormat: "Your project has a missing required package dependency. Please add a reference to the following package: {0}",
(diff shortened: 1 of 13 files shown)


Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -51,5 +51,12 @@ public static class DiagnosticDescriptors
             category: "AWSLambdaCSharpGenerator",
             DiagnosticSeverity.Error,
             isEnabledByDefault: true);
+
+        public static readonly  DiagnosticDescriptor InvalidResourceName = new DiagnosticDescriptor(id: "AWSLambda0106",
```
normj, line 55: You need to add this new diagnostic doc here https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Diagnostics/AnalyzerReleases.Shipped.md. I would just add it to the versions table as version 0.13.0

normj (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Diagnostics/DiagnosticDescriptors.cs
96malhar: Done

normj (approved):

ashovlin (approved):
I think it'd be nice to exercise the new validation via a test, but not a blocker.

Issue #1441, opened 2023-02-27
AWS Lambda .NET 7 AOT (Native) does not start when provisioned concurrency is used


vitalybibikov:
### Describe the bug

When .NET 7 AOT lamdba is being deployed to al2 runtime (Custom runtime on Amazon Linux 2)
it's impossible for it to create Provisioned concurrency as it errors with:

Custom runtime on Amazon Linux 2

'Failed FUNCTION_ERROR_INIT_FAILURE(0/10)'

In logs I can see:

![image](https://user-images.githubusercontent.com/7008739/221578445-66569ba3-95ef-4fcc-a961-cfa0176c0287.png)

### Expected Behavior

Provisioned concurrency works as expected, function initializes and starts.

### Current Behavior

![image](https://user-images.githubusercontent.com/7008739/221577410-74ae578a-9c4d-42ac-ab8d-e02097d7df3e.png)
![image](https://user-images.githubusercontent.com/7008739/221579804-f179ad75-cdea-4bea-a503-fa338fdbcd76.png)

### Reproduction Steps

We are having this issue in production for our functions ,but it is reproducable for a simple lambda.NativeAOT template:

1) Create new lambda project => dotnet new lambda.NativeAOT
2) Deploy it => dotnet lambda deploy-function
3) Create version, e.g. 1
4) Go to AWS Console => Lamdba => Configuration => Concurrency
5) Create Provisioned concurrency configuration
6) Wait

### Additional Information/Context

Project configuration:

```
<PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net7.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <AWSProjectType>Lambda</AWSProjectType>
    <AssemblyName>bootstrap</AssemblyName>
    <!-- This property makes the build directory similar to a publish directory and helps the AWS .NET Lambda Mock Test Tool find project dependencies. -->
    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
    <!-- Generate Native AOT image during publishing to improve cold start time. -->
    <PublishAot>true</PublishAot>
    <!-- StripSymbols tells the compiler to strip debugging symbols from the final executable if we're on Linux and put them into their own file. 
    This will greatly reduce the final executable's size.-->
    <StripSymbols>true</StripSymbols>
  </PropertyGroup>
```

### AWS .NET SDK and/or Package version used

[...]

PR #1444, opened 2023-02-28
Fix issue using .NET AOT and provisioned concurrency.

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/1441

*Description of changes:*
The .NET Lambda runtime client has code in it to detect if startup is part of a provision concurrency. If so then try to load up as many assemblies as possible to get jitting .NET as part of the provisioned concurrency startup. This logic is not valid in an AOT environment where the is just a single executable and so throws an exception.

The fix is to check if running in AOT mode and if so skip the prejitting logic.

I have tested this by first creating a AOT provisioned function and reproduced the failure. Then created a special build with this change deployed again as an AOT provisioned function and start ran correctly.

### DIFF
--- a/Libraries/src/Amazon.Lambda.RuntimeSupport/Amazon.Lambda.RuntimeSupport.csproj
+++ b/Libraries/src/Amazon.Lambda.RuntimeSupport/Amazon.Lambda.RuntimeSupport.csproj
@@ -4,9 +4,8 @@
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net5.0;net6.0</TargetFrameworks>
-    <VersionPrefix>1.8.3</VersionPrefix>
+    <VersionPrefix>1.8.4</VersionPrefix>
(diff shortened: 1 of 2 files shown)


Beau-Gosse-dev:
Do we know if the call to GetReferencedAssemblies is causing `warning IL2104: Assembly 'Amazon.Lambda.RuntimeSupport' produced trim warnings.`

It would be nice to fix this in a way that we can remove the warning if possible.

normj:
> Do we know if the call to GetReferencedAssemblies is causing `warning IL2104: Assembly 'Amazon.Lambda.RuntimeSupport' produced trim warnings.`
> 
> It would be nice to fix this in a way that we can remove the warning if possible.

I would like to include fixing that warning when we do a pass across the whole library for AOT warnings. For now I just want to address this specific issue.

Beau-Gosse-dev (approved):

96malhar (approved):

muhammad-othman (approved):

PR #1445, opened 2023-02-28
Add transcribe eventbridge

bjhogan:
*Description of changes:*

Adds Transcribe Job State Change Event. 


Libraries/src/Amazon.Lambda.CloudWatchEvents/TranscribeEvents/TranscribeJobStateChange.cs
```diff
@@ -0,0 +1,23 @@
+namespace Amazon.Lambda.CloudWatchEvents.TranscribeEvents
+{
+    /// <summary>
+    /// This class represents the details of a Transcribe Job State Change sent via EventBridge.
```
normj, line 4: Is the a URL you can add to where this event type is documented?

normj (changes requested):
We need a unit test for this new POCO. You can look at the [CloudWatchEventsS3ObjectCreate](https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/test/EventsTests.Shared/EventTests.cs#L3023) test as an example.

Libraries/src/Amazon.Lambda.CloudWatchEvents/TranscribeEvents/TranscribeJobStateChange.cs
bjhogan: https://docs.aws.amazon.com/transcribe/latest/dg/monitoring-events.html

normj: Sorry I wasn't clear, can you add the URL as part of the NDoc here?

normj (changes requested):

normj (approved):

normj:
PR has been released as part of version 4.2.0

PR #1463, opened 2023-03-23
Marshall IHttpRequestFeature.RawTarget

kevinstapleton:
*Description of changes:*
The `HttpRequest.Path` and `HttpRequest.PathBase` properties contain unescaped representations of the request path; however, it is sometimes useful to have access to the raw (escaped) request path as supplied exactly by the caller. Due to inconsistencies in URL escaping rules (that is, which characters are considered necessary to escape) and to the fact that the `PathString` type contains a _partial_ unescaping, it is not possible to reconstruct the original, raw request path from the `Path` and `PathBase` properties.

The `RawTarget` property exists on the `IHttpRequestFeature` interface to address the above issue; however, that property is currently not set during request marshalling. This PR updates the `MarshallRequest` implementations to set the `IHttpRequestFeature.RawTarget` property.

All existing unit tests pass and new unit tests have been added to verify new functionality.


normj (approved):
Nice! Thanks for the detail PR.

normj:
Version 8.1.0 of Amazon.Lambda.AspNetCoreServer has been released with this PR.

PR #1464, opened 2023-03-23
Fix bug that ensures the stream representing stdout and stderr can not be disposed.

normj:
*Description of changes:*
We found some libraries that wrap stdout and stderr with their own `TextWriter` can trigger a dispose call on stdout. For the normal console out `TextWriter` this dispose is a noop. Amazon.Lambda.RuntimeSupport redirects stdout to its own `TextWriter` to redirector the logs to Lambda. If the dispose was called out stdout that will trickle down the `TextWriter` used by Amazon.Lambda.RuntimeSupport and then future log messages will fail with an object disposed exception.

This PR creates a subclass of `StreamWriter` and makes the close and dispose methods a noop so unintended dispose calls will not break logging.

This behavior only happens in the .NET 6 managed runtime where the Lambda telemetry API is used for reporting logs. I switched our tests for Amazon.Lambda.RuntimeSupport to use the .NET 6 managed runtime instead of `provided.al2` to reproduce the bug and confirm the fix.

### DIFF
--- a/Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.IntegrationTests/Amazon.Lambda.RuntimeSupport.IntegrationTests.csproj
+++ b/Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.IntegrationTests/Amazon.Lambda.RuntimeSupport.IntegrationTests.csproj
@@ -20,19 +20,19 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="AWSSDK.IdentityManagement" Version="3.7.0" />
-    <PackageReference Include="AWSSDK.Lambda" Version="3.7.4.9" />
-    <PackageReference Include="AWSSDK.S3" Version="3.7.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.1.1" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1">
+    <PackageReference Include="AWSSDK.IdentityManagement" Version="3.7.100.98" />
+    <PackageReference Include="AWSSDK.Lambda" Version="3.7.105.17" />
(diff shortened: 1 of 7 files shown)


Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.IntegrationTests/Amazon.Lambda.RuntimeSupport.IntegrationTests.csproj
```diff
@@ -20,19 +20,19 @@
...
-    <PackageReference Include="AWSSDK.IdentityManagement" Version="3.7.0" />
-    <PackageReference Include="AWSSDK.Lambda" Version="3.7.4.9" />
-    <PackageReference Include="AWSSDK.S3" Version="3.7.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.1.1" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1">
+    <PackageReference Include="AWSSDK.IdentityManagement" Version="3.7.100.98" />
+    <PackageReference Include="AWSSDK.Lambda" Version="3.7.105.17" />
+    <PackageReference Include="AWSSDK.S3" Version="3.7.103.34" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.6.0-preview-20230223-05" />
```
ashovlin, line 26: Why the preview version?

philasmar (approved):

ashovlin (approved):

Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.IntegrationTests/Amazon.Lambda.RuntimeSupport.IntegrationTests.csproj
normj: Reverted back to last stable version

PR #1466, opened 2023-03-26
Added FromCustomAuthorizerAttribute and tests.

kabaluk:
*Description of changes:*
Added FromCustomAuthorizerAttribute and tests.
This will allow to retrieve values from CustomAuthorizer context without having to receive the full APIGatewayProxyRequest or APIGatewayHttpApiV2ProxyRequest..

Both REST and HttpApi supported

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/AttributeModelBuilder.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/AttributeModelBuilder.cs
@@ -53,6 +53,15 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Models.Attributes
                     Type = TypeModelBuilder.Build(att.AttributeClass, context)
                 };
             }
+            else if(att.AttributeClass.Equals(context.Compilation.GetTypeByMetadataName(TypeFullNames.FromCustomAuthorizerAttribute), SymbolEqualityComparer.Default))
+            {
+                var data = FromCustomAuthorizerAttributeBuilder.Build(att);
+                model = new AttributeModel<FromCustomAuthorizerAttribute>
+                {
+                    Data = data,
+                    Type = TypeModelBuilder.Build(att.AttributeClass, context)
+                };
+            }
             else if (att.AttributeClass.Equals(context.Compilation.GetTypeByMetadataName(TypeFullNames.HttpApiAttribute), SymbolEqualityComparer.Default))
             {
                 var data = HttpApiAttributeBuilder.Build(att);
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/FromCustomAuthorizerAttributeBuilder.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/FromCustomAuthorizerAttributeBuilder.cs
@@ -1,0 +1,22 @@
+﻿using Amazon.Lambda.Annotations.APIGateway;
+using Microsoft.CodeAnalysis;
+
+namespace Amazon.Lambda.Annotations.SourceGenerator.Models.Attributes
+{
+    public class FromCustomAuthorizerAttributeBuilder
+    {
+        public static FromCustomAuthorizerAttribute Build(AttributeData att)
+        {
+            var data = new FromCustomAuthorizerAttribute();
+            foreach (var pair in att.NamedArguments)
+            {
+                if (pair.Key == nameof(data.Name) && pair.Value.Value is string value)
+                {
+                    data.Name = value;
+                }
+            }
+
+            return data;
+        }
+    }
+}
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.cs
@@ -780,7 +780,7 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Templates
                 // string parameter does not need to be de-serialized
                 if (parameter.Type.IsString())
                 {
- 
+
             
             #line default
             #line hidden
@@ -798,7 +798,7 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Templates
                 }
                 else
                 {
- 
+
             
             #line default
             #line hidden
@@ -927,7 +927,64 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Templates
                     "\n\r\n");
             
             #line 342 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\LambdaFunctionTemplate.tt"
+           
+            }
+            else if (parameter.Attributes.Any(att => att.Type.FullName == TypeFullNames.FromCustomAuthorizerAttribute))
+            {
+                var fromAuthorizerAttribute = parameter.Attributes?.FirstOrDefault(att => att.Type.FullName == TypeFullNames.FromCustomAuthorizerAttribute) as AttributeModel<Amazon.Lambda.Annotations.APIGateway.FromCustomAuthorizerAttribute>;
+                
+                var authKey = fromAuthorizerAttribute?.Data?.Name ?? parameter.Name;
+                if(restApiAttribute != null)
+                {
+                
 
+            
+            #line default
+            #line hidden
+            this.Write("            var ");
+            
+            #line 353 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\LambdaFunctionTemplate.tt"
+            this.Write(this.ToStringHelper.ToStringWithCulture(parameter.Name));
+            
+            #line default
+            #line hidden
+            this.Write(" = __request__.RequestContext.Authorizer[\"");
+            
+            #line 353 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\LambdaFunctionTemplate.tt"
+            this.Write(this.ToStringHelper.ToStringWithCulture(authKey));
+            
+            #line default
+            #line hidden
+            this.Write("\"].ToString();\r\n");
+            
+            #line 354 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\LambdaFunctionTemplate.tt"
+
+               }
+               else
+               {
+
+            
+            #line default
+            #line hidden
+            this.Write("            var ");
+            
+            #line 359 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\LambdaFunctionTemplate.tt"
+            this.Write(this.ToStringHelper.ToStringWithCulture(parameter.Name));
+            
+            #line default
+            #line hidden
+            this.Write(" = __request__.RequestContext.Authorizer.Lambda[\"");
+            
+            #line 359 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\LambdaFunctionTemplate.tt"
+            this.Write(this.ToStringHelper.ToStringWithCulture(authKey));
+            
+            #line default
+            #line hidden
+            this.Write("\"].ToString();\r\n");
+            
+            #line 360 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\LambdaFunctionTemplate.tt"
+
+               }
             }
             else
             {
@@ -945,7 +1002,7 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Templates
                     "     if (validationErrors.Any())\r\n            {\r\n                var errorResult" +
                     " = new ");
             
-            #line 356 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\LambdaFunctionTemplate.tt"
+            #line 375 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\LambdaFunctionTemplate.tt"
             this.Write(this.ToStringHelper.ToStringWithCulture(restApiAttribute != null || httpApiAttribute?.Data.Version == Amazon.Lambda.Annotations.APIGateway.HttpApiVersion.V1 ? "Amazon.Lambda.APIGatewayEvents.APIGatewayProxyResponse" : "Amazon.Lambda.APIGatewayEvents.APIGatewayHttpApiV2ProxyResponse"));
             
             #line default
@@ -962,7 +1019,7 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Templates
                 };
 ");
             
-            #line 366 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\LambdaFunctionTemplate.tt"
+            #line 385 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\LambdaFunctionTemplate.tt"
 
                 if(_model.LambdaMethod.ReturnsIHttpResults)
                 {
@@ -974,7 +1031,7 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Templates
                     "System.Text.Json.JsonSerializer.Serialize(errorStream, errorResult);\r\n          " +
                     "      return errorStream;\r\n");
             
-            #line 373 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\LambdaFunctionTemplate.tt"
+            #line 392 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\LambdaFunctionTemplate.tt"
 
                 }
                 else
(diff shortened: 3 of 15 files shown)


normj (changes requested):
Cool PR!

Can you fix the whitespaces changes to the `LambdaFunctionTemplate.tt` file? The whole file is looks different.

Who should we handle the other type of authorizer properties. For example on the `APIGatewayHttpApiV2ProxyRequest` besides the `Lambda` property there is also `Jwt` and `IAM` properties. Should we make the string accessing the authorizer field have the prefix of the property.

kabaluk:
don't seem to be able to fix the whitespace issue to save my life :(
Please review ignoring whitespace ?? Maybe?

kabaluk:
Would be really nice to get this avail;able. I will gladly fix the conflicts if this can be looked at again.

kabaluk:
I can see you did a major refactoring in the dev branch will push again once released.

If i might be so bold, I looked at what you did. 
why not take it one step further and separate each attribute into it's own file?
Talking about the APIGatewaySetupParameters.tt
Breaking that file down, using the technique you used before, would make it much easier to add new attributes in the future 

Just a thought.

normj:
@kabaluk I suspect adding new attributes would be done in separate files. For example if we added SQS or S3 I would put them in separate TT. One of my goals with the refactor was to make it easier to add new attributes as the previous monolithic tt file was really hard to reconcile with. `HttpApi` and `RestApi` attributes are 9X percent the same code so I didn't really want to make 2 separate tt files for each one of those. Plus I thought the parameter setup was getting pretty long so I was looking for some separation which is why I divided the parameter setup and the invoke. I suspect in SQS or S3 use cases the parameter setup wouldn't be so complicated.

Another goal I had was I wanted to generate the exact same code including whitespaces to make the PR review easier and then later I could do more specific refactoring. Otherwise the PR review would have been every line is different and that is hard to review. But going forward I could see putting each of the `FromXXX` in a separate tt file as well.

paulo-stadion:
> @kabaluk I suspect adding new attributes would be done in separate files. For example if we added SQS or S3 I would put them in separate TT. One of my goals with the refactor was to make it easier to add new attributes as the previous monolithic tt file was really hard to reconcile with. `HttpApi` and `RestApi` attributes are 9X percent the same code so I didn't really want to make 2 separate tt files for each one of those. Plus I thought the parameter setup was getting pretty long so I was looking for some separation which is why I divided the parameter setup and the invoke. I suspect in SQS or S3 use cases the parameter setup wouldn't be so complicated.
> 
> Another goal I had was I wanted to generate the exact same code including whitespaces to make the PR review easier and then later I could do more specific refactoring. Otherwise the PR review would have been every line is different and that is hard to review. But going forward I could see putting each of the `FromXXX` in a separate tt file as well.

I was talking specifically of the FromXXX attributes. Apologies, I should have been more explicit. Sounds like a great change and it would make it a lot easier to add and test more FromXXX attributes. Looking forward to have that in main so i can re add the FromCustomAuthorizer attribute.
Even started looking on how to do the FromJwt. 😄

kabaluk:
Readded CustomAuthorizerAttribute..

My PC really doesn't like LambdaFunctionTemplate.cs. 
Apologies.

only whitespace changes.
no probles with the new files

kabaluk:
Any possibility of having this looked at again, please?

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.tt
```diff
@@ -244,6 +244,57 @@
...
+
+                // Use parameter name as key, if Name has not specified explicitly in the attribute definition.
+                var authKey = fromAuthorizerAttribute?.Data?.Name ?? parameter.Name;
+                if(restApiAttribute != null)
+                {
+#>
+            var <#= parameter.Name #> = default(<#= parameter.Type.FullName #>);
+            if (__request__.RequestContext?.Authorizer == null)
+            {
+                validationErrors.Add("Could not find Authorizer data for request");
```
normj, line 262: I'm wondering if there is missing data from the authorizer if we should be returning back `401` unauthorized instead of a bad parameter status code.

normj (changes requested):
I did an initial scan and it looks good. I had one comment so far. I need to test the experience end to end next. I'll try and find time to do that soon.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.tt
paulo-stadion: Good point. Away from pc for 2 weeks.
Feel free to change or I will once I am  back at desk.

kabaluk: Good point. It should return 401.
I am away from pc for 2 weeks, feel free to change or i will once back.

philasmar: Would you be able to have a look at Norm's comments?

ashishdhingra:
@kabaluk Please review the last review comment from @normj and see if you are able to address it.

kabaluk:
> @kabaluk Please review the last review comment from @normj and see if you are able to address it.

Please have a look.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.tt
```diff
@@ -247,6 +247,87 @@
...
+                        {"x-amzn-ErrorType", "AccessDeniedException"}
+                    },
+                    StatusCode = 401
+                };
+            }
+            
+            try
+            {
+                var __authValue_<#= parameter.Name #>__ = __request__.RequestContext.Authorizer["<#= authKey #>"];
+                <#= parameter.Name #> = (<#= parameter.Type.FullName #>)Convert.ChangeType(__authValue_<#= parameter.Name #>__?.ToString(), typeof(<#= parameter.Type.FullNameWithoutAnnotations #>));
```
GarrettBeatty, line 303: ## Why We Need `.ToString()` for Custom Authorizer Values

### The Data Flow

1. __Authorizer Lambda returns context__: When your custom authorizer runs, it returns a context dictionary:

   ```csharp
   Context = new Dictionary<string, object>
   {
       { "userId", "12345" },
       { "permissions", "admin" }
   }
   ```

2. __API Gateway passes this to the protected Lambda__ in the request payload as JSON:

   ```json
   {
     "requestContext": {
       "authorizer": {
         "userId": "12345",
         "permissions": "admin"
       }
     }
   }
   ```

3. __Lambda deserializes the request__ using the configured serializer (System.Text.Json or Newtonsoft.Json).

### The Problem: `Dictionary<string, object>` Deserialization

The `Authorizer` property is typed as `Dictionary<string, object>`:

```csharp
public class APIGatewayCustomAuthorizerContext : Dictionary<string, object>
```

When the serializer encounters a JSON value like `"12345"` and needs to deserialize it into `object`, it doesn't know what concrete type to use. So:

- __System.Text.Json__ wraps it in a `JsonElement` struct
- __Newtonsoft.Json__ wraps it in a `JToken` (like `JValue`)


so rather than having statements like 

```

if (__authValue__ is System.Text.Json.JsonElement jsonElement)
{
    userId = Convert.ChangeType(jsonElement.ToString(), typeof(string));
}
```

i just call toString which both of these serializers have.

not sure if there is a better way

GarrettBeatty: @normj not sure if you have any better ideas here of it ToString is good enough

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.tt
```diff
@@ -247,6 +247,87 @@
...
+            }
+
+<#
+                }
+                else
+                {
+#>
+            var <#= parameter.Name #> = default(<#= parameter.Type.FullName #>);
+            if (__request__.RequestContext?.Authorizer?.Lambda == null || __request__.RequestContext?.Authorizer?.Lambda.ContainsKey("<#= authKey #>") == false)
+            {
```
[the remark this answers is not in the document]
GarrettBeatty: this should be fixed now

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.tt
```diff
@@ -247,6 +247,89 @@
...
+                return new <#= restApiAttribute != null || httpApiAttribute?.Data.Version == Amazon.Lambda.Annotations.APIGateway.HttpApiVersion.V1 ? "Amazon.Lambda.APIGatewayEvents.APIGatewayProxyResponse" : "Amazon.Lambda.APIGatewayEvents.APIGatewayHttpApiV2ProxyResponse" #>
+                {                    
+                    Headers = new Dictionary<string, string>
+                    {
+                        {"Content-Type", "application/json"},
+                        {"x-amzn-ErrorType", "AccessDeniedException"}
+                    },
+                    StatusCode = 401
+                };
```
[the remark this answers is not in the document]
GarrettBeatty: fixed

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.tt
```diff
@@ -247,6 +247,89 @@
...
```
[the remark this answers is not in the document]
GarrettBeatty: fixed

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.tt
```diff
@@ -247,6 +247,157 @@
...
 
+<#
+                }
+            }
+            else if (parameter.Attributes.Any(att => att.Type.FullName == TypeFullNames.FromCustomAuthorizerAttribute))
+            {
+                var fromAuthorizerAttribute = parameter.Attributes?.FirstOrDefault(att => att.Type.FullName == TypeFullNames.FromCustomAuthorizerAttribute) as AttributeModel<Amazon.Lambda.Annotations.APIGateway.FromCustomAuthorizerAttribute>;
+
+                // Use parameter name as key, if Name has not specified explicitly in the attribute definition.
+                var authKey = fromAuthorizerAttribute?.Data?.Name ?? parameter.Name;
```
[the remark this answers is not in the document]
GarrettBeatty: fixed

GarrettBeatty:
https://github.com/aws/aws-lambda-dotnet/actions/runs/21604179496/job/62257446270

Libraries/test/IntegrationTests.Helpers/CloudFormationHelper.cs
```diff
@@ -0,0 +1,75 @@
+using System.Linq;
+using System.Threading.Tasks;
+using Amazon.CloudFormation;
+using Amazon.CloudFormation.Model;
+
+namespace IntegrationTests.Helpers
```
GarrettBeatty, line 6: this is all existing code just moved into a common project to share between the two integration test projects

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.tt
```diff
@@ -247,6 +247,157 @@
...
+                // REST API and HTTP API v1 both use APIGatewayProxyRequest where RequestContext.Authorizer is a dictionary.
+                // Only HTTP API v2 uses APIGatewayHttpApiV2ProxyRequest with RequestContext.Authorizer.Lambda as the dictionary.
+                if(restApiAttribute != null || httpApiAttribute?.Data.Version == Amazon.Lambda.Annotations.APIGateway.HttpApiVersion.V1)
+                {
+#>
+            var <#= parameter.Name #> = default(<#= parameter.Type.FullName #>);
+            if (__request__.RequestContext?.Authorizer == null || __request__.RequestContext?.Authorizer.ContainsKey("<#= authKey #>") == false)
+            {
+                var __unauthorized__ = new <#= restApiAttribute != null || httpApiAttribute?.Data.Version == Amazon.Lambda.Annotations.APIGateway.HttpApiVersion.V1 ? "Amazon.Lambda.APIGatewayEvents.APIGatewayProxyResponse" : "Amazon.Lambda.APIGatewayEvents.APIGatewayHttpApiV2ProxyResponse" #>
```
normj, line 272: You you add a `__context__.Logger.LogDebug` here calling out the authorizer attribute `X` was missing so returning the unauthorized response.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.tt
```diff
@@ -247,6 +247,157 @@
...
+            }
+            
+            try
+            {
+                var __authValue_<#= parameter.Name #>__ = __request__.RequestContext.Authorizer["<#= authKey #>"];
+                <#= parameter.Name #> = (<#= parameter.Type.FullName #>)Convert.ChangeType(__authValue_<#= parameter.Name #>__?.ToString(), typeof(<#= parameter.Type.FullNameWithoutAnnotations #>));
+            }
+            catch (Exception e) when (e is InvalidCastException || e is FormatException || e is OverflowException || e is ArgumentException)
+            {
+                var __unauthorized__ = new <#= restApiAttribute != null || httpApiAttribute?.Data.Version == Amazon.Lambda.Annotations.APIGateway.HttpApiVersion.V1 ? "Amazon.Lambda.APIGatewayEvents.APIGatewayProxyResponse" : "Amazon.Lambda.APIGatewayEvents.APIGatewayHttpApiV2ProxyResponse" #>
```
normj, line 312: My understanding is if we go in here there is most likely a coding error like a malformed auth value. So I think we should log a warning using `__context__.Logger.LogWarning`.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.tt
normj: Can the values be numerics instead of strings? If so how would that work out with this approach? Or are we saying for the `FromCustomerAuthorizer` we only support string parameters. If so we should make sure we are enforcing that and put out diagnostic warnings if they attempt something we don't support.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.tt
```diff
@@ -247,6 +247,157 @@
...
+
+<#
+                }
+                else
+                {
+#>
+            var <#= parameter.Name #> = default(<#= parameter.Type.FullName #>);
+            if (__request__.RequestContext?.Authorizer?.Lambda == null || __request__.RequestContext?.Authorizer?.Lambda.ContainsKey("<#= authKey #>") == false)
+            {
+                var __unauthorized__ = new <#= restApiAttribute != null || httpApiAttribute?.Data.Version == Amazon.Lambda.Annotations.APIGateway.HttpApiVersion.V1 ? "Amazon.Lambda.APIGatewayEvents.APIGatewayProxyResponse" : "Amazon.Lambda.APIGatewayEvents.APIGatewayHttpApiV2ProxyResponse" #>
```
normj, line 353: You you add a `__context__.Logger.LogDebug` here calling out the authorizer attribute `X` was missing so returning the unauthorized response.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.tt
```diff
@@ -247,6 +247,157 @@
...
+            }
+            
+            try
+            {
+                var __authValue_<#= parameter.Name #>__ = __request__.RequestContext.Authorizer.Lambda["<#= authKey #>"];
+                <#= parameter.Name #> = (<#= parameter.Type.FullName #>)Convert.ChangeType(__authValue_<#= parameter.Name #>__?.ToString(), typeof(<#= parameter.Type.FullNameWithoutAnnotations #>));
+            }
+            catch (Exception e) when (e is InvalidCastException || e is FormatException || e is OverflowException || e is ArgumentException)
+            {
+                var __unauthorized__ = new <#= restApiAttribute != null || httpApiAttribute?.Data.Version == Amazon.Lambda.Annotations.APIGateway.HttpApiVersion.V1 ? "Amazon.Lambda.APIGatewayEvents.APIGatewayProxyResponse" : "Amazon.Lambda.APIGatewayEvents.APIGatewayHttpApiV2ProxyResponse" #>
```
normj, line 393: My understanding is if we go in here there is most likely a coding error like a malformed auth value. So I think we should log a warning using `__context__.Logger.LogWarning`.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.tt
GarrettBeatty: so what im doing here is converting it to a string then using https://learn.microsoft.com/en-us/dotnet/api/system.convert.changetype?view=net-10.0 afterwards to convert to the parameters type. 

i will add an integration test to confirm

GarrettBeatty: so i added integration tests to confirm

1. int values work
2. string values work
3. boolean values work

`double` values do not work because of a limitation with api gateway not supporting it.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.tt
GarrettBeatty: added logs here and in other places

PR #1501, opened 2023-05-01
Refactor the monolithic LambdaFunctionTemplate.tt

normj:
*Description of changes:*
All code generation for Amazon.Lambda.Annotations was being done in `LambdaFunctionTemplate.tt`. This was making it hard to follow all the logic for code generation and how to expand on it. This change  makes no logic changes but breaks the logical sections of `LambdaFunction.tt` leaving that tt as the outline for code generation and then invoke child generators depending on conditions. The code is still generated the exact same and all of the test pass.

image

Each tt file has a corresponding custom partial class that adds a custom constructor to pass in context parameters. In the future we could move some more of the logic from the tt into the custom cs file.

Note: I also add a `slnf` a solution filter file to make it easier to work on just the annotations project instead of having to recompile all of the other projects.


ashovlin:
> Note: I also add a slnf a solution filter file to make it easier to work on just the annotations project instead of having to recompile all of the other projects.

🎉 Nice, thanks!

ashovlin (approved):

philasmar (approved):

PR #1502, opened 2023-05-02
Fix Annotation bugs handling header names being case insensitive and parameters with nullable annotations

normj:
*Description of changes:*

The PR fixes 2 bugs with Lambda annotations. 

* If the `FromHeader` attribute is used support the name must support name being case insensitive. Unfortunately the dictionary for headers coming from `Amazon.Lambda.APIGateways` is not case insensitive. The generated code instead uses some Linq commands to check for existence and pull the value from it.
* If a parameter from the Lambda function is using the `?` nullability operator there was a bug where the `?` annotation was being included into the `typeof` operator. That is a compile error with something like `string?`. To test this I had to update the verifier base class to enable nullablility.

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.cs
@@ -226,7 +226,7 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Templates
             this.Write(")Convert.ChangeType(q, typeof(");
             
             #line 111 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\APIGatewaySetupParameters.tt"
-            this.Write(this.ToStringHelper.ToStringWithCulture(typeArgument.FullName));
+            this.Write(this.ToStringHelper.ToStringWithCulture(typeArgument.FullNameWithoutAnnotations));
             
             #line default
             #line hidden
@@ -301,7 +301,7 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Templates
             this.Write("\"], typeof(");
             
             #line 132 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\APIGatewaySetupParameters.tt"
-            this.Write(this.ToStringHelper.ToStringWithCulture(parameter.Type.FullName));
+            this.Write(this.ToStringHelper.ToStringWithCulture(parameter.Type.FullNameWithoutAnnotations));
             
             #line default
             #line hidden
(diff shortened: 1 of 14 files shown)


Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.cs
```diff
@@ -407,14 +407,15 @@ public virtual string TransformText()
...
-            this.Write("?.ContainsKey(\"");
+            this.Write("?.Any(x => string.Equals(x.Key, \"");
             
             #line 185 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\APIGatewaySetupParameters.tt"
             this.Write(this.ToStringHelper.ToStringWithCulture(headerKey));
             
             #line default
             #line hidden
-            this.Write("\") == true)\r\n            {\r\n                ");
+            this.Write("\", StringComparison.OrdinalIgnoreCase)) == true)\r\n            {\r\n                " +
```
philasmar, line 417: nit: we don't need "== true"

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.cs
```diff
@@ -489,15 +490,15 @@ public virtual string TransformText()
...
+            this.Write("?.Any(x => string.Equals(x.Key, \"");
             
             #line 209 "C:\codebase\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator\Templates\APIGatewaySetupParameters.tt"
             this.Write(this.ToStringHelper.ToStringWithCulture(headerKey));
             
             #line default
             #line hidden
-            this.Write("\") == true)\r\n            {\r\n                try\r\n                {\r\n             " +
-                    "       ");
+            this.Write("\", StringComparison.OrdinalIgnoreCase)) == true)\r\n            {\r\n                " +
```
philasmar, line 500: nit: we don't need "== true"

philasmar (approved):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.cs
normj: This is needed by cause the `?` is used on the left side of the comparison resulting in a nullable Boolean. You can't do `if(nullable<Boolean>)` so the easy work around is to compare the nullable Boolean to `true`.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/APIGatewaySetupParameters.cs
normj: This is needed by cause the `?` is used on the left side of the comparison resulting in a nullable Boolean. You can't do `if(nullable<Boolean>)` so the easy work around is to compare the nullable Boolean to `true`.

ashovlin (approved):

PR #1503, opened 2023-05-04
Fixed issues with starting from an empty project to adding Annotations

normj:
*Description of changes:*
Starting from a class library with just a method with no Lambda stuff in it I ran into the following issues turning a function into a Lambda function using Annotations.

* After adding the Amazon.Lambda.Annotations package and marking a function with the `LambdaFunction` attribute. I got a compilation error missing "Core". This happened because there was no reference to `Amazon.Lambda.Core` which is referenced for the `ILambdaContext`. **Fixed:** Added Amazon.Lambda.Core as a dependency for Amazon.Lambda.Annotations since you are always going to need it.

* After adding `Amazon.Lambda.Core` I was able to deploy but it failed because there was no registered `ILambdaSerializer`. **Fixed:** By adding a new diagnostics error informing the user during compilation that there is no assembly `LambdaSerializerAttribute`. Also discovered we don't support method level `LambdaSerializerAttribute` but that should be fixed in a separate PR.

* Added the `HttpApi` attribute to expose the function through API Gateway. I got the expected diagnostic error about missing reference to `Amazon.Lambda.APIGatewayEvents` but also "This is a bug" with a sub null reference exception diagnostic error from source generator. **Fixed:** The source generator wasn't treating the  missing `Amazon.Lambda.APIGatewayEvents` as a fatal error and kept running and eventually ran into a NPE because it couldn't find the `HttpApiAttribute` type. The error is now treated as fatal to stop processing right away.

**Note:** I never got the original reported issue about the source generator failing if there was no `aws-lambda-tools-defaults.json` file in the project. Since that report is old I suspect it has been fixed by some other previous change.

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
@@ -86,6 +86,15 @@ namespace Amazon.Lambda.Annotations.SourceGenerator
                 {
                     var lambdaMethodModel = semanticModelProvider.GetMethodSemanticModel(lambdaMethod);
 
+                    if (!HasSerializerAttribute(context, lambdaMethodModel))
+                    {
+                        diagnosticReporter.Report(Diagnostic.Create(DiagnosticDescriptors.MissingLambdaSerializer,
+                            lambdaMethod.GetLocation()));
+
+                        foundFatalError = true;
+                        continue;
+                    }
+
                     // Check for necessary references
                     if (lambdaMethodModel.HasAttribute(context, TypeFullNames.RestApiAttribute)
                         || lambdaMethodModel.HasAttribute(context, TypeFullNames.HttpApiAttribute))
@@ -96,6 +105,9 @@ namespace Amazon.Lambda.Annotations.SourceGenerator
                             diagnosticReporter.Report(Diagnostic.Create(DiagnosticDescriptors.MissingDependencies,
                                 lambdaMethod.GetLocation(),
                                 "Amazon.Lambda.APIGatewayEvents"));
+
+                            foundFatalError = true;
+                            continue;
                         }
                     }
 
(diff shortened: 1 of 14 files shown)


Libraries/src/Amazon.Lambda.Annotations.nuspec
```diff
@@ -1,15 +1,17 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
     <metadata>
         <id>Amazon.Lambda.Annotations</id>
-        <version>0.13.2</version>
+        <version>0.13.200</version>
```
ashovlin, line 4: Is this intentional? Or maybe needed a non-cached package locally?

normj: That was an accident leftover from local testing. I have reverted back to the current 0.13.2 version

ashovlin (approved):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
```diff
@@ -189,6 +201,16 @@ public void Execute(GeneratorExecutionContext context)
...
         }
 
+        private bool HasSerializerAttribute(GeneratorExecutionContext context, IMethodSymbol methodModel)
+        {
+            if(methodModel.ContainingAssembly.HasAttribute(context, TypeFullNames.LambdaSerializerAttribute))
+            {
+                return true;
+            }
+
+            return false;
```
philasmar, line 211: nit: simplify by returning the condition directly
```
return methodModel.ContainingAssembly.HasAttribute(context, TypeFullNames.LambdaSerializerAttribute)
```

Libraries/src/Amazon.Lambda.Annotations.nuspec
```diff
@@ -9,7 +9,9 @@
         <projectUrl>https://github.com/aws/aws-lambda-dotnet</projectUrl>
         <license type="expression">Apache-2.0</license>
         <dependencies>
-            <group targetFramework=".net6.0"/>
+            <group targetFramework=".net6.0">
+				<dependency id="Amazon.Lambda.Core" version="2.1.0" />
+			</group>
```
philasmar, line 14: not sure if this is a GitHub thing, but the alignment looks weird

philasmar (approved):

Issue #1340, opened 2022-10-17
Lambda annotations YAML generation does not support short hand syntax


jaknor:
### Describe the bug

When using Lambda Annotations we get a build error whenever our Cloudformation contains any of the short hand Cloudformation syntax, such as `!Ref`

### Expected Behavior

The solution should build without errors

### Current Behavior

We receive the error message

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	AWSLambda0001	This is a bug. Please run the build with detailed verbosity (dotnet build --verbosity detailed) and file a bug at https://github.com/aws/aws-lambda-dotnet with the build output and stack trace Encountered an unresolved tag '!Ref'   at YamlDotNet.Serialization.NodeTypeResolvers.PreventUnknownTagsNodeTypeResolver.YamlDotNet.Serialization.INodeTypeResolver.Resolve(NodeEvent nodeEvent, Type& currentType)	dotnet_lambda
```

### Reproduction Steps

Using lambda annotations define a lambda in code like below
```
[LambdaFunction(Name = "LambdaWithShortHandYaml")]
[RestApi(LambdaHttpMethod.Get, "/hello-world")]
public async Task LambdaWithShortHandYaml()
{
    await Task.Delay(1);
}
```
Then modify the Cloudformation to add an environment variable using the short hand synxax
```
  LambdaWithShortHandYaml:
    Type: AWS::Serverless::Function
    Metadata:
      Tool: Amazon.Lambda.Annotations
      SyncedEvents:
        - RootGet
    Properties:
      Runtime: dotnet6
      CodeUri: .
      MemorySize: 256
      Timeout: 30
      Policies:
        - AWSLambdaBasicExecutionRole
      PackageType: Zip
      Handler: dotnet_lambda::dotnet_lambda.Functions.Api_LambdaWithShortHandYaml_Generated::LambdaWithShortHandYaml
      Events:
        RootGet:
          Type: Api
          Properties:
            Path: /hello-world
            Method: GET
      Environment:
        Variables:
          MyVariable: !Ref Variable
```

Attempt to build the solution

### Possible Solution

It seems you are using YamlDotNet for parsing the YAML. They have an [open issue](https://github.com/aaubry/YamlDotNet/issues/521) to support Cloudformation short hand syntax which if finished would probably fix the issue

### Additional Information/Context

This fails even if disconnect the lambda in cloudformation from the annotations framework by removing the code below
```
Metadata:
[...]

PR #1532, opened 2023-06-16
fix: fix error thrown when template includes intrinsic function

philasmar:
*Issue #, if available:*
DOTNET-4300, #1340

*Description of changes:*
An error was thrown when a short form intrinsic function appears in a CloudFormation template. This error was due to an invalid casting to `Dictionary<string, object>`. Modifying this to a `Dictionary<string, YamlMappingNode>` for YAML templates fixed the issue.

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationWriter.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationWriter.cs
@@ -284,7 +284,7 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Writers
             }
 
             var toRemove = new List<string>();
-            foreach (var resourceName in _templateWriter.GetToken<Dictionary<string, object>>("Resources").Keys)
+            foreach (var resourceName in _templateWriter.GetKeys("Resources"))
             {
                 var resourcePath = $"Resources.{resourceName}";
                 var type = _templateWriter.GetToken<string>($"{resourcePath}.Type", string.Empty);
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/ITemplateWriter.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/ITemplateWriter.cs
@@ -1,4 +1,7 @@
-﻿namespace Amazon.Lambda.Annotations.SourceGenerator.Writers
+﻿using System.Collections;
+using System.Collections.Generic;
+
+namespace Amazon.Lambda.Annotations.SourceGenerator.Writers
 {
     /// <summary>
     /// This interface contains utility methods to manipulate a YAML or JSON blob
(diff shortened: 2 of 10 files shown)


philasmar:
> Can you also update the unit tests in [YamlWriterTests.cs](https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/WriterTests/YamlWriterTests.cs) to verify that YAML blobs with intrinsic functions are parsed correctly?

I added an intrinsic function to some of the tests, but it doesn't apply to all since this PR is not introducing a fix to parsing intrinsic functions. The PR is updating the Lambda Annotations code so workaround this limitation.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationWriter.cs
```diff
@@ -283,8 +284,18 @@ private void RemoveOrphanedLambdaFunctions(HashSet<string> processedLambdaFuncti
                 return;
             }
 
+            var resources = new List<string>();
+            if (_templateWriter is YamlWriter)
```
normj, line 288: Putting this logic in `CloudFormationWriter` breaks the abstraction of this class which is supposed to be agnostic of JSON or Yaml. Can you push this logic in `YamlWriter`'s `GetToken` method so that if the `T` is of type `Dictionary<string, object>` handle the conversion of `Dictionary<string, YamlMappingNode>` to `Dictionary<string, object>`?

Another option is to add to `ITemplateWriter` a `GetKeys(string path)` since all we need here is keys.

normj (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationWriter.cs
philasmar: Added a new function `GetKeys(string path)`

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationWriter.cs
```diff
@@ -10,6 +10,7 @@
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using YamlDotNet.RepresentationModel;
```
normj, line 13: nit: Can you remove this using statement now that it is no longer used.

normj (approved):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationWriter.cs
ashovlin: And the one in [JsonWriter.cs](https://github.com/aws/aws-lambda-dotnet/pull/1532/files#diff-743a8b526e00fd576280ee0173731c5f72116c40bc32306a7e6cded1ad540960) too?

96malhar (approved):

ashovlin (approved):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationWriter.cs
philasmar: Done

Issue #1398, opened 2022-12-14
Referencing other projects breaks AWS Lambda Test Tool


nathanvj:
### Describe the bug

Sorry I was quite busy and only just got the time to create an example repository.
My [previous issue](https://github.com/aws/aws-lambda-dotnet/issues/1371) got closed, so I decided to create a new one.

So to summarize, the following exception is thrown when I reference another project from my Lambda project:
```
FileNotFoundException: Could not load file or assembly 'LambdaTestToolError.Lambda, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified.
```

I would be very grateful if someone could take a look. The example repository can be found here:
[https://github.com/nathanvj/LambdaTestToolError](https://github.com/nathanvj/LambdaTestToolError)

Kind regards

### Expected Behavior

Be able to reference other projects from my Lambda project without a problem.

### Current Behavior

Whenever I add a project reference, the lambda test tool breaks. Note that the function works if I publish to AWS and do a test run from there, but I want to be able to use the Mock Lambda Test Tool to debug.

### Reproduction Steps

Example repo:
[https://github.com/nathanvj/LambdaTestToolError](https://github.com/nathanvj/LambdaTestToolError)

1. Clone
2. Open project
3. Run Mock Lambda Test Tool

### AWS .NET SDK and/or Package version used

Amazon.Lambda.Core 2.1.0
Amazon.Lambda.Serialization.SystemTextJson 2.3.0

### Targeted .NET Platform

.NET 6

### Operating System and version

Windows 10

ashishdhingra:
Reproducible using customer's code.

ashishdhingra:
This could be duplicate of https://github.com/aws/aws-lambda-dotnet/issues/1235.

nathanvj:
Seems indeed to be the same issue (apologies), and the provided workaround (deleting the .deps.json file) works fine for now but is pretty annoying to constantly do during development.

ashishdhingra:
@nathanvj Could you please verify if using the latest version `0.14.0` of Lambda Test tool fixes the issue reported here.

nathanvj:
Seems like this is fixed! Great work 👍🏻 

However if multiple projects contain a `aws-lambda-tools-defaults.json` file, only the functions defined in one of them are being shown in the tool - and only one of them can be found in `bin` folder. Is this expected behavior? Is there a way I can at least select which `aws-lambda-tools-defaults.json` file to use?

My use case is:
[...]

Issue #1235, opened 2022-07-08
Unable to get the LambdaTestTool to find the assembly that needs to be debugged when using multiple projects.


geedsen:
### Describe the bug

Unable to get the LambdaTestTool to find the assembly that needs to be debugged

### Expected Behavior

Debugging starts in the correct assembly

### Current Behavior

It always errors that it cannot load the assembly

### Reproduction Steps

Have a solution with multiple projects, for example 
1) Project.Api
2) Project.Lambda
And Project.Lambda has a dependency on Project.Api.

When compiling, the bin\debug\net6 directory for Project.Lambda contains a  Project.Api.deps.json AND a Project.Lambda.deps.json.
The tools picks up the Project.Api.deps.json (FirstOrDefault) and can therefor never find the Project.Lambda.dll assembly

### Additional Information/Context

Workaround is to delete the not required *.deps.json files from the directory before running the tool.

### AWS .NET SDK and/or Package version used

AWSSDK.Lambda 3.7.13.18

### Targeted .NET Platform

NET6

### Operating System and version

Window 10

ashishdhingra:
Hi @geedsen,

Good morning.

Thanks for reporting the issue. Could you please share the sample code solution for reproduction? Also request you to add proper issue title for easy tracking. :blush:

Thanks,
Ashish

geedsen:
Sorry, I dont have a sample, Just production code. Could you not reproduce it?

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

ashishdhingra:
@geedsen Somehow, I'm unable to reproduce the issue. Used Visual Studio 2022 and `Empty Lambda Function` project template to create the Lambda project, and added `Project.Api` class library project. Used `AWS .NET Core 6.0 Mock Lambda Test Tool (0.12.3)` for debugging.
**Project.Api.csproj**
```XML
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

</Project>
```
**Project.Api.Operations**
```C#
namespace Project.Api
{
    public class Operations
    {
        public static int Add(int x, int y) => x + y;
    }
}
```
**Project.Lambda.csproj**
```XML
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
[...]

PR #1542, opened 2023-07-06
fix: Lambda test tool debugs the correct assembly when the Lambda project depends on other executable projects

96malhar:
*Issue #, if available:*
* DOTNET-6421
* https://github.com/aws/aws-lambda-dotnet/issues/1235
* https://github.com/aws/aws-lambda-dotnet/issues/1398

*Description of changes:*
When the Lambda project references other executable projects (such as an ASP.NET Core Web API project) the `./bin/{CONFIGURATION}/{TARGET_FRAMEWORK}` will contain multiple `*deps.json` file. The Lambda test tool currently selects the first `deps.json` file it finds and will try to load the assembly context from the corresponsing `.dll` file.

This PR ensures that the right assembly is selected via the following algorithm
* Loop through all `*deps.json` files and extract all the keys in the `dependencies` JSON blob.
* Loop through all `*deps.json` files again:
  * Extract the project name by removing `.deps.json` suffix
  * Check if the project name is part of the `dependencies` collection.
  * If it isn't then this is the root project. Resolve the assembly path for this project and return it

### DIFF
--- a/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LocalLambdaRuntime.cs
+++ b/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LocalLambdaRuntime.cs
@@ -53,20 +53,15 @@ namespace Amazon.Lambda.TestTool.Runtime
                 throw new DirectoryNotFoundException($"Directory containing built Lambda project does not exist {directory}");
             }
 
-            var depsFile = Directory.GetFiles(directory, "*.deps.json").FirstOrDefault();
-            if (depsFile == null)
-            {
-                throw new Exception($"Failed to find a deps.json file in the specified directory ({directory})");
-            }
+            var lambdaAssemblyPath = Utils.FindLambdaAssemblyPath(directory);
 
-            var fileName = depsFile.Substring(0, depsFile.Length - ".deps.json".Length) + ".dll";
-            if (!File.Exists(fileName))
+            if (string.IsNullOrEmpty(lambdaAssemblyPath) || !File.Exists(lambdaAssemblyPath))
             {
                 throw new Exception($"Failed to find Lambda project entry assembly in the specified directory ({directory})");
             }
 
             // The resolver provides the ability to load the assemblies containing the select Lambda function.
-            var resolver = new LambdaAssemblyLoadContext(fileName);
+            var resolver = new LambdaAssemblyLoadContext(lambdaAssemblyPath);
 
             var runtime = new LocalLambdaRuntime(resolver, directory, awsService);
             return runtime;
--- a/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Utils.cs
+++ b/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Utils.cs
@@ -9,6 +9,7 @@ using System.Runtime.InteropServices;
 using System.Text;
 using System.Text.Json;
 using Amazon.Lambda.TestTool.Runtime;
+using Newtonsoft.Json.Linq;
 
 namespace Amazon.Lambda.TestTool
 {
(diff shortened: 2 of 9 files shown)


ashishdhingra (commented):
There might be a weird scenario where there are multiple `.csproj` files in the the same directory. So using this logic, we might end up loading the incorrect DLL.

96malhar:
> There might be a weird scenario where there are multiple `.csproj` files in the the same directory. So using this logic, we might end up loading the incorrect DLL.
> 
> Although I'm unaware fully about the internals of test tool, just curious, why we could not load all the DLLs for projects found using this logic. At runtime, when function handler is executed, it should be able to resolve the DLL that contains that handler.

I am unsure of the speed and memory implications of loading all available dlls and would let @normj opine more on this.

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LocalLambdaRuntime.cs
```diff
@@ -53,20 +53,17 @@ public static ILocalLambdaRuntime Initialize(string directory, IAWSService awsSe
                 throw new DirectoryNotFoundException($"Directory containing built Lambda project does not exist {directory}");
             }
 
-            var depsFile = Directory.GetFiles(directory, "*.deps.json").FirstOrDefault();
-            if (depsFile == null)
-            {
-                throw new Exception($"Failed to find a deps.json file in the specified directory ({directory})");
-            }
+            var lambdaProjectDirectory = Utils.FindLambdaProjectDirectory(directory);
```
normj, line 56: There are at least 2 problems with the approach:
1) The assembly name and the folder name are not guaranteed to be 1 to 1. The assembly name can be changed checking the `AssemblyName` MSBuild property. The `AssemblyName` property could be in the project file or imported in from another msbuild file.
2) There not actually be any source or a project file. It is possible to have the test tool point only the binaries.

What I would suggest is if there are multiple deps.json files inspect the JSON files to see which one is the root. For example in the provided repository https://github.com/nathanvj/LambdaTestToolError there are 2 deps.json files,  `LambdaTestToolError.API.deps.json` and `LambdaTestToolError.Lambda.deps.json` . My rough algorithm would be:
* If only one deps.json file return using that file
* Loop through each deps.json file
  * Extract the name of the assembly from the file name of the deps.json. (Remove the .deps.json suffix)
  * Loop through all of the other deps.json files
  * For each `target` in the file inspect each assembly to see if the assembly for the deps.json file being inspect for rootness is in there. If it is not referenced in any other deps.json return true other false for whether it is the root deps.json file. 
For example in the example when we inspect the `LambdaTestToolError.API.deps.json` file for being the root and then search to see if `LambdaTestToolError.API` is referenced `LambdaTestToolError.Lambda.deps.json` we would return false for being root. Looking at the JSON snippet below the `LambdaTestToolError.Lambda` assembly referenced in the deps.json file has a dependency to `LambdaTestToolError.API`

```
  "targets": {
    ".NETCoreApp,Version=v6.0": {
      "LambdaTestToolError.Lambda/1.0.0": {
        "dependencies": {
          "Amazon.Lambda.Core": "2.1.0",
          "Amazon.Lambda.Serialization.SystemTextJson": "2.3.0",
          "LambdaTestToolError.API": "1.0.0"
        },
        "runtime": {
          "LambdaTestToolError.Lambda.dll": {}
        }
      },
      "Amazon.Lambda.Core/2.1.0": {
        "runtime": {
          "lib/net6.0/Amazon.Lambda.Core.dll": {
            "assemblyVersion": "1.0.0.0",
            "fileVersion": "1.0.0.0"
          }
        }
      },
     ...
```
When checking `LambdaTestToolError.Lambda.deps.json` we would see that ``LambdaTestToolError.Lambda` is not referenced as a dependency in any other file.

normj (changes requested):
You are going to need new tests for this fix. I think if you implement my suggested solution you can just add a collection of deps.json files in different folders to test different scenarios for determining what is the root deps.json file. I don't think you need to add full Lambda projects for all of the scenarios.

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Runtime/LocalLambdaRuntime.cs
96malhar: Implemented the suggested logic and added unit tests.

Also ensured that the  https://github.com/nathanvj/LambdaTestToolError project can be debugged by the updated local version of the test tool.

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Utils.cs
```diff
@@ -241,5 +241,56 @@ public static string SearchLatestCompilationDirectory(string debugDirectory)
...
+
+                var isRootProject = true;
+
+                for (int j = 0; j < depsFiles.Length; j++)
+                {
+                    if (i == j)
+                        continue;
+                    
+                    var depsFileContent = File.ReadAllText(depsFiles[j]);
+                    if (depsFileContent.Contains($"\"{projectName}\":"))
```
normj, line 278: You need to parse the JSON file and search each `dependencies` object and see if any of them have `projectName` as a property key. Just doing a string contains will trigger false positives. For example if the Lambda project is called `MyCompany.WebApp.Lambda` but there is another project called `MyCompany.WebApp.Lambda.Common` when seaching the deps file for `MyCompany.WebApp.Lambda` it return back false for being root because you would still find the string `MyCompany.WebApp.Lambda`  inside `MyCompany.WebApp.Lambda.Common`.

normj: Make sure the test cases cover this scenario.

normj (changes requested):

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Utils.cs
96malhar: Eventhough I am not specifically searching for the `dependencies` blob, I am always searching for `"PROJECT_NAME":` (notice the enclosing double-quotes and colon suffix). In the example you provided, we would correctly return `MyCompany.WebApp.Lambda` since we are checking for `"MyCompany.WebApp.Lambda":` in the file contents.

However, to make this more robust, I have updated the PR by parsing all available `dependencies` blob.

The above scenario is already covered by the `MultipleDepsJsonFile` where the dependency graph is as follows:
1. `LambdaDemo` depends on `LambdaDemo.Api1` and `LambdaDemo.Api2`
2. `LambdaDemo.Api1` depends on `LambdaDemo.Api2`

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Utils.cs
```diff
@@ -241,5 +242,66 @@ public static string SearchLatestCompilationDirectory(string debugDirectory)
...
+        private static void ExtractDependenciesFromDepsJson(JsonElement node, HashSet<string> dependencies)
+        {
+            if (node.ValueKind != JsonValueKind.Object)
+                return;
+
+            if (node.TryGetProperty("dependencies", out var depenciesBlob)) 
+            {
+                foreach (var dependency in depenciesBlob.EnumerateObject())
+                {
+                    dependencies.Add(dependency.Name);
```
normj, line 296: Since the same dependency could be used across multiple assemblies you should check to see if `dependency.Name` has already been added to `dependencies`.

normj (approved):
Giving provisional approval assuming the `Contains` check is added for the `dependencies` collection.

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Utils.cs
96malhar: The dependencies collection is already a Hashset so we don't need to check for duplicates.

96malhar:
> Giving provisional approval assuming the `Contains` check is added for the `dependencies` collection.

The dependencies collection is already a HashSet so we don't have to worry about duplicates.

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/Utils.cs
normj: My mistake. I was thinking the `Add` method would throw an exception if attempting to add a duplicate. I see now that the `Add` returns back a boolean whether the item was added.

philasmar (approved):

Issue #1553, opened 2023-07-30
Partial batch failure support with Amazon.Lambda.KinesisEvents


danny-zegel-zocdoc:
### Describe the feature

When processing a kinesis stream configured with `reportBatchItemFailures` as `true` I expected to find a `StreamsEventResponse` object defined in `Amazon.Lambda.KinesisEvents` so I can report a partial batch failure similar to how `Amazon.Lambda.DynamoDBEvents` has `StreamsEventResponse` and `Amazon.Lambda.SQSEvents` has `SQSBatchResponse`.

The documentation here https://docs.aws.amazon.com/lambda/latest/dg/with-kinesis.html#services-kinesis-batchfailurereporting indicates that this is already available for Java and I would appreciate this feature added for dotnet.

### Use Case

This feature would be appreciated when using a Lambda to process a Kinesis stream that has `ReportBatchItemFailures` configured to `true`.

### Proposed Solution

Implement `StreamsEventResponse` in the `Amazon.Lambda.KinesisEvents` package.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

Amazon.Lambda.KinesisEvents

### Targeted .NET Platform

.NET 6 and up

### Operating System and version

N/A

ashishdhingra:
This has been implemented in [Amazon.Lambda.KinesisEvents v2.1.0](https://www.nuget.org/packages/Amazon.Lambda.KinesisEvents/2.1.0).

Issue #1538, opened 2023-06-27
Tumbling Window Events for Kinesis & Dynamo


benlongo:
### Describe the feature

The dotnet SDK appears to be missing this event type <https://javadoc.io/static/com.amazonaws/aws-lambda-java-events/3.11.2/com/amazonaws/services/lambda/runtime/events/KinesisTimeWindowEvent.html>

### Use Case

I'm trying to create a tumbling window event function in dotnet for kinesis. 

### Proposed Solution

Implement `KinesisTimeWindowEvent`

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

<PackageVersion Include="Amazon.Lambda.KinesisEvents" Version="2.0.0" />

### Targeted .NET Platform

.NET 7

### Operating System and version

Windows 11

ashishdhingra:
@benlongo Thanks for submitting feature request. This needs to be reviewed with the team. Feel free to contribute PR with associated test case which could be reviewed by the team.

Thanks,
Ashish

benlongo:
I misunderstood how tumbling windows behaved so I don't need this anymore, but it still is a gap in the SDK so I'll leave the issue open.

PR #1559, opened 2023-08-01
Added support for DynamoDBTimeWindowEvent, KinesisTimeWindowEvent and StreamsEventResponse (for KinesisEvent).

ashishdhingra:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/1538
https://github.com/aws/aws-lambda-dotnet/issues/1553

*Description of changes:*
- https://github.com/aws/aws-lambda-dotnet/issues/1538 (commit https://github.com/aws/aws-lambda-dotnet/pull/1559/commits/78a42a0ba9f384a71ce838aec668fe957ba9f927): Added support for `DynamoDBTimeWindowEvent` and `KinesisTimeWindowEvent`.
  - Documentation at https://docs.aws.amazon.com/lambda/latest/dg/with-kinesis.html#services-kinesis-windows has a note `For Java functions, we recommend using a Map<String, String> to represent the state.`. So in the code change, `Dictionary<string, string>` type was used. However, Newtonsoft Json deserializer throws error while converting int/long value to `string` key value. Hence, new `JsonConverter` was also added in `Amazon.Lambda.Serialization.SystemTextJson` (googled around for logic submitted by community).

- https://github.com/aws/aws-lambda-dotnet/issues/1553 (commit https://github.com/aws/aws-lambda-dotnet/pull/1559/commits/b1c92b9d52256e645c1e948ce3364010511fda8f): Added new `StreamsEventResponse` class for reporting batch item failures when processing streams for `KinesisEvent`.

**Reference:**
- https://github.com/aws/aws-lambda-go/blob/main/events/time_window.go#L14
- https://docs.aws.amazon.com/lambda/latest/dg/with-kinesis.html#services-kinesis-windows
- Java Lambda implementation:
  - https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/KinesisTimeWindowEvent.java
  - https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/DynamodbTimeWindowEvent.java
  - https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/models/TimeWindow.java
___


Libraries/src/Amazon.Lambda.KinesisEvents/KinesisTimeWindowEvent.cs
```diff
@@ -0,0 +1,58 @@
+namespace Amazon.Lambda.KinesisEvents
+{
+    using System;
+    using System.Collections.Generic;
+
```
philasmar, line 5: same comment about Amazon header and using statements.

Libraries/src/Amazon.Lambda.KinesisEvents/KinesisTimeWindowResponse.cs
```diff
@@ -0,0 +1,48 @@
+﻿namespace Amazon.Lambda.KinesisEvents
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Runtime.Serialization;
+
```
philasmar, line 6: same comment about Amazon header and using statements.

Libraries/src/Amazon.Lambda.KinesisEvents/README.md
```diff
@@ -36,3 +36,86 @@ public class Function
...
+                return new StreamsEventResponse() { BatchItemFailures = batchItemFailures };
+            }
+        }
+
+        return new StreamsEventResponse() { BatchItemFailures = batchItemFailures };
+    }
+}
+```
+
+The following is a sample class and Lambda function that receives Amazon Kinesis event when using time windows as an input and uses `KinesisTimeWindowResponse` object to demonstrates how to aggregate and then process the final state. (Note that by default anything written to Console will be logged as CloudWatch Logs events.)
```
philasmar, line 72: nit: "demonstrates" should be "demonstrate"

Libraries/src/Amazon.Lambda.KinesisEvents/StreamsEventResponse.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace Amazon.Lambda.KinesisEvents
+{
+    using System.Collections.Generic;
+    using System.Runtime.Serialization;
+
```
philasmar, line 5: same comment as before

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBTimeWindowEvent.cs
```diff
@@ -0,0 +1,58 @@
+namespace Amazon.Lambda.DynamoDBEvents
```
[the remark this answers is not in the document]
ashishdhingra: None of the Events classes have licensing headers. I recall there was a call out from Alex many months back to create backlog item to add licensing header to all classes.

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBTimeWindowEvent.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    using System;
+    using System.Collections.Generic;
```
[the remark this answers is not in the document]
ashishdhingra: Thanks for pointing out. I just followed the convention in existing classes (copy/paste).

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBTimeWindowResponse.cs
```diff
@@ -0,0 +1,48 @@
+﻿namespace Amazon.Lambda.DynamoDBEvents
```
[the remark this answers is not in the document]
ashishdhingra: None of the Events classes have licensing headers. I recall there was a call out from Alex many months back to create backlog item to add licensing header to all classes.

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBTimeWindowResponse.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Runtime.Serialization;
```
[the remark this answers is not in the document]
ashishdhingra: Thanks for pointing out. I just followed the convention in existing classes (copy/paste).

Libraries/src/Amazon.Lambda.KinesisEvents/KinesisTimeWindowResponse.cs
ashishdhingra: Thanks for pointing out. I just followed the convention in existing classes (copy/paste). None of the event classes have licensing header.

Libraries/src/Amazon.Lambda.DynamoDBEvents/README.md
```diff
@@ -56,3 +56,54 @@ public class Function
     }
 }
 ```
+
+The following is a sample class and Lambda function that receives Amazon DynamoDB event when using time windows as an input and uses `DynamoDBTimeWindowResponse` object to demonstrates how to aggregate and then process the final state. (Note that by default anything written to Console will be logged as CloudWatch Logs events.)
```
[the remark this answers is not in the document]
ashishdhingra: Fixed

Libraries/src/Amazon.Lambda.KinesisEvents/README.md
ashishdhingra: Fixed

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBTimeWindowResponse.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// <summary>
+    /// Response type to return a new state for the time window and to report batch item failures for DynamoDBTimeWindowResponse.
```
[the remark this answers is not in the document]
ashishdhingra: Removed DynamoDBTimeWindowResponse from documentation comment. The code was ported from Java/Go lambda library.

Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/Converters/LongToStringJsonConverter.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using System.Buffers;
+using System.Buffers.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Amazon.Lambda.Serialization.SystemTextJson.Converters
+{
+    public class LongToStringJsonConverter : JsonConverter<string>
```
normj, line 9: This converter makes me nervous. It will run for every `string` value that is serialized. Since this is just for the `State` property for Kinesis we shouldn't add this to the global list of converters we run. 

I suggest we:
* Change the converter to use `Dictionary<string, string>` instead of `string` as the generic and then implement for a Dictionary.
* Rename it to `DictionaryLongToStringJsonConverter`
* Instead of adding the converter to the global list decorate the `State` attributes in the Kinesis POCOs with the `JsonConverter` attribute. That way this converter is only used on the scenarios where we know the special scenario exists.
* Update the documentation on the Kinesis POCOs informing users that they are only compatible with the serializers using System.Text.json unless you intend to also write the equivalent for Newtonsoft.

normj (changes requested):

Libraries/src/Amazon.Lambda.DynamoDBEvents/Converters/DictionaryLongToStringJsonConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+                dictionary.Add(propertyName, keyValue);
+            }
+
+            return dictionary;
+        }
+
+        public override void Write(Utf8JsonWriter writer, Dictionary<string, string> value, JsonSerializerOptions options)
+        {
+            // Use the built-in serializer, because it can handle dictionaries with string keys.
+            JsonSerializer.Serialize(writer, value, options);
```
normj, line 51: Wouldn't this write the value as a string but given the context of this converter shouldn't it be written as a long?

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBTimeWindowResponse.cs
```diff
@@ -0,0 +1,56 @@
+﻿namespace Amazon.Lambda.DynamoDBEvents
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Runtime.Serialization;
+
+#if !NETSTANDARD2_0
```
normj, line 7: Instead of using the `!` operator just say `NETCOREAPP_3_1` like every else in our libraries. Pretty soon we will need to do a sweep across all of our libraries to change all references of `NETCOREAPP_3_1` to `NETCOREAPP3_1_OR_GREATER` and it will be easier if we are consistent with are preprocessor usage. I don't want to do that work because I need to double check the .NET SDK version running in the build system to ensure it is new enough that it has the `GREATER` preprocessor symbols.

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBTimeWindowEvent.cs
```diff
@@ -0,0 +1,66 @@
...
+
+#if !NETSTANDARD2_0
```
normj, line 6: Instead of using the `!` operator just say `NETCOREAPP_3_1` like every else in our libraries. Pretty soon we will need to do a sweep across all of our libraries to change all references of `NETCOREAPP_3_1` to `NETCOREAPP3_1_OR_GREATER` and it will be easier if we are consistent with are preprocessor usage. I don't want to do that work because I need to double check the .NET SDK version running in the build system to ensure it is new enough that it has the `GREATER` preprocessor symbols.

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBTimeWindowEvent.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        /// <summary>
+        /// Time window for the records in the event.
+        /// </summary>
+        public TimeWindow Window { get; set; }
+
+        /// <summary>
+        /// State being built up to this invoke in the time window.
+        /// </summary>
+#if !NETSTANDARD2_0
```
normj, line 25: Instead of using the `!` operator just say `NETCOREAPP_3_1` like every else in our libraries. Pretty soon we will need to do a sweep across all of our libraries to change all references of `NETCOREAPP_3_1` to `NETCOREAPP3_1_OR_GREATER` and it will be easier if we are consistent with are preprocessor usage. I don't want to do that work because I need to double check the .NET SDK version running in the build system to ensure it is new enough that it has the `GREATER` preprocessor symbols.

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBTimeWindowResponse.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        /// <summary>
+        /// New state after processing a batch of records.
+        /// </summary>
+        [DataMember(Name = "state")]
+#if NETCOREAPP_3_1
+        [System.Text.Json.Serialization.JsonPropertyName("state")]
+#endif
+#if !NETSTANDARD2_0
+        [JsonConverter(typeof(DictionaryLongToStringJsonConverter))]
```
normj, line 26: This can be combined with the previous `NETCOREAPP_3_1` condition block.

normj (changes requested):

Libraries/src/Amazon.Lambda.DynamoDBEvents/Converters/DictionaryLongToStringJsonConverter.cs
ashishdhingra: Yes, it would. I still haven’t received response from Lambda team. My hunch is that the reason there is a note to use `Map<string, string>` for Java is that sometimes the state might contain string values. In such case, we would not be sure on whether to convert it to long or string. That’s the reason I’m handling token type string as well in ExtractValue. Please advise.

philasmar (approved):
Approved, assuming you will address Norm's comments.

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBTimeWindowResponse.cs
ashishdhingra: I have changed it to `NETCOREAPP3_1_OR_GREATER` (else it was failing test case) since test case was having this conditional compilation check. Tested locally.

Libraries/src/Amazon.Lambda.DynamoDBEvents/Converters/DictionaryLongToStringJsonConverter.cs
ashishdhingra: As discussed, got response from Lambda team that `State` field could have values represented as strings and in some cases long.

Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/Converters/LongToStringJsonConverter.cs
ashishdhingra: Done for most part. I don't think we need to update the documentation (for last point) since it should work.

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBTimeWindowResponse.cs
ashishdhingra: I have changed it to `NETCOREAPP3_1_OR_GREATER` (else it was failing test case) since test case was having this conditional compilation check. Tested locally.

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBTimeWindowEvent.cs
ashishdhingra: I have changed it to `NETCOREAPP3_1_OR_GREATER` (else it was failing test case) since test case was having this conditional compilation check. Tested locally.

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBTimeWindowEvent.cs
ashishdhingra: I have changed it to `NETCOREAPP3_1_OR_GREATER` (else it was failing test case) since test case was having this conditional compilation check. Tested locally.

normj (approved):

Issue #1571, opened 2023-08-21
Support for executable assemblies when using Lambda Annotations


jeastham1993:
### Describe the feature

The current implementation of Lambda Annotations relies on using the class based handler approach. Add support for executable assemblies so that runtimes that require the Lambda runtime to be manually bootstrapped can also be used. Behind the scenes, this will generate the static main method used as the entrypoint.

### Use Case

As a developer, I want to use non-managed runtime versions of .NET but this requires me to understand how to bootstrap the runtime manually.

### Proposed Solution

Update the existing source generator to generate a `Program.cs` file if targetting an executable assembly. If there are multiple handlers in the same Lambda function add an Environment Variable to determine which handler is used.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

N/A, this is a new feature.

### Targeted .NET Platform

.NET 6, .NET Standard 2.0

### Operating System and version

All

jeastham1993:
I have a PR ready to raise for this, just looking to get the tests working.

ashishdhingra:
Needs review with the team. @jeastham1993 Feel free to contribute PR.

jeastham1993:
@ashishdhingra PR raised
https://github.com/aws/aws-lambda-dotnet/pull/1573

normj:
Closing the issue now that the PR has been merged and released.

Issue #1572, opened 2023-08-21
Method-level JSON serialisers are not respected in Annotations


mrmoo-zoopla:
### Describe the bug

When using the new Annotations framework, method-level JSON serialisers are "lost" during the code generation process, with the result that only one serialiser can be used, the one declared at assembly level.

### Expected Behavior

The serialiser specified on each lambda function method should be used for that function.

### Current Behavior

The assembly-level serialiser is the only one available.

### Reproduction Steps

To reproduce, create a C# library project which contains two classes which each express an operation as a lambda function.  Mark each with a `LambdaSerializer` attribute specifying a non-default serialiser which exhibits some of kind of special deserialisation behaviour

Test the two functions and see that the behaviour of the assembly-level serialiser is used.

### AWS .NET SDK and/or Package version used

Amazon.Lambda.Annotations 1.0.0

### Targeted .NET Platform

.NET 6

### Operating System and version

Windows 10 and Ubuntu

ashishdhingra:
This is most likely a feature request. Needs review with the team.

ashishdhingra:
Refer https://github.com/aws/aws-lambda-dotnet/pull/1573#discussion_r1301445977.

normj:
In newer versions of the Annotations library it respects the method level seriailizer attribute. I believe support was added in version 1.1.

PR #1573, opened 2023-08-21
#1571  Add annotations support for executable assemblies

jeastham1993:
*Issue #, if available:*
#1571 

*Description of changes:*
Add functionality for executable assemblies for annotations framework. This will auto generate the `static Main` method.

I'm particularly interested in feedback on the use of a new `LambdaOutputExecutable` annotation, I couldn't find a better way to establish if it was targetting an output assembly.


jeastham1993:
One other area I'd like some feedback on is the way the serializer is set within the executable assembly template. Currently it's hardcoded, in the same way it is in the LambdaFunctionModelBuilder class. It looks like there is an active TODO which would solve both points.

Libraries/src/Amazon.Lambda.Annotations/LambdaFunctionAttribute.cs
```diff
@@ -8,6 +8,11 @@ namespace Amazon.Lambda.Annotations
     [AttributeUsage(AttributeTargets.Method)]
     public class LambdaFunctionAttribute : Attribute
     {
+        /// <summary>
+        /// The serializer to use.
+        /// </summary>
+        public string Serializer { get; set; }
```
normj, line 14: We have a mix of approaches here. For class libraries user's don't set this value but instead use the assembly attribute. But for executable they would need to set this property. This will cause class library users to think they can set this property and that will work. I think we need to resolve what is the serializer via the attributes so we have a consistent experience.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
```diff
@@ -175,9 +195,17 @@ public void Execute(GeneratorExecutionContext context)
                     // report every generated file to build output
                     diagnosticReporter.Report(Diagnostic.Create(DiagnosticDescriptors.CodeGeneration, Location.None, $"{model.GeneratedMethod.ContainingType.Name}.g.cs", sourceText));
 
+                    lambdaModels.Add(model);
                     annotationReport.LambdaFunctions.Add(model);
                 }
 
+                if (receiver.IsExecutable)
+                {
+                    var executableAssembly = new ExecutableAssembly(lambdaModels, lambdaModels[0].LambdaMethod.ContainingNamespace);
```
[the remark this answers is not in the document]
normj: We also need to add a diagnostic error if the project is missing a reference to `Amazon.Lambda.RuntimeSupport`. Otherwise users will get confused when they add the new executable attribute and they start getting weird compiler errors. You can find some examples in this library where we check to make sure their is dependency to the API Gateway event package if they are creating an API Gateway based function.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/ExecutableAssembly.tt
```diff
@@ -0,0 +1,69 @@
...
+using Amazon.Lambda.Core;
+
+namespace <#= this._containingNamespace #>;
+
+public class Program
+{
+    private static async Task Main(string[] args)
+    {
+<#
+    if (this._lambdaFunctions.Count == 1)
```
normj, line 25: Do we need to special case the situation where there is just one function? The CloudFormation template should be updated to set the `ANNOTATIONS_HANDLER` environment variable no matter if there is 1 or a 100 functions.

normj (changes requested):
The code in the `Writers` namespace needs to be updated to sync the CloudFormation template with the handler being the assembly name and setting the `ANNOTATIONS_HANDLER` environment variable when using executable.

Libraries/src/Amazon.Lambda.Annotations/LambdaOutputExecutableAttribute.cs
```diff
@@ -0,0 +1,13 @@
+﻿
+using System;
+
+namespace Amazon.Lambda.Annotations
+{
+    /// <summary>
+    /// Indicates this the Lambda function is going to target an executable instead of a class based handler.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
```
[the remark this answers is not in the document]
jeastham1993: I can add it as an assembly attribute, but I didn't think it was possible to access an assembly attribute in a source generator because the assembly doesn't exist yet

Libraries/src/Amazon.Lambda.Annotations/LambdaOutputExecutableAttribute.cs
```diff
@@ -0,0 +1,13 @@
...
+    public class LambdaOutputExecutableAttribute : Attribute
```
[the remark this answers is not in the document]
jeastham1993: Yep, agreed

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/ExecutableAssembly.tt
```diff
@@ -0,0 +1,69 @@
...
+    {
+#>
+        Func<<#= string.Join(", ", this._lambdaFunctions[0].GeneratedMethod.Parameters.Select(p => $"{p.Type.FullName}")) #>, <#= this._lambdaFunctions[0].GeneratedMethod.ReturnType.FullName #>> <#=this._lambdaFunctions[0].LambdaMethod.Name.ToLower()#> = <#=this._lambdaFunctions[0].LambdaMethod.ContainingType.Name#>_<#=this._lambdaFunctions[0].LambdaMethod.Name#>_Generated.<#=this._lambdaFunctions[0].LambdaMethod.Name#>;
```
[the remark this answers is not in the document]
jeastham1993: That's handled above with the check if the return type is void.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/ExecutableAssembly.tt
```diff
@@ -0,0 +1,69 @@
...
+        await Amazon.Lambda.RuntimeSupport.LambdaBootstrapBuilder.Create(<#=this._lambdaFunctions[0].LambdaMethod.Name.ToLower()#>, new <#= this._lambdaFunctions[0].Serializer #>()).Build().RunAsync();
+<#
+    }
+    else
+    {
+        #>
+
+        switch (Environment.GetEnvironmentVariable("HANDLER"))
```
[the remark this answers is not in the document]
jeastham1993: Yep, agreed.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
```diff
@@ -19,14 +20,14 @@ using <#= ns #>;
 
 namespace <#= _model.LambdaMethod.ContainingNamespace #>
 {
-    public class <#= _model.GeneratedMethod.ContainingType.Name #>
+    public <#= _model.IsExecutable ? "static " : "" #>class <#= _model.GeneratedMethod.ContainingType.Name #>
```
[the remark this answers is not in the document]
jeastham1993: Yeah, that's a good point. Although if you were using mock lambda test tool for this implementation you'd be using the executable assembly feature anyway right?

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/ExecutableAssembly.tt
jeastham1993: I added this to remove the need for the environment variable when you've only got one function, but that does then have the knock on effect of complicating the cloud formation writer code. Updated to always use environment variables.

Libraries/src/Amazon.Lambda.Annotations/LambdaFunctionAttribute.cs
jeastham1993: Understood and updated to use the assembly OR the method attribute. I think this also fixes another issue that was opened yesterday #1572

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/SyntaxReceiver.cs
```diff
@@ -61,12 +67,26 @@ public void OnVisitSyntaxNode(GeneratorSyntaxContext context)
             if (context.Node is ClassDeclarationSyntax classDeclarationSyntax && classDeclarationSyntax.AttributeLists.Count > 0)
             {
                 // Get the symbol being declared by the class, and keep it if its annotated
-                var methodSymbol = context.SemanticModel.GetDeclaredSymbol(classDeclarationSyntax);
+                var methodSymbol = ModelExtensions.GetDeclaredSymbol(
```
normj, line 70: `ModelExtensions.GetDeclaredSymbol` is a C# extension method for `SemanticModel`. Is the reason you switch this code to call the extension method directly through the extension method type holder?

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/SyntaxReceiver.cs
```diff
@@ -61,12 +67,26 @@ public void OnVisitSyntaxNode(GeneratorSyntaxContext context)
...
+                    classDeclarationSyntax);
                 if (methodSymbol.GetAttributes().Any(attr => attr.AttributeClass.Name == nameof(LambdaStartupAttribute)))
                 {
                     StartupClasses.Add(classDeclarationSyntax);
                 }
             }
+            
+            if (context.Node is MethodDeclarationSyntax methodDeclaration)
+            {
+                var model = ModelExtensions.GetDeclaredSymbol(
```
normj, line 81: Similar comment as above. Just call the method from `context.SemanticModel.GetDeclaredSymbol`.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/ExecutableAssembly.tt
```diff
@@ -0,0 +1,56 @@
...
+<#
+        foreach (var model in this._lambdaFunctions)
+        {
+            #>
+            case "<#=model.LambdaMethod.Name#>":
+<#
+            if (model.GeneratedMethod.ReturnType.FullName == "void")
+            {
+#>
+                Action<<#= string.Join(", ", model.GeneratedMethod.Parameters.Select(p => $"{p.Type.FullName}")) #>> <#= model.LambdaMethod.Name.ToLower() #> = <#= model.LambdaMethod.ContainingType.Name #>_<#= model.LambdaMethod.Name #>_Generated.<#= model.LambdaMethod.Name #>;
```
normj, line 36: I think you are missing creating the instance of the generated type. For example in the test I see 
```
Func<string, string> toupper = Functions_ToUpper_Generated.ToUpper;
```
and expected to see
```
Func<string, string> toupper = new Functions_ToUpper_Generated().ToUpper;
```

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/ExecutableAssembly.tt
```diff
@@ -0,0 +1,56 @@
...
+                await Amazon.Lambda.RuntimeSupport.LambdaBootstrapBuilder.Create(<#= model.LambdaMethod.Name.ToLower() #>, new <#= this._lambdaFunctions[0].Serializer #>()).Build().RunAsync();
+                break;
+<#    
+            }
+            else
+            {
+#>
+                Func<<#= string.Join(", ", model.GeneratedMethod.Parameters.Select(p => $"{p.Type.FullName}")) #>, <#= model.GeneratedMethod.ReturnType.FullName #>> <#= model.LambdaMethod.Name.ToLower() #> = <#= model.LambdaMethod.ContainingType.Name #>_<#= model.LambdaMethod.Name #>_Generated.<#= model.LambdaMethod.Name #>;
```
normj, line 44: Same comment as above about creating the instance.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/FieldsAndConstructor.tt
```diff
@@ -8,18 +8,32 @@
     if (_model.LambdaMethod.UsingDependencyInjection)
     {
 #>
-        private readonly ServiceProvider serviceProvider;
+        private <#= _model.IsExecutable ? "static " : "" #>readonly ServiceProvider serviceProvider;
```
normj, line 11: The ternary operator for static should be removed since we aligned on the wrapping generated class always being an instance class.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/FieldsAndConstructor.tt
```diff
@@ -8,18 +8,32 @@
...
 <#
     }
     else
     {
 #>
-        private readonly <#= _model.LambdaMethod.ContainingType.Name #> <#= _model.LambdaMethod.ContainingType.Name.ToCamelCase() #>;
+        private <#= _model.IsExecutable ? "static " : "" #>readonly <#= _model.LambdaMethod.ContainingType.Name #> <#= _model.LambdaMethod.ContainingType.Name.ToCamelCase() #>;
```
normj, line 17: The ternary operator for static should be removed since we aligned on the wrapping generated class always being an instance class.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/FieldsAndConstructor.tt
```diff
@@ -8,18 +8,32 @@
...
+        private <#= _model.IsExecutable ? "static " : "" #>readonly <#= _model.Serializer #> serializer;
```
normj, line 18: The ternary operator for static should be removed since we aligned on the wrapping generated class always being an instance class.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/FieldsAndConstructor.tt
```diff
@@ -8,18 +8,32 @@
...
 <#
     }
 #>
 
+<#
+    if (_model.IsExecutable)
```
normj, line 24: Remove the static condition and static constructor. We are always using the instance constructor.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/LambdaFunctionTemplate.tt
```diff
@@ -26,7 +27,7 @@ this.Write(new FieldsAndConstructor(_model).TransformText());
 #>
 
 
-        public <#= _model.LambdaMethod.ReturnsVoidOrGenericTask ? "async " : "" #><#= _model.GeneratedMethod.ReturnType.FullName #> <#= _model.LambdaMethod.Name #>(<#= string.Join(", ", _model.GeneratedMethod.Parameters.Select(p => $"{p.Type.FullName} {p.Name}")) #>)
+        public <#= _model.IsExecutable ? "static " : "" #><#= _model.LambdaMethod.ReturnsVoidOrGenericTask ? "async " : "" #><#= _model.GeneratedMethod.ReturnType.FullName #> <#= _model.LambdaMethod.Name #>(<#= string.Join(", ", _model.GeneratedMethod.Parameters.Select(p => $"{p.Type.FullName} {p.Name}")) #>)
```
normj, line 30: Remove the static logic.

Libraries/src/Amazon.Lambda.Annotations/LambdaGenerateMainAttribute.cs
```diff
@@ -0,0 +1,13 @@
+﻿
+using System;
+
+namespace Amazon.Lambda.Annotations
+{
+    /// <summary>
+    /// Indicates this the Lambda function is going to target an executable instead of a class based handler.
```
normj, line 7: Doc suggestion:
```
Used when deploying Lambda functions as a .NET executable instead of a class library. Ensure the .NET project's output type is configured for `Console Application`. In the project file the `OutputType` property will be set to `Exe`.

Deploying as an executable versus a class library is required when compiling functions with Native AOT. It can also be useful to deploy as an executable to include specific versions of `Amazon.Lambda.RuntimeSupport` the .NET Lambda runtime client.
```

normj (changes requested):
Looks like we are getting close. Minor issues, some generated compilation errors address and some left over static logic that needs to be removed. Direction wise I don't have any major comments.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/SyntaxReceiver.cs
jeastham1993: Nope, no good reason. Updated.

normj (changes requested):
The integ tests are failing for me till I update this [assert](https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/test/TestServerlessApp.IntegrationTests/IntegrationTestContextFixture.cs#L58) from 25 to 26 for the new test function you added. Can you update the value as part of your PR?



I also build the package locally and did some adhoc testing which mostly went well but when I tried AOT I ran into problems. The problem is the library is setting the `runtime` in the CloudFormation template to `dotnet6` but for AOT we need it set for `provided.al2`. There is logic in the .NET deploy tooling preventing deploying an AOT executable to .NET 6.

What I'm wondering if we should repurpose the `LambdaGenerateMain` attribute into a general global properties attribute. Kind of mapping to the CloudFormation template global section. Something like:

```
[assembly:LambdaGlobalProperties(GenerateMain = true, Runtime = "provided.al2");
```

What do you think?

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
```diff
@@ -80,13 +84,29 @@ public void Execute(GeneratorExecutionContext context)
...
+                
+                var assemblyAttributes = context.Compilation.Assembly.GetAttributes();
+                
+                // Let's find the AssemblyTitleAttribute
+                var generateMainAttribute = assemblyAttributes
+                    .FirstOrDefault(attr => attr.AttributeClass.Name == nameof(LambdaGenerateMainAttribute));
+
+                if (generateMainAttribute != null)
+                {
+                    isExecutable = true;
```
normj, line 97: Check the `context.Compilation.CommonOptions.OutputKind` property to see if it is [OutputKind.ConsoleApplication](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.outputkind?view=roslyn-dotnet-4.6.0). Then write a diagnostic error. I forget to change this when I was converting the blueprint in Visual Studio to use this feature and had to figure out after deployment why I was getting an `Internal Error` which is not a great experience.

normj (changes requested):

jeastham1993:
@normj I like your idea of having Lambda Global properties. It also keeps it flexible for any future changes we might want to make. Updated and addressed all the comments.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
```diff
@@ -85,16 +91,38 @@ public void Execute(GeneratorExecutionContext context)
                 }
 
                 var isExecutable = false;
+
+                bool foundFatalError = false;
                 
                 var assemblyAttributes = context.Compilation.Assembly.GetAttributes();
                 
                 // Let's find the AssemblyTitleAttribute
                 var generateMainAttribute = assemblyAttributes
```
normj, line 100: nit: Rename variable to `globalPropertiesAttribute`

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
```diff
@@ -85,16 +91,38 @@ public void Execute(GeneratorExecutionContext context)
...
-                    .FirstOrDefault(attr => attr.AttributeClass.Name == nameof(LambdaGenerateMainAttribute));
+                    .FirstOrDefault(attr => attr.AttributeClass.Name == nameof(LambdaGlobalPropertiesAttribute));
+
+                var defaultRuntime = "dotnet6";
 
                 if (generateMainAttribute != null)
                 {
-                    isExecutable = true;
+                    var generateMain = generateMainAttribute.NamedArguments.FirstOrDefault(kvp => kvp.Key == "GenerateMain").Value;
```
normj, line 107: We are missing some null checks here for both `generateMain` and `runtime` since the `FirstOrDefault` will return null. That also means we are missing some unit test cases since it looks like we are only testing with both values set.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
```diff
@@ -85,16 +91,38 @@ public void Execute(GeneratorExecutionContext context)
...
+                    {
+                        diagnosticReporter.Report(Diagnostic.Create(DiagnosticDescriptors.InvalidRuntimeSelection, Location.None));
+                        return;
+                    }
+
+                    defaultRuntime = runtime;
+                    
+                    isExecutable = bool.Parse(generateMain.Value.ToString());
+                    
+                    // if (isExecutable && context.Compilation.Options.OutputKind != OutputKind.ConsoleApplication)
```
normj, line 121: Why is this code commented out? Seems like we need it.

Libraries/src/Amazon.Lambda.Annotations/LambdaGlobalPropertiesAttribute.cs
```diff
@@ -9,7 +9,10 @@ namespace Amazon.Lambda.Annotations
     /// Deploying as an executable versus a class library is required when compiling functions with Native AOT. It can also be useful to deploy as an executable to include specific versions of `Amazon.Lambda.RuntimeSupport` the .NET Lambda runtime client.
     /// </summary>
     [AttributeUsage( AttributeTargets.Assembly)]
-    public class LambdaGenerateMainAttribute : Attribute
+    public class LambdaGlobalPropertiesAttribute : Attribute
```
normj, line 12: We need some docs on the class and each of the properties.

normj (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
jeastham1993: @normj I wanted to have a chat about it, I meant to message you. When enabled, all the tests fail. I think it's because the test framework is not an executable, the OutputKind is always set to DLL. Regardless what options are set in the csproj file.

normj: I was able to get the `VerifyExecutableAssembly` test to pass when I uncommented this code and in the test I set the `OutputKind` property on the `TestState`  and when you are including the sources of `Amazon.Lambda.RuntimeSupport` exclude the `Program.cs` from that project. Below is my updated version of the test. I think you can do the same think to the other executable tests that are failing.

```
[Fact]
public async Task VerifyExecutableAssembly()
{
    var expectedTemplateContent = (await File.ReadAllTextAsync(Path.Combine("Snapshots", "ServerlessTemplates", "subnamespace_executableimage.template"))).ToEnvironmentLineEndings();
    var expectedSubNamespaceGenerated = (await File.ReadAllTextAsync(Path.Combine("Snapshots", "Functions_AsyncStartupToUpper_Generated.g.cs"))).ToEnvironmentLineEndings();
    var expectedProgramGenerated = (await File.ReadAllTextAsync(Path.Combine("Snapshots", "Program.g.cs"))).ToEnvironmentLineEndings();
            
    var test = new VerifyCS.Test
    {
        TestState =
        {
            OutputKind = OutputKind.ConsoleApplication,
            Sources =
            {
                (Path.Combine("TestExecutableServerlessApp", "Sub1", "Functions.cs"), await File.ReadAllTextAsync(Path.Combine("TestExecutableServerlessApp", "Sub1", "Functions.cs"))),
                (Path.Combine("TestExecutableServerlessApp", "Startup.cs"), await File.ReadAllTextAsync(Path.Combine("TestExecutableServerlessApp", "Startup.cs"))),
                (Path.Combine("TestExecutableServerlessApp", "Services", "SimpleCalculatorService.cs"), await File.ReadAllTextAsync(Path.Combine("TestExecutableServerlessApp", "Services", "SimpleCalculatorService.cs"))),
                (Path.Combine("Amazon.Lambda.Annotations", "LambdaFunctionAttribute.cs"), await File.ReadAllTextAsync(Path.Combine("Amazon.Lambda.Annotations", "LambdaFunctionAttribute.cs"))),
                (Path.Combine("Amazon.Lambda.Annotations", "LambdaStartupAttribute.cs"), await File.ReadAllTextAsync(Path.Combine("Amazon.Lambda.Annotations", "LambdaStartupAttribute.cs"))),
                (Path.Combine("Amazon.Lambda.Annotations", "LambdaGlobalPropertiesAttribute.cs"), await File.ReadAllTextAsync(Path.Combine("Amazon.Lambda.Annotations", "LambdaGlobalPropertiesAttribute.cs"))),
                (Path.Combine("TestExecutableServerlessApp", "AssemblyAttributes.cs"), await File.ReadAllTextAsync(Path.Combine("TestExecutableServerlessApp", "AssemblyAttributes.cs"))),
            },
            GeneratedSources =
            {
                (
                    typeof(SourceGenerator.Generator),
                    "Functions_ToUpper_Generated.g.cs",
                    SourceText.From(expectedSubNamespaceGenerated, Encoding.UTF8, SourceHashAlgorithm.Sha256)
                ),
                (
                    typeof(SourceGenerator.Generator),
                    "Program.g.cs",
                    SourceText.From(expectedProgramGenerated, Encoding.UTF8, SourceHashAlgorithm.Sha256)
                )
            },
            ExpectedDiagnostics =
            {
                new DiagnosticResult("AWSLambda0103", DiagnosticSeverity.Info).WithArguments("Functions_ToUpper_Generated.g.cs", expectedSubNamespaceGenerated),
                new DiagnosticResult("AWSLambda0103", DiagnosticSeverity.Info).WithArguments($"TestExecutableServerlessApp{Path.DirectorySeparatorChar}serverless.template", expectedTemplateContent),
            },
            ReferenceAssemblies = ReferenceAssemblies.Net.Net60
        }
    };

    foreach (var file in Directory.GetFiles(
                    Path.Combine("Amazon.Lambda.RuntimeSupport"),
                    "*.cs", SearchOption.AllDirectories))
    {
        var content = await File.ReadAllTextAsync(file);

        // Don't include RuntimeSupport's entry point.
        if (file.EndsWith("Program.cs") && content.Contains("Task Main(string[] args)"))
            continue;

        test.TestState.Sources.Add((file, content));
    }

    await test.RunAsync();
}
```

Libraries/src/Amazon.Lambda.Annotations/LambdaGlobalPropertiesAttribute.cs
```diff
@@ -0,0 +1,24 @@
+﻿
+using System;
+
+namespace Amazon.Lambda.Annotations
+{
+    /// <summary>
+    /// Used when deploying Lambda functions as a .NET executable instead of a class library. Ensure the .NET project's output type is configured for `Console Application`. In the project file the `OutputType` property will be set to `Exe`.
```
normj, line 7: This is docs for the `GenerateMain` property. The docs here should be about how this attribute is for setting global properties that will affect the code and CloudFormation generation for all of the projects in the assembly.

normj (changes requested):

normj (changes requested):
I have been playing with this PR and for the most part things have worked well. The one issue I have come across is if I change the `runtime` property to `provided.al2` all of the new functions get `provided.al2` as the runtime. But all of the existing functions didn't get updated. That makes it hard/confusing switching from .NET 6 managed to a .NET 7 Executable for a project.

jeastham1993:
@normj all feedback actioned, apologies for the delay.

Libraries/Amazon.Lambda.Annotations.slnf
```diff
@@ -2,11 +2,36 @@
   "solution": {
     "path": "Libraries.sln",
     "projects": [
-      "src\\Amazon.Lambda.APIGatewayEvents\\Amazon.Lambda.APIGatewayEvents.csproj",
       "src\\Amazon.Lambda.Annotations.SourceGenerator\\Amazon.Lambda.Annotations.SourceGenerator.csproj",
       "src\\Amazon.Lambda.Annotations\\Amazon.Lambda.Annotations.csproj",
+      "src\\Amazon.Lambda.APIGatewayEvents\\Amazon.Lambda.APIGatewayEvents.csproj",
```
normj, line 7: Why were all of these other packages added to the Annotations solution filter? I can't see why we would need all of these event projects or the ASP.NET Core bridge library.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
```diff
@@ -212,11 +277,89 @@ public void Execute(GeneratorExecutionContext context)
...
+                    Diagnostic.Create(
+                        DiagnosticDescriptors.MainMethodExists,
+                        Location.None));
+                    
+                return null;
+            }
+
+            return new ExecutableAssembly(
+                lambdaModels,
+                lambdaModels[0].LambdaMethod.ContainingNamespace);
```
normj, line 317: I suspect this is the code that is causing the bug that if there are no Lambda functions we give back the "AWSLambda001 This is a bug. Please run the build ..." compilation error.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/ExecutableAssembly.tt
```diff
@@ -0,0 +1,56 @@
...
+<#
+        foreach (var model in this._lambdaFunctions)
+        {
+            #>
+            case "<#=model.LambdaMethod.Name#>":
+<#
+            if (model.GeneratedMethod.ReturnType.FullName == "void")
+            {
+#>
+                Action<<#= string.Join(", ", model.GeneratedMethod.Parameters.Select(p => $"{p.Type.FullName}")) #>> <#= model.LambdaMethod.ExecutableAssemblyHandlerName #> = new <#= model.LambdaMethod.ContainingNamespace #>.<#= model.LambdaMethod.ContainingType.Name #>_<#= model.LambdaMethod.Name #>_Generated().<#= model.LambdaMethod.Name #>;
```
normj, line 36: If there are no parameters like the following function

```csharp
    [LambdaFunction]
    public void Post()
    {

    }
```

then the Action is setup with an empty generic parameter instead of no generic parameter
```
            case "Post":
                Action<> post_handler = new JamesAnnotationTest.Functions_Post_Generated().Post;
                await Amazon.Lambda.RuntimeSupport.LambdaBootstrapBuilder.Create(post_handler, new Amazon.Lambda.Serialization.SystemTextJson.DefaultLambdaJsonSerializer()).Build().RunAsync();
                break;
```

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/ExecutableAssembly.tt
```diff
@@ -0,0 +1,56 @@
...
+                await Amazon.Lambda.RuntimeSupport.LambdaBootstrapBuilder.Create(<#= model.LambdaMethod.ExecutableAssemblyHandlerName #>, new <#= this._lambdaFunctions[0].Serializer #>()).Build().RunAsync();
+                break;
+<#    
+            }
+            else
+            {
+#>
+                Func<<#= string.Join(", ", model.GeneratedMethod.Parameters.Select(p => $"{p.Type.FullName}")) #>, <#= model.GeneratedMethod.ReturnType.FullName #>> <#= model.LambdaMethod.ExecutableAssemblyHandlerName #> = new <#= model.LambdaMethod.ContainingNamespace #>.<#= model.LambdaMethod.ContainingType.Name #>_<#= model.LambdaMethod.Name #>_Generated().<#= model.LambdaMethod.Name #>;
```
normj, line 44: The same issue as the Action above. This function
```csharp
    [LambdaFunction]
    public string Default()
    {

    }
```

Creates a func with invalid comma at the start
```
                Func<, string> default_handler = new JamesAnnotationTest.Functions_Default_Generated().Default;
                await Amazon.Lambda.RuntimeSupport.LambdaBootstrapBuilder.Create(default_handler, new Amazon.Lambda.Serialization.SystemTextJson.DefaultLambdaJsonSerializer()).Build().RunAsync();
                break;
```

normj (changes requested):

Libraries/Amazon.Lambda.Annotations.slnf
jeastham1993: I don't know how that happened, removed.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
jeastham1993: Added a specific Diagnostic error to handle this.

Libraries/Amazon.Lambda.Annotations.slnf
normj: Doesn't look like you have pushed your change to remove all of these extra projects to the solution filter.

Libraries/Amazon.Lambda.Annotations.slnf.bak
```diff
@@ -0,0 +1,40 @@
+{
```
normj, line 1: This "bak" file should be removed from the PR.

normj (changes requested):

Libraries/Amazon.Lambda.Annotations.slnf
ashovlin: I also think we need Amazon.Lambda.RuntimeSupport added to the solution filter now:


```
Severity	Code	Description	Project	File	Line	Suppression State
Error	NU1105	Unable to find project information for 'C:\dev\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.RuntimeSupport\Amazon.Lambda.RuntimeSupport.csproj'. If you are using Visual Studio, this may be because the project is unloaded or not part of the current solution so run a restore from the command-line. Otherwise, the project file may be invalid or missing targets required for restore.	Amazon.Lambda.Annotations.SourceGenerators.Tests	C:\dev\aws-lambda-dotnet\Libraries\test\Amazon.Lambda.Annotations.SourceGenerators.Tests\Amazon.Lambda.Annotations.SourceGenerators.Tests.csproj	1	
Error	NU1105	Unable to find project information for 'C:\dev\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.RuntimeSupport\Amazon.Lambda.RuntimeSupport.csproj'. If you are using Visual Studio, this may be because the project is unloaded or not part of the current solution so run a restore from the command-line. Otherwise, the project file may be invalid or missing targets required for restore.	TestExecutableServerlessApp	C:\dev\aws-lambda-dotnet\Libraries\test\TestExecutableServerlessApp\TestExecutableServerlessApp.csproj	1	
```

jeastham1993: Updated, thanks @ashovlin

normj (approved):

ashovlin:
Still working my way through the code review, but could you add a header and section to https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.Annotations/README.md covering the new assembly attributes?

We have more content in there than the developer guides for annotations. 

Also okay if in a separate PR also targeting `dev` that ships as the same time as this.

jeastham1993:
@ashovlin all comments addressed

Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/SourceGeneratorTests.cs
```diff
@@ -218,26 +246,62 @@ public async Task ComplexCalculator()
...
+                        (Path.Combine("TestExecutableServerlessApp", "Services", "SimpleCalculatorService.cs"), await File.ReadAllTextAsync(Path.Combine("TestExecutableServerlessApp", "Services", "SimpleCalculatorService.cs"))),
+                        (Path.Combine("Amazon.Lambda.Annotations", "LambdaFunctionAttribute.cs"), await File.ReadAllTextAsync(Path.Combine("Amazon.Lambda.Annotations", "LambdaFunctionAttribute.cs"))),
+                        (Path.Combine("Amazon.Lambda.Annotations", "LambdaStartupAttribute.cs"), await File.ReadAllTextAsync(Path.Combine("Amazon.Lambda.Annotations", "LambdaStartupAttribute.cs"))),
+                        (Path.Combine("Amazon.Lambda.Annotations", "LambdaGlobalPropertiesAttribute.cs"), await File.ReadAllTextAsync(Path.Combine("Amazon.Lambda.Annotations", "LambdaGlobalPropertiesAttribute.cs"))),
+                        (Path.Combine("TestExecutableServerlessApp", "AssemblyAttributesInvalidRuntime.cs"), await File.ReadAllTextAsync(Path.Combine("TestExecutableServerlessApp", "AssemblyAttributesInvalidRuntime.cs"))),
```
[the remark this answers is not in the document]
ashovlin: After the latest commit aaaa5a8 I'm still failing a full build with
```
C:\dev\aws-lambda-dotnet\Libraries\test\TestExecutableServerlessApp\AssemblyAttributes.cs(4,12): error CS0579: Duplicat
e 'LambdaSerializer' attribute [C:\dev\aws-lambda-dotnet\Libraries\test\TestExecutableServerlessApp\TestExecutableServe
rlessApp.csproj]
C:\dev\aws-lambda-dotnet\Libraries\test\TestExecutableServerlessApp\AssemblyAttributes.cs(5,12): error CS0579: Duplicat
e 'LambdaGlobalProperties' attribute [C:\dev\aws-lambda-dotnet\Libraries\test\TestExecutableServerlessApp\TestExecutabl
eServerlessApp.csproj]
    741 Warning(s)
    2 Error(s)
```

I think due to [AttributeNullValues.cs](https://github.com/aws/aws-lambda-dotnet/pull/1573/files#diff-3311170d79f546f3ab1d44a7f86ffb84d454937a7c9d2c92cbb4ad6d470df5a7)

I also see this in Visual Studio, can we set `TestExecutableServerlessApp` to exe?
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	AWSLambda0111	AssemblyAttribute Amazon.Lambda.Annotations.LambdaGlobalPropertiesAttribute is configured to generate a static main method but the assembly itself is not configured to output an executable. Set the 'OutputType' property in the .csproj file to be 'exe'.	TestExecutableServerlessApp	C:\dev\aws-lambda-dotnet\Libraries\test\TestExecutableServerlessApp\TestExecutableServerlessApp.csproj	1	Active
```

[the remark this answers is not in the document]
jeastham1993: @ashovlin fixed.

ashovlin (approved):

normj:
Merged into `feature/annotations-executable` branch so I can do the version bump and added `provided.al2023` runtime.

Issue #1574, opened 2023-08-30
AWS Lamdba - Supporting binary response in ASP.NET Core 6


hugohlln:
### Describe the bug

Hello,

I'm creating an ASP.NET Core 6 Rest API running in a lambda behind an API Gateway.
We have somme issues since we would like to return image/webp content and we are not able to do it.

We've seen that a solution exists for ASP.NET Core 2 by calling the following method :
`RegisterResponseContentEncodingForContentType("image/webp", ResponseContentEncoding.Base64);`

But unforunately we didn't find any information or tutorial to do the same thing using ASP.NET Core 6 cause the lambda support is done as below in the Program.cs file :
`builder.Services.AddAWSLambdaHosting(LambdaEventSource.RestApi);`

Is there any way to register image/webp conte type for Base64 transformation in ASP.NET Core 6 using the `builder.Services.AddAWSLambdaHosting(LambdaEventSource.RestApi);` method ?

Thanks

### Expected Behavior

Be able to register content-types for Base64 transformation in ASP.NET Core 6 Program.cs file

### Current Behavior

Unable to register content-types for Base64 transformation in ASP.NET Core 6 Program.cs file

### Reproduction Steps

- Create an ASP.NET Core 6 rest api 
- Add the lambda support in the Program.cs file `(builder.Services.AddAWSLambdaHosting(LambdaEventSource.RestApi);)`
- Return a webp image in your endpoint :
`[HttpGet]
        public ActionResult GetWebp()
        {
            var image = System.IO.File.ReadAllBytes("myImage.webp");
            return File(image, "image/webp");
        }
    }`

### AWS .NET SDK and/or Package version used

Amazon.Lambda.AspNetCoreServer.Hosting 1.6.0

### Targeted .NET Platform

ASP.NET Core 6.0

### Operating System and version

Any

ashishdhingra:
@hugohlln Good morning. Thanks for opening the issue. Please refer to https://github.com/aws/aws-lambda-dotnet/discussions/1527, it might help. You need to use `LambdaEventSource.HttpApi` and return `APIGatewayHttpApiV2ProxyResponse`.

hugohlln:
Hi @ashishdhingra,

Thanks for your reply.
We wan't to keep our current returned object `ActionResult` ant not `APIGatewayHttpApiV2ProxyResponse` in our controllers.
[...]

PR #1575, opened 2023-08-31
Register encoding options when adding the service to the service collection

hugohlln:
**Issue:**: #1574

**Description of changes:**
Can now register content type for transformation by passing options to the HostingOptions object


brignolij:
Good idea, Very usefull for webp images with api gateway.

Libraries/src/Amazon.Lambda.AspNetCoreServer.Hosting/HostingOptions.cs
```diff
@@ -12,4 +13,9 @@ public class HostingOptions
     /// back to JSON to return to Lambda.
     /// </summary>
     public ILambdaSerializer Serializer { get; set; }
+
+    /// <summary>
+    /// The encoding response options used by the Lambda
+    /// </summary>
+    public IEncodingOptions EncodingOptions { get; set; }
```
normj, line 20: `IEncodingOptions` is defined in an `Internal` namespace so it shouldn't be exposed on the public API.

Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/EncodingOptions.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+
+namespace Amazon.Lambda.AspNetCoreServer.Internal
```
normj, line 3: Seems like the `EncodingOptions` and interface should be in `Amazon.Lambda.AspNetCoreServer.Hosting`. They are basically metadata of the `HostingOptions`.

Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/EncodingOptions.cs
```diff
@@ -0,0 +1,10 @@
...
+{
+    public class EncodingOptions : IEncodingOptions
```
normj, line 5: We need docs on the class and interface.

Libraries/src/Amazon.Lambda.AspNetCoreServer.Hosting/HostingOptions.cs
```diff
@@ -12,4 +13,9 @@ public class HostingOptions
     /// back to JSON to return to Lambda.
     /// </summary>
     public ILambdaSerializer Serializer { get; set; }
+
+    /// <summary>
+    /// The encoding response options used by the Lambda
```
normj, line 18: Can you add a sentence that informs the reader that this is used to adjust what content types have their data sent to API Gateway as binary data?

normj (changes requested):
Thanks for the PR. I think the approach to solve the problem of configuring binary data is good. A few minor comments plus we also need some tests in the `Amazon.Lambda.AspNetCoreServer.Test` project.

hugohlln:
@normj Thanks for your feedbacks

I just dit a new commit with following changes :
- Add comments
- Change the namespace of IEncodingOptions and EncodingOptions

For unit tests, what would be the best way to test the injection of IEncodingOptions ?

Libraries/src/Amazon.Lambda.AspNetCoreServer.Hosting/Internal/LambdaRuntimeSupportServer.cs
```diff
@@ -87,9 +87,28 @@ public class APIGatewayHttpApiV2MinimalApi : APIGatewayHttpApiV2ProxyFunction
             /// Create instances
             /// </summary>
             /// <param name="serviceProvider">The IServiceProvider created for the ASP.NET Core application</param>
-            public APIGatewayHttpApiV2MinimalApi(IServiceProvider serviceProvider)
+            /// <param name="encodingOptions">The encoding options to register</param>
+            public APIGatewayHttpApiV2MinimalApi(IServiceProvider serviceProvider, IEncodingOptions? encodingOptions = null)
                 : base(serviceProvider)
             {
+                if(encodingOptions != null)
```
CamileDahdah, line 94: This code block and the one under method APIGatewayRestApiMinimalApi are identical. Can you create a private method and call it from both methods?

hugohlln:
@CamileDahdah 

I just moved the registration of IEncodingOptions to LambdaRuntimeSupportServer level and removed redundant code from each AbstractAspNetCoreFunction<TREQUEST, TRESPONSE> implementations (APIGatewayHttpApiV2MinimalApi, APIGatewayRestApiMinimalApi and ApplicationLoadBalancerMinimalApi)

normj:
Closing as this feature was address in PR https://github.com/aws/aws-lambda-dotnet/pull/2267 that exposed all of the extension points from Amazon.Lambda.AspNetCoreServer onto the HostingOptions including `RegisterResponseContentEncodingForContentEncoding`. That PR was released as part of version 1.10.0 of Amazon.Lambda.AspNetCoreServer.Hosting.

@hugohlln I appreciate your effort but it made sense finally address getting all of the extension points over.

PR #1580, opened 2023-09-06
chore: restrict permissions for IAM role used by smoke tests

philasmar:
*Issue #, if available:*
DOTNET-7003

*Description of changes:*
* Added more detailed permissions for the smoke tests instead of specifying * from actions and resources.
* Removed creation of IAM role from smoke tests to Pipeline stack so that the tests don't require permissions to create, attach and delete roles which could pose security risks.

### DIFF
--- a/LambdaRuntimeDockerfiles/SmokeTests/test/ImageFunction.SmokeTests/ImageFunctionTests.cs
+++ b/LambdaRuntimeDockerfiles/SmokeTests/test/ImageFunction.SmokeTests/ImageFunctionTests.cs
@@ -41,7 +41,6 @@ namespace ImageFunction.SmokeTests
         private static readonly RegionEndpoint TestRegion = RegionEndpoint.USWest2;
         private readonly AmazonLambdaClient _lambdaClient;
         private readonly AmazonIdentityManagementServiceClient _iamClient;
-        private readonly string _executionRoleName;
         private string _executionRoleArn;
 
         private static readonly string LambdaAssumeRolePolicy =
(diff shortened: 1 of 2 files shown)


LambdaRuntimeDockerfiles/SmokeTests/test/ImageFunction.SmokeTests/ImageFunctionTests.cs
```diff
@@ -67,14 +66,14 @@ public class ImageFunctionTests : IDisposable
...
         {
-            _executionRoleName = $"{TestIdentifier}-{Guid.NewGuid()}";
             _functionName = $"{TestIdentifier}-{Guid.NewGuid()}";
             var lambdaConfig = new AmazonLambdaConfig()
             {
                 RegionEndpoint = TestRegion
             };
             _lambdaClient = new AmazonLambdaClient(lambdaConfig);
             _iamClient = new AmazonIdentityManagementServiceClient(TestRegion);
+            _executionRoleArn = Environment.GetEnvironmentVariable("AWS_LAMBDA_SMOKETESTS_LAMBDA_ROLE");
```
ashovlin, line 76: nit: Should we `Assert.NotNull` like the other environment variable? Not sure how this fails if it isn't set.

ashovlin (approved):

normj (commented):
Does this PR affect how you would run these tests locally since locally we are not likely to have the `AWS_LAMBDA_SMOKETESTS_LAMBDA_ROLE` environment variable set.

philasmar:
> Does this PR affect how you would run these tests locally since locally we are not likely to have the `AWS_LAMBDA_SMOKETESTS_LAMBDA_ROLE` environment variable set.

Previously, the tests themselves were creating the role and permissions you need. This was flagged by AppSec. So now to run these locally, you need to provide a role ARN via this environment variable.

LambdaRuntimeDockerfiles/SmokeTests/test/ImageFunction.SmokeTests/ImageFunctionTests.cs
ashovlin: Thanks, I'd rather fail fast then attempt to create things and fail.

normj (approved):

PR #1584, opened 2023-09-14
feat: added json library in `JSONExceptionWriter`

Anirudhrawat:
rewrite the *JsonExceptionWriter classes to use a JSON library instead of building strings directly.

*Description of changes:*
Rewrite the *JsonExceptionWriter classes to use a System.Text.JSON library instead of building strings directly.


Libraries/src/Amazon.Lambda.RuntimeSupport/ExceptionHandling/JsonExceptionWriterHelpers.cs
```diff
@@ -22,96 +22,16 @@ internal class JsonExceptionWriterHelpers
...
-
-                if (num < 10)
-                    hex[3 - i] = (char)('0' + num);
-                else
-                    hex[3 - i] = (char)('A' + (num - 10));
-
-                n >>= 4;
-            }
-            return hex;
+            return escapedString.Trim('"');
```
CamileDahdah, line 34: We should call .Trim() too here for consistency

Libraries/src/Amazon.Lambda.RuntimeSupport/ExceptionHandling/JsonExceptionWriterHelpers.cs
```diff
@@ -22,96 +22,16 @@ internal class JsonExceptionWriterHelpers
...
-            var sb = new StringBuilder(n * 2);
-            for (int i = 0; i < n; i++)
-            {
-                char c = str[i];
-                switch (c)
-                {
-                    case '\n':
-                        sb.Append(@"\n");
-                        break;
+            string escapedString = JsonSerializer.Serialize(str);
```
CamileDahdah, line 32: I was testing the output of this method with the one that already exists. They seem to differ.
For instance, with System.Text.Json, we use \u0027 to represent a single quote `'`. With the original code, we don't.

Input example which leads to different output results: `"  {\r\n    _id: '{{objectId()}}',\r\n    index: '{{index()}}',\r\n    guid: '{{guid()}}',\r\n    isActive: '{{bool()}}',\r\n    balance: '{{floating(1000, 4000, 2, \"$0,0.00\")}}',\r\n    picture: 'http://placehold.it/32x32',\r\n    age: '{{integer(20, 40)}}',\r\n    eyeColor: '{{random(\"blue\", \"brown\", \"green\")}}',\r\n    name: '{{firstName()}} {{surname()}}',\r\n    gender: '{{gender()}}',\r\n    company: '{{company().toUpperCase()}}',\r\n    email: '{{email()}}',\r\n    phone: '+1 {{phone()}}',\r\n    address: '{{integer(100, 999)}} {{street()}}, {{city()}}, {{state()}}, {{integer(100, 10000)}}',\r\n    about: '{{lorem(1, \"paragraphs\")}}',\r\n    registered: '{{date(new Date(2014, 0, 1), new Date(), \"YYYY-MM-ddThh:mm:ss Z\")}}',\r\n    latitude: '{{floating(-90.000001, 90)}}',\r\n    longitude: '{{floating(-180.000001, 180)}}',\r\n    tags: [\r\n      '{{repeat(7)}}',\r\n      '{{lorem(1, \"words\")}}'\r\n    ],  "`

Libraries/src/Amazon.Lambda.RuntimeSupport/ExceptionHandling/JsonExceptionWriterHelpers.cs
```diff
@@ -22,96 +22,16 @@ internal class JsonExceptionWriterHelpers
         /// <summary>
         /// This method escapes a string for use as a JSON string value.
         /// It was adapted from the PutString method in the LitJson.JsonWriter class.
```
CamileDahdah, line 24: Comment on line 24 should be deleted

Libraries/src/Amazon.Lambda.RuntimeSupport/ExceptionHandling/JsonExceptionWriterHelpers.cs
```diff
@@ -13,7 +13,7 @@
  * permissions and limitations under the License.
  */
 
-using System.Text;
+using System.Text.Json;
```
CamileDahdah, line 16: It seems like in this [commit](https://github.com/aws/aws-lambda-dotnet/commit/b19394e0f77cd7cc17e4826e8fbe7bcf9633ab40), we migrated from LitJSON to System.Text.Json and left this method as is. I am not sure why though.

Libraries/src/Amazon.Lambda.RuntimeSupport/ExceptionHandling/JsonExceptionWriterHelpers.cs
CamileDahdah: However, if we use LitJson. Code example below
`// Create a JsonData object to hold the string
JsonData jsonData = new JsonData(str);

// Serialize the JsonData object to a JSON string
string litjsonString = JsonMapper.ToJson(jsonData);
litjsonString = litjsonString.Trim('"').Trim();`

We get the same result

peterrsongg (changes requested):
To echo @CamileDahdah's point, I just wrote a unit test that tests the functionality of just the cases where `c > ' ' ` so basically everything except the default. Looks like your implementation does not work for the escaped quotation. At a minimum the PR for this should include a unit test that verifies that your implementation works. 

```
    public class UnitTest1
    {
 
        private static string convertCharacterStrings = "\n\r\t\"\\\f\b\u0085\u2028\u2029";
        [Fact]
        public void TestEscapeStringForJsonReturnsSameOutputAsSystemTextJson()
        {
            string escapeStringForJsonCharacter;
            string systemTextJsonCharacter;
            escapeStringForJsonCharacter = EscapeStringForJson(convertCharacterStrings);
            systemTextJsonCharacter = JsonSerializer.Serialize(convertCharacterStrings).Trim('"');

            Assert.Equal(escapeStringForJsonCharacter, systemTextJsonCharacter);
        }
}
```

peterrsongg:
Thank you for the PR but we can't take a dependency on LitJson. In the SDK we have a source-copy of LitJson, so I can see why you added the dependency but for this library we don't even have a source-copy in this library and we don't want to add another dependency. Going to close out this PR.

PR #1589, opened 2023-09-28
chore: update .NET8 OCIs to use AL2023 preview image

philasmar:
*Description of changes:*
* Switch from using Amazon AL2 base image to AL2023
* Use `dnf` package manager instead of `yum` as it is not available on AL2023
* Install LIBICU on both AMD64 and ARM64 as it is not available by default in AL2023
* Install LIBICU using `dnf` instead of building from source

### DIFF
--- a/LambdaRuntimeDockerfiles/Images/net8/amd64/Dockerfile
+++ b/LambdaRuntimeDockerfiles/Images/net8/amd64/Dockerfile
@@ -4,9 +4,11 @@ ARG ASPNET_VERSION=8.0.0-rc.1.23421.29
 ARG ASPNET_SHA512=d5f9e7bffbf2b48b26a317dd1d78bc866973b4a2cda448cd7a7ee64c0ffaf98fa3c4b8584d32528026674bdfd99f602f0fdac8242176815705e080df83825efa
 
 ARG LAMBDA_RUNTIME_NAME=dotnet8
-ARG AMAZON_LINUX=public.ecr.aws/lambda/provided:al2
+ARG AMAZON_LINUX=public.ecr.aws/lambda/provided:al2023-preview
 
 FROM $AMAZON_LINUX AS base
+
+RUN dnf install libicu-67.1-7.amzn2023.0.3.x86_64 --assumeyes
 
 FROM base AS builder-net8
 ARG ASPNET_VERSION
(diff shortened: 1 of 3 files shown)


LambdaRuntimeDockerfiles/Images/net8/arm64/Dockerfile
```diff
@@ -3,40 +3,12 @@
...
-    && rm icu-src.tgz
-
-# Build ICU library
-RUN mkdir /libicu
-WORKDIR /icu/source/
-RUN ./configure --prefix=/libicu \
-    && make \
-    && make install
-
+RUN dnf install libicu --assumeyes
```
ashovlin, line 11: Do you recall why we were building libicu from source on AL2? Okay with the change, but just looking for context.

philasmar: If I remember correctly, the AL2 had LIBICU preinstalled but it had a version older than the one we needed for newer .NET versions. It's package manager had specific packages and versions available for installation, and LIBICU was not one of them. On AL2023, LIBICU is not preinstalled on the image, but `dnf` has version 67 available which works for the .NET versions we have. As soon as newer LIBICU versions are required, we will go back to installing from source.

ashovlin (approved):

normj (changes requested):
I did do some manual sanity checks and everything worked fine. I didn't really do anything that our automation checks don't cover but just want to run in action just to double check.

LambdaRuntimeDockerfiles/Images/net8/arm64/Dockerfile
philasmar: Done

LambdaRuntimeDockerfiles/Images/net8/amd64/Dockerfile
```diff
@@ -4,18 +4,20 @@ ARG ASPNET_VERSION=8.0.0-rc.1.23421.29
 ARG ASPNET_SHA512=d5f9e7bffbf2b48b26a317dd1d78bc866973b4a2cda448cd7a7ee64c0ffaf98fa3c4b8584d32528026674bdfd99f602f0fdac8242176815705e080df83825efa
 
 ARG LAMBDA_RUNTIME_NAME=dotnet8
-ARG AMAZON_LINUX=public.ecr.aws/lambda/provided:al2
+ARG AMAZON_LINUX=public.ecr.aws/lambda/provided:al2023-preview
 
 FROM $AMAZON_LINUX AS base
 
+RUN dnf install libicu --assumeyes
```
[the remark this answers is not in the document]
philasmar: Done

normj (approved):

Issue #1594, opened 2023-10-16
OutOfMemoryException thrown after updating .NET 8 application to Amazon.Lambda.RuntimeSupport 1.9.0


martincostello:
### Describe the bug

Spotting that the 1.9.0 release of Amazon.Lambda.RuntimeSupport [added a `net8.0` TFM](https://github.com/aws/aws-lambda-dotnet/blob/4ebdeca4e8173b9d7a5d453b5e94080b041a65ba/RELEASE.CHANGELOG.md#release-2023-10-13), I [updated an existing application of mine](https://github.com/martincostello/adventofcode/pull/1277) running .NET 8 RC2 in AWS Lambda to consume the updated library.

After merging the change, my GitHub Actions CI tests [started to fail on Linux and Windows](https://github.com/martincostello/adventofcode/actions/runs/6531004538/attempts/1) with `OutOfMemoryException` errors.

This is interesting for several reasons:

1. The GitHub Actions CI isn't hosted in AWS Lambda, but does use [my test library](https://github.com/martincostello/lambda-test-server) to _simulate_ running in AWS Lambda.
2. I've not seen these exceptions before in my CI.
3. It didn't happen in the CI on the pull request itself.
4. It doesn't _seem_ to affect macOS.

The test library [sets the `AWS_LAMBDA_FUNCTION_MEMORY_SIZE` environment variable to simulate the Lambda runtime environment](https://github.com/search?q=repo%3Amartincostello%2Flambda-test-server%20AWS_LAMBDA_FUNCTION_MEMORY_SIZE&type=code), so it may be that this failure is "expected" based on the change in behaviour and my tests need to be configured to use a higher memory limit for the application's workload.

As the `net8.0` version explicitly changes the runtime's memory limits, that's the likely root cause culprit for the exceptions.

I'm logging this issue just in case there's maybe some missing runtime guards or something where the library _shouldn't_ adjust the memory limits other than in the explicit cases already commented in the code.

https://github.com/aws/aws-lambda-dotnet/blob/3dea2c0158f5943f34ede73a2a6a9303ae297087/Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/LambdaBootstrap.cs#L268-L279

/cc @normj

### Expected Behavior

`OutOfMemoryException` is not thrown.

### Current Behavior

`OutOfMemoryException` is thrown.

### Reproduction Steps

[...]

PR #1595, opened 2023-10-19
Add AWS_LAMBDA_DOTNET_DISABLE_MEMORY_LIMIT_CHECK env variable to turn…

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/1594

*Description of changes:*
Add the `AWS_LAMBDA_DOTNET_DISABLE_MEMORY_LIMIT_CHECK` environment variable to disable the code added for .NET 8 that configures the .NET runtime to use the amount of memory the Lambda function is configured for. Code to deployed to Lambda should never need this but for tools that are emulating Lambda environment including the memory check can cause problems for things like test runners running Lambda functions in process with other components.

@martincostello Please let me know if this PR would work for you.

### DIFF
--- a/Libraries/src/Amazon.Lambda.RuntimeSupport/Amazon.Lambda.RuntimeSupport.csproj
+++ b/Libraries/src/Amazon.Lambda.RuntimeSupport/Amazon.Lambda.RuntimeSupport.csproj
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net5.0;net6.0;net8.0</TargetFrameworks>
-    <VersionPrefix>1.9.0</VersionPrefix>
+    <VersionPrefix>1.9.1</VersionPrefix>
     <Description>Provides a bootstrap and Lambda Runtime API Client to help you to develop custom .NET Core Lambda Runtimes.</Description>
     <AssemblyTitle>Amazon.Lambda.RuntimeSupport</AssemblyTitle>
     <AssemblyName>Amazon.Lambda.RuntimeSupport</AssemblyName>
--- a/Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/Constants.cs
+++ b/Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/Constants.cs
@@ -24,6 +24,8 @@ namespace Amazon.Lambda.RuntimeSupport.Bootstrap
         // variable should never be set when function is deployed to Lambda.
         internal const string ENVIRONMENT_VARIABLE_AWS_LAMBDA_DOTNET_DEBUG_RUN_ONCE = "AWS_LAMBDA_DOTNET_DEBUG_RUN_ONCE";
 
+        internal const string ENVIRONMENT_VARIABLE_DISABLE_HEAP_MEMORY_LIMIT = "AWS_LAMBDA_DOTNET_DISABLE_MEMORY_LIMIT_CHECK";
+
         internal const string ENVIRONMENT_VARIABLE_AWS_LAMBDA_DOTNET_PREJIT = "AWS_LAMBDA_DOTNET_PREJIT";
         internal const string ENVIRONMENT_VARIABLE_AWS_LAMBDA_INITIALIZATION_TYPE = "AWS_LAMBDA_INITIALIZATION_TYPE";
         internal const string ENVIRONMENT_VARIABLE_LANG = "LANG";
--- a/Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/LambdaBootstrap.cs
+++ b/Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/LambdaBootstrap.cs
@@ -265,6 +265,15 @@ namespace Amazon.Lambda.RuntimeSupport
         {
             try
             {
+                // Check the environment variable to see if the user has opted out of RuntimeSupport adjusting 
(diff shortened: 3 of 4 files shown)


martincostello (approved):
This looks great.

martincostello:
One wrinkle I've noticed with this is that trying to write a test to show that "too much" memory can or can't be used, it seems that once the GC has the value given to it it's frozen at that value and can't be changed.

The API's semantics seem to suggest you can change it as much as you like, but it doesn't seem to behave that way and I can't find any documentation on that specific point either way. I've tried increasing and decreasing the value (and unsetting it) after it's been set once, but it seems to keep the first value.

I'm not sure if that's by design or not, but means that if intentional, users who want/need to use this setting will need to ensure it's set before the first invocation of the code that sets `GCHeapHardLimit`.

normj:
> One wrinkle I've noticed with this is that trying to write a test to show that "too much" memory can or can't be used, it seems that once the GC has the value given to it it's frozen at that value and can't be changed.

That is not what I'm seeing using the .NET API. I ran the following code to test it out:
```
AdjustMemory(200);
AdjustMemory(400);
AdjustMemory(100);


void AdjustMemory(int mb)
{
    const int OneMegabyte = 1024 * 1024;

    ulong memoryInBytes = (ulong)mb * OneMegabyte;

    AppContext.SetData("GCHeapHardLimit", memoryInBytes);
    GC.RefreshMemoryLimit();

    Console.WriteLine(GC.GetGCMemoryInfo().TotalAvailableMemoryBytes.ToString("N0"));
}
```

and this produced the following output showing the memory has changed.
```
209,715,200
419,430,400
104,857,600
```

Libraries/src/Amazon.Lambda.RuntimeSupport/Amazon.Lambda.RuntimeSupport.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net5.0;net6.0;net8.0</TargetFrameworks>
-    <VersionPrefix>1.9.0</VersionPrefix>
+    <VersionPrefix>1.9.1</VersionPrefix>
```
ashovlin, line 7: I was on the fence with "minor" - it's a new setting, but really offering a way to fallback to old behavior. Okay either way.

ashovlin (approved):

martincostello:
Interesting - what OS was that on?

normj:
> Interesting - what OS was that on?

That was on Windows but I just now tried it in Lambda to checkout Linux which worked just fine.

```
public class Function
{
    private static async Task Main(string[] args)
    {
        Action<string, ILambdaContext> handler = FunctionHandler;
        await LambdaBootstrapBuilder.Create(handler, new DefaultLambdaJsonSerializer())
            .Build()
            .RunAsync();
    }

    public static void FunctionHandler(string input, ILambdaContext context)
    {
        AdjustMemory(200, context);
        AdjustMemory(400, context);
        AdjustMemory(100, context);
    }

    static void AdjustMemory(int mb, ILambdaContext context)
    {
        const int OneMegabyte = 1024 * 1024;

        ulong memoryInBytes = (ulong)mb * OneMegabyte;

        AppContext.SetData("GCHeapHardLimit", memoryInBytes);
        GC.RefreshMemoryLimit();

        context.Logger.LogInformation(GC.GetGCMemoryInfo().TotalAvailableMemoryBytes.ToString("N0"));
    }
}
```

Log
```
START RequestId: ac03ada6-e53c-4c3c-ad13-14d498918b1d Version: $LATEST
2023-10-19T21:31:36.326Z	ac03ada6-e53c-4c3c-ad13-14d498918b1d	info	209,715,200
2023-10-19T21:31:36.406Z	ac03ada6-e53c-4c3c-ad13-14d498918b1d	info	419,430,400
2023-10-19T21:31:36.406Z	ac03ada6-e53c-4c3c-ad13-14d498918b1d	info	104,857,600
END RequestId: ac03ada6-e53c-4c3c-ad13-14d498918b1d
REPORT RequestId: ac03ada6-e53c-4c3c-ad13-14d498918b1d	Duration: 416.09 ms	Billed Duration: 854 ms	Memory Size: 256 MB	Max Memory Used: 60 MB	Init Duration: 437.16 ms	
```

Libraries/src/Amazon.Lambda.RuntimeSupport/Amazon.Lambda.RuntimeSupport.csproj
normj: I had similar thought but I didn't want to give the impression we added a feature since it is really just a way to turn off a feature.

normj:
@martincostello Make sure the type being passed into your call of AppContext.SetData is of type `ulong`. Otherwise the method will be quietly ignored.

martincostello:
Ahhh. That must be where I went wrong, I was just passing around `int` and `long` to reset the value. I'll try `ulong` when I'm back at my laptop tomorrow.

The only reason I asked the OS as I was getting a test that was failing only in CI for macOS earlier in Actions which was confusing me, but I've been doing all my local dev on Windows.

96malhar (approved):

normj:
@martincostello Can you give me one more confirmation this will work for you after you double check the wrinkle you have? If you are good then I'll ship it out.

martincostello:
@normj Yep, it was not using `ulong` that was causing my issues. With those changes my tests are behaving as expected if I build against your branch locally.

PR #1596, opened 2023-10-25
Add trimming support for RuntimeSupport, SystemTextJson serializer and Lambda event packages

normj:
*Description of changes:*
Updated the Lambda RuntimeSupport, SystemTextJson serializer and Lambda events to support trimming.

* Replaced all `NETCOREAPP_3_1` preprocessor references to `NETCOREAPP3_1_OR_GREATER`
* Added `net8.0` target framework to projects and mark the projects as trimmable.
* Marked all serializer in the SystemTextJson as requiring unreferenced code except for the source generator seriralizer
* There were a couple places where an event was directly using the SystemTextJson serializer. For .NET 8 I switched it to use the source generator approach.
* In Amazon.Lambda.RuntimeSupport marked the code paths used by the class library approach of loading the customer function by reflection and build Expressions as `RequiresUnreferencedCode`. These code paths are not used when deployed as an executable which is the AOT use case.
* Updated the Events.NET6 project to multi target .NET 8. At some point we should rename this project but I don't want to mess with the CI system right now.
* Added a new constructor for `SourceGeneratorLambdaJsonSerializer` that takes in the JsonSerializerContext so the serializer doesn't have to use reflection to get an instance of the generic parameter.

I have verified with these changes I can package a Lambda function using just these libraries and received no trim warnings.

Work not included in this PR
* Update the ASP.NET Core bridge to support trimming
* Update Lambda Annotations to support .NET 8 and trimming

### DIFF
--- a/Libraries/src/Amazon.Lambda.ConnectEvents/Amazon.Lambda.ConnectEvents.csproj
+++ b/Libraries/src/Amazon.Lambda.ConnectEvents/Amazon.Lambda.ConnectEvents.csproj
@@ -6,7 +6,7 @@
     <Description>Amazon Lambda .NET Core support - Amazon Connect package.</Description>
     <TargetFrameworks>netstandard2.0;net8.0</TargetFrameworks>
     <AssemblyTitle>Amazon.Lambda.ConnectEvents</AssemblyTitle>
-    <VersionPrefix>1.0.0</VersionPrefix>
+    <VersionPrefix>1.1.0</VersionPrefix>
     <AssemblyName>Amazon.Lambda.ConnectEvents</AssemblyName>
     <PackageId>Amazon.Lambda.ConnectEvents</PackageId>
     <PackageTags>AWS;Amazon;Lambda;Connect</PackageTags>
@@ -15,5 +15,6 @@
 	<PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
 		<WarningsAsErrors>IL2026,IL2067,IL2075</WarningsAsErrors>
 		<IsTrimmable>true</IsTrimmable>
+    <EnableTrimAnalyzer>true</EnableTrimAnalyzer>
 	</PropertyGroup>
 </Project>
--- a/Libraries/src/Amazon.Lambda.ConnectEvents/Amazon.Lambda.ConnectEvents.csproj
+++ b/Libraries/src/Amazon.Lambda.ConnectEvents/Amazon.Lambda.ConnectEvents.csproj
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <Description>Amazon Lambda .NET Core support - Amazon Connect package.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net8.0</TargetFrameworks>
     <AssemblyTitle>Amazon.Lambda.ConnectEvents</AssemblyTitle>
     <VersionPrefix>1.0.0</VersionPrefix>
     <AssemblyName>Amazon.Lambda.ConnectEvents</AssemblyName>
@@ -12,4 +12,8 @@
     <PackageTags>AWS;Amazon;Lambda;Connect</PackageTags>
   </PropertyGroup>
 
+	<PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
+		<WarningsAsErrors>IL2026,IL2067,IL2075</WarningsAsErrors>
+		<IsTrimmable>true</IsTrimmable>
+	</PropertyGroup>
 </Project>
--- a/Libraries/src/Amazon.Lambda.DynamoDBEvents/Converters/DictionaryLongToStringJsonConverter.cs
+++ b/Libraries/src/Amazon.Lambda.DynamoDBEvents/Converters/DictionaryLongToStringJsonConverter.cs
@@ -47,8 +47,12 @@ namespace Amazon.Lambda.DynamoDBEvents.Converters
 
         public override void Write(Utf8JsonWriter writer, Dictionary<string, string> value, JsonSerializerOptions options)
         {
+#if NET8_0_OR_GREATER
+            JsonSerializer.Serialize(writer, value, typeof(Dictionary<string, string>), new DictionaryStringStringJsonSerializerContext(options));
+#else
             // Use the built-in serializer, because it can handle dictionaries with string keys.
             JsonSerializer.Serialize(writer, value, options);
+#endif
         }
 
         private string ExtractValue(ref Utf8JsonReader reader, JsonSerializerOptions options)
@@ -67,5 +71,18 @@ namespace Amazon.Lambda.DynamoDBEvents.Converters
                     throw new JsonException($"'{reader.TokenType}' is not supported.");
             }
         }
+
+
     }
+
+#if NET8_0_OR_GREATER
+    /// <summary>
+    /// Context used for writing converter
+    /// </summary>
+    [JsonSerializable(typeof(Dictionary<string, string>))]
+    public partial class DictionaryStringStringJsonSerializerContext : JsonSerializerContext
+    {
+
+    }
+#endif
 }
--- a/Libraries/src/Amazon.Lambda.DynamoDBEvents/Converters/DictionaryLongToStringJsonConverter.cs
+++ b/Libraries/src/Amazon.Lambda.DynamoDBEvents/Converters/DictionaryLongToStringJsonConverter.cs
@@ -48,6 +48,7 @@ namespace Amazon.Lambda.DynamoDBEvents.Converters
         public override void Write(Utf8JsonWriter writer, Dictionary<string, string> value, JsonSerializerOptions options)
         {
 #if NET8_0_OR_GREATER
+            // For .NET 8+ use source generation for serialization to be trimming complaint
             JsonSerializer.Serialize(writer, value, typeof(Dictionary<string, string>), new DictionaryStringStringJsonSerializerContext(options));
 #else
             // Use the built-in serializer, because it can handle dictionaries with string keys.
--- a/Libraries/src/Amazon.Lambda.KinesisEvents/Converters/DictionaryLongToStringJsonConverter.cs
+++ b/Libraries/src/Amazon.Lambda.KinesisEvents/Converters/DictionaryLongToStringJsonConverter.cs
@@ -47,8 +47,12 @@ namespace Amazon.Lambda.KinesisEvents.Converters
 
         public override void Write(Utf8JsonWriter writer, Dictionary<string, string> value, JsonSerializerOptions options)
         {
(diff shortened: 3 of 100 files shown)


Libraries/src/Amazon.Lambda.RuntimeSupport/Client/InternalClientAdapted.cs
```diff
@@ -584,39 +584,8 @@ public async System.Threading.Tasks.Task<SwaggerResponse<StatusResponse>> ErrorW
             }
         }
 
-        private string ConvertToString(object value, System.Globalization.CultureInfo cultureInfo)
+        private string ConvertToString(string value, System.Globalization.CultureInfo cultureInfo)
         {
-            if (value is System.Enum)
```
normj, line 589: This code was not trimming compatible and it turn out our only use of the method was to the `value` being a string. So I choose to just simplify this method. Along time again this class was generated from a swagger file which I assume the generator just differentiated for all possible use cases.

Libraries/src/Amazon.Lambda.RuntimeSupport/ExceptionHandling/StackFrameInfo.cs
```diff
@@ -26,24 +27,40 @@ public StackFrameInfo(string path, int line, string label)
...
-            if (method != null)
+            try
             {
-                var methodTypeName = method.DeclaringType?.Name;
-                if (methodTypeName == null)
-                {
-                    Label = method.Name;
-                }
-                else
+                var method = stackFrame.GetMethod();
```
normj, line 41: This `GetMethod` call is marked as `RequiresUnreferencedCode`. The code is required so I just made the code more defensive in case something gets trimmed. Since we only really need the that is unlikely to be metadata trimmed.

Libraries/src/Amazon.Lambda.ConnectEvents/Amazon.Lambda.ConnectEvents.csproj
```diff
@@ -4,12 +4,16 @@
 
   <PropertyGroup>
     <Description>Amazon Lambda .NET Core support - Amazon Connect package.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net8.0</TargetFrameworks>
     <AssemblyTitle>Amazon.Lambda.ConnectEvents</AssemblyTitle>
     <VersionPrefix>1.0.0</VersionPrefix>
```
ashovlin, line 9: Missing version update.

Libraries/test/TestFunctionFSharp/TestFunctionCSharp/TestFunctionCSharp.csproj
```diff
@@ -14,7 +14,7 @@
     <PackageReference Include="xunit" Version="2.3.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
-    <PackageReference Include="FSharp.Core" Version="*" />
+   <!-- <PackageReference Include="FSharp.Core" Version="*" /> -->
```
ashovlin, line 17: nit: Is this meant to be removed? Just checking if it was from local testing and not meant to be committed?

ashovlin (changes requested):
And for the failed CI, can you add a .NET 8 block here?
https://github.com/aws/aws-lambda-dotnet/blob/68396f56cf8ad17c9715e23fe409017d5641f478/.github/workflows/source-generator-ci.yml#L17-L24

Libraries/src/Amazon.Lambda.RuntimeSupport/Amazon.Lambda.RuntimeSupport.csproj
```diff
@@ -18,6 +18,12 @@
   <PropertyGroup Condition=" '$(ExecutableOutputType)'=='true' ">
     <OutputType>Exe</OutputType>
   </PropertyGroup>
+	
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
+    <WarningsAsErrors>IL2026,IL2067,IL2075</WarningsAsErrors>
+    <IsTrimmable>true</IsTrimmable>
+	<EnableTrimAnalyzer>true</EnableTrimAnalyzer>
```
philasmar, line 25: Why does RuntimeSupport need `EnableTrimAnalyzer` but the rest don't?

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/LambdaBootstrap.cs
```diff
@@ -112,6 +112,11 @@ private LambdaBootstrap(HttpClient httpClient, LambdaBootstrapHandler handler, L
         /// </summary>
         /// <param name="cancellationToken"></param>
         /// <returns>A Task that represents the operation.</returns>
+#if NET8_0_OR_GREATER
+        [System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage("ReflectionAnalysis", "IL2026",
+            Justification = "Unreferenced code paths are excluding when RuntimeFeature.IsDynamicCodeSupported is false.")]
```
philasmar, line 117: "excluding" should be "excluded"

Libraries/src/Amazon.Lambda.ConnectEvents/Amazon.Lambda.ConnectEvents.csproj
normj: Fixed

Libraries/src/Amazon.Lambda.DynamoDBEvents/Converters/DictionaryLongToStringJsonConverter.cs
```diff
@@ -47,8 +47,12 @@ public class DictionaryLongToStringJsonConverter : JsonConverter<Dictionary<stri
 
         public override void Write(Utf8JsonWriter writer, Dictionary<string, string> value, JsonSerializerOptions options)
         {
+#if NET8_0_OR_GREATER
+            JsonSerializer.Serialize(writer, value, typeof(Dictionary<string, string>), new DictionaryStringStringJsonSerializerContext(options));
```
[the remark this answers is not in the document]
normj: Done

Libraries/src/Amazon.Lambda.KinesisEvents/Converters/DictionaryLongToStringJsonConverter.cs
```diff
@@ -47,8 +47,12 @@ public class DictionaryLongToStringJsonConverter : JsonConverter<Dictionary<stri
 
         public override void Write(Utf8JsonWriter writer, Dictionary<string, string> value, JsonSerializerOptions options)
         {
+#if NET8_0_OR_GREATER
+            JsonSerializer.Serialize(writer, value, typeof(Dictionary<string, string>), new DictionaryStringStringJsonSerializerContext(options));
```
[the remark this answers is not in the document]
normj: Done

Libraries/src/Amazon.Lambda.RuntimeSupport/Amazon.Lambda.RuntimeSupport.csproj
normj: Added `EnableTrimAnalyzer` to the other projects. I added this property so we can see future breaks to trimming at compile time for these libraries.

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/LambdaBootstrap.cs
normj: Fixed

Libraries/src/Amazon.Lambda.SNSEvents/Amazon.Lambda.SNSEvents.csproj
```diff
@@ -4,12 +4,16 @@
 
   <PropertyGroup>
     <Description>Amazon Lambda .NET Core support - SNSEvents package.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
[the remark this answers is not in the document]
normj: Fixed

Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/Converters/ConstantClassConverter.cs
```diff
@@ -41,10 +41,14 @@ public override bool CanConvert(Type typeToConvert)
         /// <param name="typeToConvert"></param>
         /// <param name="options"></param>
         /// <returns></returns>
+#if NET8_0_OR_GREATER
+        [System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage("ReflectionAnalysis", "IL2067",
+            Justification = "Constant classes are only used in the DynamoDB event referencing the SDK. S3 originally referenced the SDK but has been rewritten to no longer reference the SDK or ConstantClass. Suppressing this trim warning because we have marked the DynamoDB event as ")]
```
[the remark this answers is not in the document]
normj: Fixed

Libraries/test/TestFunctionFSharp/TestFunctionCSharp/TestFunctionCSharp.csproj
normj: Yes and removed now. It was constantly causing versioning problems this reference and the reference brought in by the `TestFunctionFSharp` dependency project. Found out I could remove the whole think in this project so it would pick up the reference from `TestFunctionFSharp` and no more versioning pain. It seemed to break in the past every time there was a new version of .NET.

normj:
> And for the failed CI, can you add a .NET 8 block here?
> 
> https://github.com/aws/aws-lambda-dotnet/blob/68396f56cf8ad17c9715e23fe409017d5641f478/.github/workflows/source-generator-ci.yml#L17-L24

Fixed the CI build. Since .NET 8 isn't GA I had to set the workflow to the current RC. Will need to update that to 8.0.x once .NET 8 has been released.

ashovlin (approved):

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/UserCodeInit.cs
```diff
@@ -140,6 +140,9 @@ public static void LoadStringCultureInfo()
             }
         }
 
+#if NET8_0_OR_GREATER
+    [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode("PreJitAssembly is not used for Native AOT")]
```
Beau-Gosse-dev, line 144: Technically someone could want to trim their assembly and not use native AOT, although obviously native AOT is the big use case.

PR #1610, opened 2023-11-14
Add trim support for Amazon.Lambda.Annotations

normj:
*Description of changes:*
Add trimming support for Lambda Annotations. Using trimming requires users to use the `SourceGeneratorLambdaJsonSerializer` lambda serializer so that JSON serialization is not being done via reflection.

The `HttpResults` class was modified to use the `JsonSerializerContext` that the `SourceGeneratorLambdaJsonSerializer` is registered with via the generic parameter. It uses the `JsonSerializerContext` to serialize the HTTP response body. It means the user has to register any types returned via HttpResults in the body with the  `JsonSerializerContext` via the `JsonSerializable` attribute.

Another slight change to `HttpResults` is it used to handle the serialization as soon as part of the constructor but there is no way to get the `JsonSerializerContext` into the constructor. So I changed body serialization to the when we serialize the `HttpResults`. This affected the unit tests because they were unintended reusing the same `IHttpResult` for each serialization format. So I change the test to create a new instance every time we run the serializer.


Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/HttpResultsTest.cs
```diff
@@ -14,9 +14,9 @@ public class HttpResultsTest
         [Fact]
         public void OkNoBody()
         {
-            var result = HttpResults.Ok();
+            var result = () => HttpResults.Ok();
```
normj, line 17: This is the change mention in the description where `ValidateResult` runs across the different combinations of API Gateway responses. The unit tests were change to pass in a `Func` instead of a single instance so we didn't reuse the same `IHttpResult` for each iteration in `ValidateResult`.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
```diff
@@ -162,9 +162,9 @@ public void Execute(GeneratorExecutionContext context)
                         }
                     }
                     
-                    var serializerString = GetSerializerAttribute(context, lambdaMethodModel);
+                    var serializerInfo = GetSerializerInfoAttribute(context, lambdaMethodModel);
```
normj, line 165: This was changed to `GetSerializerInfoAttribute` because now we are getting the optional `JsonSerializerContext` as well as the serializer name.

Libraries/test/TestExecutableServerlessApp/SourceGenerationSerializationExample.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+    }
+
+    [JsonSerializable(typeof(APIGatewayHttpApiV2ProxyRequest))]
+    [JsonSerializable(typeof(APIGatewayHttpApiV2ProxyResponse))]
+    [JsonSerializable(typeof(Person))]
+    public partial class HttpApiJsonSerializerContext : JsonSerializerContext
+    {
+
+    }
```
ashovlin, line 31: Could you add a section to the annotations-specific README discussing this?

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/LambdaSerializerInfo.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public class LambdaSerializerInfo
+    {
+        public LambdaSerializerInfo(string serializerName, string serializerJsonContextName)
+        {
+            SerializerName = serializerName;
+            SerializerJsonContextName = serializerJsonContextName;
+        }
+
+        public string SerializerName { get; }
```
ashovlin, line 15: nit: I like documenting even internal classes, including sample or expected values for strings. As a reviewer not too familiar with JSON+trimming, easier to see here than working backwards from call sites.

Libraries/test/TestExecutableServerlessApp/SourceGenerationSerializationExample.cs
ashovlin: Per internal discussion, planning a separate PR.

ashovlin (approved):

jeastham1993 (approved):
All looks good to me. Testing with .NET 8 and native AOT and it works perfectly.

Issue #1620, opened 2023-11-17
Install tzdb on provided.al2023 runtime so that TimeZoneInfo.FindSystemTimeZoneById() works


martincostello:
### Describe the feature

_Logging this here as it's .NET-specific, so if not implemented for `provided.al2023`, it might be something desired for any forthcoming .NET 8 managed runtime._

I migrated a .NET 8 application using the `provided.al2` runtime which uses the `TimeZoneInfo.FindSystemTimeZoneById()` method to convert UTC times to a specific time zone to `provided.al2023` earlier this week, and have discovered that compared to the previous version of the provided runtime, tzdb is not installed. Calling the `TimeZoneInfo.GetSystemTimeZones()` method returns 0 items as well.

This causes code paths using built-in .NET time zone handling APIs that typically work outside of the Lambda runtime environment (e.g. Linux build agents, developers' Windows machines) to fail when deployed to AWS Lambda.

To fix the functionality in the application, I've had to re-introduce a dependency on NodaTime (we used to use this, but removed it in favour of the new `TimeProvider` API).

```diff
+ using NodaTime;

var utcNow = TimeProvider.System.GetUtcNow();
- var timeZone = TimeZoneInfo.FindSystemTimeZoneById("Europe/London");
- var localNow = TimeZoneInfo.ConvertTimeFromUtc(utcNow, timeZone);
+ var timeZone = DateTimeZoneProviders.Tzdb["Europe/London"];
+ var localNow = Instant.FromDateTimeUtc(utcNow).InZone(timeZone).ToDateTimeUnspecified();
```

### Use Case

The application converts times in UTC to a specific time zone. For example:

```csharp
var utcNow = TimeProvider.System.GetUtcNow();
var timeZone = TimeZoneInfo.FindSystemTimeZoneById("Europe/London");
var nowLocal = TimeZoneInfo.ConvertTimeFromUtc(utcNow, timeZone);
```

With tzdb installed, this code will work correctly.

Without it, an exception is thrown at runtime:

```
System.TimeZoneNotFoundException: The time zone ID 'Europe/London' was not found on the local computer.
 ---> System.IO.DirectoryNotFoundException: Could not find a part of the path '/usr/share/zoneinfo/Europe/London'.
   at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirError)
[...]

PR #1636, opened 2023-12-08
Add smoke test to Lambda OCI images for TZData being available.

normj:
*Description of changes:*
Add a smoke test to our Lambda OCI publish process that confirms the availability of TZData. The `provided.al2023` did not initially have this as reported by issue https://github.com/aws/aws-lambda-dotnet/issues/1620 and has sense been added.


boblodgett (approved):

LambdaRuntimeDockerfiles/SmokeTests/test/ImageFunction.SmokeTests/ImageFunctionTests.cs
```diff
@@ -170,6 +171,7 @@ private async Task CreateFunctionAsync()
                         MemorySize = 512,
                         Role = _executionRoleArn,
                         PackageType = PackageType.Image,
+                        Timeout = 30,
```
normj, line 174: I added the timeout because the default value is 3 seconds and apparently listing all of the timezones at least for the first time can be slow.

ashovlin (approved):

philasmar (commented):
Could you please add some documentation for the new smoke test to indicate why we are looking for time zone data, and what it means if the test fails? If someone new is looking at a failed test, they probably wouldn't know why it's failing.

Issue #1639, opened 2023-12-12
LambdaTestTool - Support for Sync Invoke Waiting


huntharo:
### Describe the feature

Make `LambdaClient.InvokeAsync` calls work locally.

I have written a version of this but I need some guidance on how to fit this into the overall design here and to make sure I'm not breaking an intentionally existing feature.

### Use Case

Complex applications that mix `LambdaClient.InvokeAsync` with local actions cannot currently be tested locally with both the client application and the lambda itself attached to a debugger.  There are some ways around this using a deployed proxy Lambda, but that entire approach is orthogonal to LambdaTestTool's existence.

`LambdaClient.InvokeAsync` should be able to test blocking / sync invocations of Lambda functions, locally.

For complex projects this is an essential part of development, debugging, and testing.

### Proposed Solution

- The test event endpoint used by the web ui is also annotated with *almost* the right path to allow `InvokeAsync` calls from clients to work
  - https://github.com/aws/aws-lambda-dotnet/blob/f1c00fdc2b6c794e4dc222ca0143c0b74d107ffb/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Controllers/RuntimeApiController.cs#L22
  - `[HttpPost("/2015-03-31/functions/function/invocations")]`
  - When this endpoint is called though, it always immediately returns instead of waiting for the invoke to finish and relaying the response
  - This endpoint also has the hard-coded function name of `function`
- Add a new endpoint with annotation: `[HttpPost("/2015-03-31/functions/{functionName}/invocations")]`
  - Wait for a TaskCompletionSource to indicate that the response is ready
  - Relay the response or error back to the caller
- Related, particularly if debugging or testing invokes that last longer than 1 minute:
  - Allow configurability of the request timeouts

### Other Information

I'll submit a draft PR with my implementation that I've been using locally.

I'm looking for feedback on:
- Is the existing hard-coded `function` route usable?
  - Is it used for async invokes only?
  - If it is used, I suppose the new route would need to detect that the invoke type is not sync and operate as the current route does
- Guidance on usage of `Action` for completion callbacks
  - It does not seem there is much/any usage of `Action` in the tool
  - Is there an alternative preferred mechanism to use?

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

Built from source `master` branch

### Targeted .NET Platform

.NET 8

[...]

PR #1640, opened 2023-12-12
Issue 1639 - LambdaTestTool - Sync Invoke Endpoint

huntharo:
*Issue #, if available:* #1639

*Relates to:* #1349

*Description of changes:*
- Draft PR for feedback on approach
- Allows `AmazonLambdaClient.InvokeAsync` to work locally for testing complex projects that involve sync Invokes between a caller and Lambda
- Feedback needed:
  - Is the hard-coded function name `function` invoke endpoint in usage for async event testing?
    - If so, I assume I can determine the sync/async invoke type from the payload or context and only wait for sync
  - Use of `Action` for completion notifications - Is there a preferred alternative?
  - Thoughts on how to allow configuration of the two timeouts that were changed
    - For the immediately-returning endpoints this was not a problem, but it is a problem for sync invokes that block
    - Unfortunately it seems that Blazor does not allow changing the timeouts on a per-route basis
 - The test pages were throwing exceptions due to the `Response` not being available
    - I made a fix but I think there may be a more comprehensive change needed for the UI?
    - My solution was acceptable for my testing: it just showed the event as still running with no response
    - When the response was received the item moved into the finished events section and a new event was processed, so there did not appear to be any harm from my approach

I have personally been using this for a week now to build and test just such a project using .NET 8.0.

### DIFF
--- a/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Controllers/RuntimeApiController.cs
+++ b/Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Controllers/RuntimeApiController.cs
@@ -19,7 +19,6 @@ namespace Amazon.Lambda.TestTool.BlazorTester.Controllers
         }
 
         [HttpPost("/runtime/test-event")]
-        [HttpPost("/2015-03-31/functions/function/invocations")]
         public async Task<IActionResult> PostTestEvent()
         {
             using var reader = new StreamReader(Request.Body);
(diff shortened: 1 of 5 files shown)


ashishdhingra:
@huntharo Good afternoon. Please confirm if this PR is being actively worked upon or if it is good to close.

normj:
We are working on a new version of the test tool that does support this scenario. We just released the first early preview and admittedly our documentation is light and focused on Aspire integration. But the use case of using the SDK to invoke the Lambda function via the new test tool is supported. In fact part of the new tool is an API Gateway emulator that uses the SDK to invoke Lambda functions to the Lambda emulator side of the tool. https://github.com/aws/aws-lambda-dotnet/blob/master/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs#L159

I'm closing this PR since the focus is on our new tooling which will cover this use case.

huntharo:
Cool, thanks for the detailed update and feature add @normj !

PR #1237, opened 2022-07-08
remove S3 SDK deps

jcstorms1:
*Issue #, if available:*
#1226 
*Description of changes:*

- Updates `S3Event.cs` to handle all properties correctly
- Removes SDK.S3 dependency 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1648, opened 2024-01-09
Remove SDK dependency from DynamoDBEvent

dscpinheiro:
*Issue #, if available:* https://github.com/aws/aws-sdk-net/issues/1116 (and internal issue `DOTNET-4497`)

*Description of changes:* Similar to #1237, this PR removes the `AWSSDK.DynamoDBv2` dependency from the `DynamoDBEvent` class (by moving the properties from the `Amazon.DynamoDBv2.Model.Record` class).

### DIFF
--- a/Libraries/src/Amazon.Lambda.DynamoDBEvents/Amazon.Lambda.DynamoDBEvents.csproj
+++ b/Libraries/src/Amazon.Lambda.DynamoDBEvents/Amazon.Lambda.DynamoDBEvents.csproj
@@ -3,18 +3,14 @@
   <Import Project="..\..\..\buildtools\common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netcoreapp3.1;net8.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net8.0</TargetFrameworks>
     <Description>Amazon Lambda .NET Core support - DynamoDBEvents package.</Description>
     <AssemblyTitle>Amazon.Lambda.DynamoDBEvents</AssemblyTitle>
-    <VersionPrefix>2.3.0</VersionPrefix>
+    <VersionPrefix>3.0.0</VersionPrefix>
     <AssemblyName>Amazon.Lambda.DynamoDBEvents</AssemblyName>
     <PackageId>Amazon.Lambda.DynamoDBEvents</PackageId>
     <PackageTags>AWS;Amazon;Lambda</PackageTags>
   </PropertyGroup>
-
-  <ItemGroup>
-    <PackageReference Include="AWSSDK.DynamoDBv2" Version="3.7.3.24" />
-  </ItemGroup>
 
   <ItemGroup>
     <Compile Remove="Converters\DictionaryLongToStringJsonConverter.cs" />
@@ -23,10 +19,10 @@
   <ItemGroup Condition=" '$(TargetFramework)' != 'netstandard2.0' ">
     <Compile Include="Converters\DictionaryLongToStringJsonConverter.cs" />
   </ItemGroup>
-	
-	<PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
-		<WarningsAsErrors>IL2026,IL2067,IL2075</WarningsAsErrors>
-		<IsTrimmable>true</IsTrimmable>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
+    <WarningsAsErrors>IL2026,IL2067,IL2075</WarningsAsErrors>
+    <IsTrimmable>true</IsTrimmable>
     <EnableTrimAnalyzer>true</EnableTrimAnalyzer>
-	</PropertyGroup>
+  </PropertyGroup>
 </Project>
--- a/Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBEvent.cs
+++ b/Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBEvent.cs
@@ -226,7 +226,7 @@ namespace Amazon.Lambda.DynamoDBEvents
             ///  <c>"BOOL": true</c>
             /// </para>
             /// </summary>
-            public bool BOOL { get; set; }
+            public bool? BOOL { get; set; }
 
             /// <summary>
             /// <para>
@@ -304,7 +304,7 @@ namespace Amazon.Lambda.DynamoDBEvents
             ///  <c>"NULL": true</c> 
             /// </para>
             /// </summary>
-            public bool NULL { get; set; }
+            public bool? NULL { get; set; }
 
             /// <summary>
             /// <para>
--- a/Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBEvent.cs
+++ b/Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBEvent.cs
@@ -1,17 +1,14 @@
 namespace Amazon.Lambda.DynamoDBEvents
 {
-    using Amazon.DynamoDBv2.Model;
     using System;
     using System.Collections.Generic;
+    using System.IO;
 
     /// <summary>
     /// AWS DynamoDB event
     /// http://docs.aws.amazon.com/lambda/latest/dg/with-ddb.html
     /// http://docs.aws.amazon.com/lambda/latest/dg/eventsources.html#eventsources-ddb-update
     /// </summary>
-#if NET8_0_OR_GREATER
-    [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode("DynamoDBEvent has a reference to the AWS SDK for .NET. The ConstantClass used to represent enums in the SDK is not supported in the Lambda serializer SourceGeneratorLambdaJsonSerializer for trimming scenarios.")]
-#endif
     public class DynamoDBEvent
     {
         /// <summary>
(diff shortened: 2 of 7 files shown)


Libraries/src/Amazon.Lambda.DynamoDBEvents/Amazon.Lambda.DynamoDBEvents.csproj
```diff
@@ -3,30 +3,26 @@
   <Import Project="..\..\..\buildtools\common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netcoreapp3.1;net8.0</TargetFrameworks>
```
dscpinheiro, line 6: Removed `netstandard2.0` since it was removed from #1237 as well. Please let me know if I should revert this.

Libraries/test/EventsTests.Shared/EventTests.cs
```diff
@@ -482,6 +482,23 @@ public void DynamoDbUpdateTest(Type serializerType)
             var recordDateTime = record.Dynamodb.ApproximateCreationDateTime;
             Assert.Equal(recordDateTime.Ticks, 636162388200000000);
 
+            var topLevelList = record.Dynamodb.NewImage["misc1"].L;
```
dscpinheiro, line 485: In `Libraries/test/EventsTests.Shared/dynamodb-event.json`, I added properties for all possible `AttributeValue` (to confirm they could be deserialized correctly).

---

I also noticed the xUnit signature is `Equal<T>(T expected, T actual)`, but most of the existing tests instead check `Equal(actual, expected)`... I followed the correct format here, but did not change the existing assert statements.

Libraries/test/EventsTests.Shared/dynamodb-event.json
```diff
@@ -41,10 +54,14 @@
       },
       {
          "eventID":"f07f8ca4b0b26cb9c4e5e77e42f274ee",
-         "eventName":"INSERT",
```
dscpinheiro, line 44: Updated this record too to check for `UserIdentity` (it's only available in `REMOVE` events).

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBEvent.cs
```diff
@@ -21,15 +18,315 @@ public class DynamoDBEvent
...
+            /// <summary>
+            /// <para>
+            /// An attribute of type Boolean. For example:
+            /// </para>
+            ///  
+            /// <para>
+            ///  <c>"BOOL": true</c>
+            /// </para>
+            /// </summary>
+            public bool BOOL { get; set; }
```
normj, line 229: Use the nullable type. Here use `bool?`.

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBEvent.cs
```diff
@@ -21,15 +18,315 @@ public class DynamoDBEvent
...
+            /// <summary>
+            /// <para>
+            /// An attribute of type Null. For example:
+            /// </para>
+            ///  
+            /// <para>
+            ///  <c>"NULL": true</c> 
+            /// </para>
+            /// </summary>
+            public bool NULL { get; set; }
```
normj, line 307: Use the nullable type. Here use `bool?`.

Libraries/src/Amazon.Lambda.Serialization.Json/AwsResolver.cs
```diff
@@ -86,7 +86,7 @@ protected override IList<JsonProperty> CreateProperties(Type type, MemberSeriali
                     }
                 }
             }
-            else if (type.FullName.Equals("Amazon.DynamoDBv2.Model.StreamRecord", StringComparison.Ordinal))
+            else if (type.FullName.Equals("Amazon.Lambda.DynamoDBEvents.DynamoDBEvent+StreamRecord", StringComparison.Ordinal))
```
normj, line 89: Is there a reason we can't leave the old `else if`? That way if somebody updates `Amazon.Lambda.Serialization.Json` but not the event package they wouldn't get a break in behavior.

Libraries/src/Amazon.Lambda.Serialization.Json/AwsResolver.cs
```diff
@@ -96,7 +96,7 @@ protected override IList<JsonProperty> CreateProperties(Type type, MemberSeriali
                     }
                 }
             }
-            else if (type.FullName.Equals("Amazon.DynamoDBv2.Model.AttributeValue", StringComparison.Ordinal))
+            else if (type.FullName.Equals("Amazon.Lambda.DynamoDBEvents.DynamoDBEvent+AttributeValue", StringComparison.Ordinal))
```
normj, line 99: Is there a reason we can't leave the old `else if`? That way if somebody updates `Amazon.Lambda.Serialization.Json` but not the event package they wouldn't get a break in behavior.

Libraries/test/EventsTests.Shared/EventTests.cs
```diff
@@ -482,6 +482,23 @@ public void DynamoDbUpdateTest(Type serializerType)
...
+            Assert.Equal(0, topLevelList.Count);
+
+            var nestedMap = record.Dynamodb.NewImage["misc2"].M;
+            Assert.NotNull(nestedMap);
```
normj, line 489: Since we are doing this work to make properties not set in the event JSON are null on the POCO can you add some asserts to confirm they are null.

normj (changes requested):

Libraries/src/Amazon.Lambda.Serialization.Json/AwsResolver.cs
```diff
@@ -86,7 +86,7 @@ protected override IList<JsonProperty> CreateProperties(Type type, MemberSeriali
                     }
                 }
             }
-            else if (type.FullName.Equals("Amazon.DynamoDBv2.Model.StreamRecord", StringComparison.Ordinal))
```
normj, line 89: You will also need to bump the version number for this project in the csproj file.

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBEvent.cs
dscpinheiro: Updated in https://github.com/aws/aws-lambda-dotnet/pull/1648/commits/3bd10bc4900e4d21ebba4d47fe9700cd0076a74a#diff-03c240223072abf4e9dea6645203768f84845f84f54ea472b360aeff7d3960cfL229

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBEvent.cs
dscpinheiro: Updated in https://github.com/aws/aws-lambda-dotnet/commit/3bd10bc4900e4d21ebba4d47fe9700cd0076a74a#diff-03c240223072abf4e9dea6645203768f84845f84f54ea472b360aeff7d3960cfR307

Libraries/src/Amazon.Lambda.Serialization.Json/AwsResolver.cs
dscpinheiro: No, that's a good call out; I did not realize this was a separate package. I've updated this `if` to include both type names in: https://github.com/aws/aws-lambda-dotnet/commit/3bd10bc4900e4d21ebba4d47fe9700cd0076a74a#diff-ba736be4021bfb4dcd5cb24447e5ba3396f9bf3b18b6f32a2407101383c22e72R89

Libraries/src/Amazon.Lambda.Serialization.Json/AwsResolver.cs
dscpinheiro: Updated the version in https://github.com/aws/aws-lambda-dotnet/commit/3bd10bc4900e4d21ebba4d47fe9700cd0076a74a#diff-87dd4211c373a801da796dc71249823e727ca7e71d70f0a2f87bac50b1288244R12

I didn't do a major version bump there, please let me know if that was needed.

Libraries/test/EventsTests.Shared/EventTests.cs
dscpinheiro: Added a few asserts in https://github.com/aws/aws-lambda-dotnet/commit/3bd10bc4900e4d21ebba4d47fe9700cd0076a74a#diff-3eb0f1f2aaf46a4d65fa620dd99f96f1fba13a909a06a2f1e43c125cb6e1112e

normj (approved):

Libraries/src/Amazon.Lambda.DynamoDBEvents/DynamoDBEvent.cs
```diff
@@ -21,15 +18,315 @@ public class DynamoDBEvent
...
+            /// </li>
+            /// 
+            /// <li>
+            /// <para>
+            ///  <c>NEW_AND_OLD_IMAGES</c> - both the new and the old item images of the item.
+            /// </para>
+            /// </li>
+            /// </ul>
+            /// </summary>
+            public string StreamViewType { get; set; }
```
ashovlin, line 174: This is an enum in the SDK, should we try to reproduce here? Though I suppose a plain .NET `enum` isn't as flexible as the generated constant class we use.

https://github.com/aws/aws-sdk-net/blob/459d2e5410af5a5b0e26ba6951e75bf0c420351a/sdk/src/Services/DynamoDBv2/Generated/Model/StreamRecord.cs#L180

normj: We have to use `string` to be forward compatible with new enum values and we don't have the `ConstantClass` to use here. Which the `ConstantClass` cause other fun JSON serialization issues.

ashovlin (approved):

PR #1658, opened 2024-01-18
Add support for .NET 8 to Lambda Annotations

ashovlin:
*Issue #, if available:* DOTNET-7321

*Description of changes:* This adds support for the upcoming .NET 8 managed runtime to Lambda Annotations.
   * **Previously:** Since `dotnet6` was the only managed runtime when annotations launched we hardcoded `dotnet6` when generating the CloudFormation for a `zip` package type
   * **Now:** We support `dotnet8` in either the `LambdaGlobalProperties` attribute or by detecting it in the user's csproj.

### On Detection
1. From internal discussions, we didn't want to simply parse the .csproj's XML for `<TargetFramework>` since that leaves us susceptible to issues like https://github.com/aws/aws-extensions-for-dotnet-cli/issues/211
2. I could not find a way to detect the TFM from the context available to the source generator (`GeneratorExecutionContext`)
    * https://stackoverflow.com/questions/59843783/determine-target-framework-inside-a-visual-studio-code-analyzer-roslyn
3. I explored using `Micorosft.Build` and `Microsoft.Build.Locator`, but those don't target .NET Standard so I don't think we can use them from a source generator.
4. ~~Norm suggested the new `dotnet msbuild -getProperty` which is available with 17.8 (8.0.1)~~ _Edit: see https://github.com/aws/aws-lambda-dotnet/pull/1658/#issuecomment-1943056310_
    * https://learn.microsoft.com/en-us/visualstudio/msbuild/evaluate-items-and-properties?view=vs-2022


Libraries/Amazon.Lambda.Annotations.slnf
```diff
@@ -2,15 +2,16 @@
...
-      "src\\Amazon.Lambda.APIGatewayEvents\\Amazon.Lambda.APIGatewayEvents.csproj",
-      "src\\Amazon.Lambda.RuntimeSupport\\Amazon.Lambda.RuntimeSupport.csproj",
       "src\\Amazon.Lambda.Core\\Amazon.Lambda.Core.csproj",
+      "src\\Amazon.Lambda.RuntimeSupport\\Amazon.Lambda.RuntimeSupport.csproj",
       "src\\Amazon.Lambda.Serialization.SystemTextJson\\Amazon.Lambda.Serialization.SystemTextJson.csproj",
       "test\\Amazon.Lambda.Annotations.SourceGenerators.Tests\\Amazon.Lambda.Annotations.SourceGenerators.Tests.csproj",
-      "test\\TestServerlessApp.IntegrationTests\\TestServerlessApp.IntegrationTests.csproj",
       "test\\TestExecutableServerlessApp\\TestExecutableServerlessApp.csproj",
+      "test\\TestServerlessApp.IntegrationTests\\TestServerlessApp.IntegrationTests.csproj",
+      "test\\TestServerlessApp.NET8\\TestServerlessApp.NET8.csproj",
```
ashovlin, line 14: fyi: I was on adding line 14, by my VS sorted the list.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Diagnostics/AnalyzerReleases.Shipped.md
```diff
@@ -1,6 +1,16 @@
 ; Shipped analyzer releases
 ; https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
 
+## Release 1.1.0
```
ashovlin, line 4: Unrelated to this change, just moving the errors we added in 1.1.0 to the shipped file.

Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/Snapshots/Functions_ToUpper_Generated_NET8.g.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            var envValue = new StringBuilder();
+
+            // If there is an existing execution environment variable add the annotations package as a suffix.
+            if(!string.IsNullOrEmpty(Environment.GetEnvironmentVariable(envName)))
+            {
+                envValue.Append($"{Environment.GetEnvironmentVariable(envName)}_");
+            }
+
+            envValue.Append("amazon-lambda-annotations_1.1.0.0");
```
philasmar, line 40: Should we just increment the version to the one we intend to release this PR in? (and do it for all the other snaphsot files)

Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/WriterTests/ProjectFileHandlerTests.cs
```diff
@@ -96,5 +97,143 @@ public void IsNotOptedOut_PresentButFalse()
...
+        /// <summary>
+        /// Asserts that parsing a csproj with a multiple values in the plural TargetFrameworks property fails
+        /// </summary>
+        [Fact]
+        public void TryDetermineTargetFramework_Multiple_Failure()
+        {
+            var tempFile = Path.GetTempFileName();
+            var csprojContent = @"<Project Sdk=""Microsoft.NET.Sdk"">
+                                    <PropertyGroup>
+                                        <TargetFrameworks>net6.0;net8.0</TargetFrameworks>
```
philasmar, line 221: Is this a use case we'd want to support? Can't we just pick the minimum supported version?

philasmar (changes requested):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/ProjectFileHandler.cs
```diff
@@ -9,6 +11,14 @@ namespace Amazon.Lambda.Annotations.SourceGenerator
     /// </summary>
     public class ProjectFileHandler
     {
+        /// <summary>
+        /// Timeout for the `dotnet msbuild -getProperty` command we use to determine target framework
+        /// </summary>
+        private const int dotnetMsbuildTimeoutMs = 5000;
```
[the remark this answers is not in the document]
ashovlin: Fixed in https://github.com/aws/aws-lambda-dotnet/pull/1658/commits/1646e3a19d80516490a654c8ba9385ec1c58b5e5

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/ProjectFileHandler.cs
```diff
@@ -42,5 +52,95 @@ public static bool IsTelemetrySuppressed(string projectFilePath, IFileManager fi
...
+                return false;
+            }
+
+            // If there isn't the Properties key in the JSON, we failed to read values for either
+            if (!parsedJson.ContainsKey("Properties"))
+            {
+                return false;
+            }
+
+            // If <TargetFramework> (singular) is specified, that takes precendece over <TargetFrameworks> (plural)
```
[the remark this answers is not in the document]
ashovlin: Fixed in https://github.com/aws/aws-lambda-dotnet/pull/1658/commits/1646e3a19d80516490a654c8ba9385ec1c58b5e5

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/ProjectFileHandler.cs
```diff
@@ -42,5 +52,95 @@ public static bool IsTelemetrySuppressed(string projectFilePath, IFileManager fi
...
+
+                if (!string.IsNullOrEmpty(targetFramework))
+                {
+                    outTargetFramework = targetFramework;
+                    return true;
+                }
+            }
+            
+            // Otherwise fallback to <TargetFrameworks> (plural)
+            if (parsedJson["Properties"]["TargetFrameworks"] != null)
```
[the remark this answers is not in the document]
ashovlin: Dropped in your suggestion in https://github.com/aws/aws-lambda-dotnet/pull/1658/commits/1646e3a19d80516490a654c8ba9385ec1c58b5e5

Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/Snapshots/Functions_ToUpper_Generated_NET8.g.cs
ashovlin: Given the churn we've had with .NET 8, I didn't want to tie this to a version yet so would just do the separate versioning PR.

Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/WriterTests/ProjectFileHandlerTests.cs
ashovlin: For plural `TargetFrameworks`, I believe we rely on the user specifying the one they actually intend to publish via something like `dotnet lambda deploy-serverless --framework net8.0` or by `"framework": "net8.0"` in `aws-lambda-tools-defaults.json`. I didn't think we have access to that from the source generator, so didn't want to guess.

normj (approved):

normj (changes requested):
Can you update the following comment block to include `dotnet8`.

https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.Annotations/LambdaGlobalPropertiesAttribute.cs#L22

ashovlin:
Summarizing today's commits from internal discussion: I had seen [prior art](https://github.com/search?q=build_property.TargetFramework+language%3AC%23&type=code&l=C%23&p=2) for the following:
```
context.AnalyzerConfigOptions.GlobalOptions.TryGetValue("build_property.TargetFramework", out var targetFramework);
```

But when I initially tried it out, `targetFramework` was `null` so I gave up. However I was testing this through our `CSharpSourceGeneratorTest` which wasn't setting the TFM appropriately. When manually testing the full package in a test project `targetFramework` was set appropriately. 

https://github.com/aws/aws-lambda-dotnet/pull/1658/commits/e6ec2801f1d8968737d9ffa2bc3052d6ea085c53 now pivots to this approach and removes the `dotnet msbuild -getProperty` which should be more performant.

ashovlin:
> Can you update the following comment block to include `dotnet8`.
> 
> https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.Annotations/LambdaGlobalPropertiesAttribute.cs#L22

Fixed in https://github.com/aws/aws-lambda-dotnet/pull/1658/commits/a1aac5a120518bdc39a268aaf60b58a3748c8d50, thanks.

normj (approved):

philasmar (approved):

PR #1668, opened 2024-02-11
Add support for .NET 8 and Native AOT for ASP.NET Core bridge libraries.

normj:
*Description of changes:*
Update the ASP.NET Core bridge library to target .NET 8 and support Native AOT. For `Amazon.Lambda.AspNetCoreServer` I did a major version bump because I dropped support for .NET Core 3.1. .NET Core 3.1 has been out of support for quite sometime and Lambda has blocked function updates since May 2023. By dropping .NET Core 3.1 that simplified the logic especially using the trimming attributes which aren't available in .NET Core 3.1. I didn't major version `Amazon.Lambda.AspNetCoreServer.Hosting` because it never targeted .NET Core 3.1 and the changes done in `Amazon.Lambda.AspNerCoreServer` won't affect hosting.

The main trick for getting Native AOT working via Amazon.Lambda.AspNetCoreServer.Hosting was forcing the user to pass in the source generator serializer with a new overload of the `AddAWSLambdaHosting`. The existing one defaults to the reflection based source generator so it was marked as `RequiresUnreferencedCode`. To avoid duplicating logic I pushed a lot of the logic of AddAWSLambdaHosting into a utility method leaving the `AddAWSLambdaHosting` overrides to handle configuring the serializer. So the code got moved all around but it is mostly just to put in a separate method to avoid code duplication.

Along the way I did some clean up of test projects for example the RuntimeSupport unit tests were still targeting .NET Core 2.1 and updated them to .NET 8. 

## Testing
I have made sure all of the tests continue to work and I have build a ASP.NET Core app with these library for Native AOT and confirmed I got no trim warnings and the function worked correctly.

Some unit test cases were deleted because they tested deprecated behavior that has been removed. Specifically the `IWebHostBuilder` which isn't used anymore.


Libraries/src/Amazon.Lambda.AspNetCoreServer/AbstractAspNetCoreFunction.cs
```diff
@@ -206,60 +206,6 @@ public void RegisterResponseContentEncodingForContentEncoding(string contentEnco
         /// <param name="builder"></param>
         protected virtual void Init(IWebHostBuilder builder) { }
 
-        /// <summary>
-        /// Creates the IWebHostBuilder similar to WebHost.CreateDefaultBuilder but replacing the registration of the Kestrel web server with a 
-        /// registration for Lambda.
-        /// </summary>
-        /// <returns></returns>
-        [Obsolete("Functions should migrate to CreateHostBuilder and use IHostBuilder to setup their ASP.NET Core application. In a future major version update of this library support for IWebHostBuilder will be removed for non .NET Core 2.1 Lambda functions.")]
```
normj, line 214: This obsolete method was deleted because it was here to support .NET Core 2.1. The message even said we would delete in the future. The future is now.

Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/SourceGeneratorLambdaJsonSerializer.cs
```diff
@@ -95,6 +95,7 @@ public SourceGeneratorLambdaJsonSerializer(Action<JsonSerializerOptions> customi
         /// <param name="jsonSerializerContext"></param>
         /// <param name="customizer"></param>
         /// <param name="jsonWriterCustomizer"></param>
+        [Obsolete("The method is marked obsolete because the jsonSerializerContext parameter is unlikely to be created with the required JsonSerializerOptions for Lambda serialization. This will trigger confusing NullReferenceException.")]
```
normj, line 98: This overload was added last fall to pass in the instance of `JsonSerializerContext` to the `SourceGeneratorLambdaJsonSerializer`. It was done to avoid the reflection calls but the reflections used by `SourceGeneratorLambdaJsonSerializer` are annotated so they can be used in Native AOT. So this constructor wasn't strictly necessary. 

I'm marking as `Obsolete` because it turns out if you don't create the `JsonSerializerContext` with the correct json serialization options that `SourceGeneratorLambdaJsonSerializer` does like use case insensitive mapping the serialization fails at runtime with missing data. I lost a couple hours on this and I suspect others will too. To avoid this I think we should mark it as `Obsolete`.

Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.UnitTests/HandlerTests.cs
```diff
@@ -395,16 +393,20 @@ private async Task<ExecutionInfo> ExecHandlerAsync(string handler, string dataIn
                     Client = testRuntimeApiClient
                 };
 
+                var loggerAction = actionWriter.ToLoggingAction();
```
normj, line 396: This was moved down because now that the tests are targeting .NET 8 instead of .NET Core 2.1 it uses the newer log level logger. The problem with that is the `new LambdaBootstrap(handlerWrapper, initializer.InitializeAsync)` done a little bit above this overrides the test hack done by `UserCodeLoader.SetCustomerLoggerLogAction` to redirect the logging. I put this after `new LambdaBootstrap(handlerWrapper, initializer.InitializeAsync)` to override custom logger action after bootstrap does its work.

Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/SourceGeneratorLambdaJsonSerializer.cs
ashovlin: I'd add a resolution path for any users seeing this, maybe something like `You can now use the other constructors in a Native AOT deployment because the reflection calls have been annotated.` (assuming my understanding is correct)

ashovlin (approved):

Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/Utilities.cs
```diff
@@ -22,12 +23,21 @@ namespace Amazon.Lambda.AspNetCoreServer.Internal
     /// </summary>
     public static class Utilities
     {
+        /// <summary>
+        /// Method to make sure the Lambda implementation is the only registered implementaiton of IServer for ASP.NET Core runtime.
```
philasmar, line 27: nit: typo in "implementaiton"

Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/Utilities.cs
```diff
@@ -22,12 +23,21 @@ namespace Amazon.Lambda.AspNetCoreServer.Internal
...
+        /// </summary>
+        /// <param name="services"></param>
         public static void EnsureLambdaServerRegistered(IServiceCollection services)
         {
             EnsureLambdaServerRegistered(services, typeof(LambdaServer));
         }
 
-        public static void EnsureLambdaServerRegistered(IServiceCollection services, Type serverType)
+        /// <summary>
+        /// Method to make sure the Lambda implementation is the only registered implementaiton of IServer for ASP.NET Core runtime.
```
philasmar, line 36: nit: typo in "implementaiton"

philasmar (approved):

Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/Utilities.cs
normj: Fixed

Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/Utilities.cs
normj: Fixed

Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/SourceGeneratorLambdaJsonSerializer.cs
normj: Done

PR #1674, opened 2024-02-17
Update Lambda PowerShell to target .NET 8 runtime

normj:
*Description of changes:*
In preparation for the upcoming .NET 8 release update the `AWSLambdaPSCore` module for publishing PowerShell scripts to target .NET 8 and the upcoming .NET 8 Lambda runtime.

This follows the pattern we have done for previous runtime releases. The module has a major version update to v4. Users that take the v4 major version update will deploy to .NET 8. If users want to stay with .NET 6 then they need to keep using the v3.

### DIFF
--- a/Libraries/src/Amazon.Lambda.PowerShellHost/Amazon.Lambda.PowerShellHost.csproj
+++ b/Libraries/src/Amazon.Lambda.PowerShellHost/Amazon.Lambda.PowerShellHost.csproj
@@ -13,7 +13,7 @@
   </PropertyGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
-    <PackageReference Include="Microsoft.PowerShell.SDK" Version="7.2.1" />
+    <PackageReference Include="Microsoft.PowerShell.SDK" Version="7.2.18" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
     <PackageReference Include="Microsoft.PowerShell.SDK" Version="7.4.1" />
--- a/Libraries/src/Amazon.Lambda.PowerShellHost/Amazon.Lambda.PowerShellHost.csproj
+++ b/Libraries/src/Amazon.Lambda.PowerShellHost/Amazon.Lambda.PowerShellHost.csproj
@@ -3,20 +3,20 @@
   <Import Project="..\..\..\buildtools\common.props" />
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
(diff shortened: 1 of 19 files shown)


Libraries/src/Amazon.Lambda.PowerShellHost/Amazon.Lambda.PowerShellHost.csproj
```diff
@@ -3,20 +3,20 @@
...
+    <VersionPrefix>3.0.0</VersionPrefix>
     <AssemblyName>Amazon.Lambda.PowerShellHost</AssemblyName>
     <PackageId>Amazon.Lambda.PowerShellHost</PackageId>
     <PackageTags>AWS;Amazon;Lambda;PowerShell</PackageTags>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.1'">
-    <PackageReference Include="Microsoft.PowerShell.SDK" Version="6.0.4" />
+  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <PackageReference Include="Microsoft.PowerShell.SDK" Version="7.2.1" />
```
ashovlin, line 16: Just curious/don't need to see changes, but any reason for `7.2.1` over `7.2.18` (looks like the most recent .NET 6 release)?

ashovlin (approved):

Libraries/src/Amazon.Lambda.PowerShellHost/Amazon.Lambda.PowerShellHost.csproj
normj: Updated to version `7.2.18`.

Libraries/src/Amazon.Lambda.PowerShellHost/Amazon.Lambda.PowerShellHost.csproj
```diff
@@ -3,20 +3,20 @@
   <Import Project="..\..\..\buildtools\common.props" />
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFrameworks>net6.0;net8.0</TargetFrameworks>
```
afroz429, line 6: Is net6.0 target needed?

normj: Technically somebody could stick with the old major version of the module and have a "PowerShell Lambda .NET project" and update to this new major version of Amazon.Lambda.PowerShellHost. An unlikely scenario but since .NET 6 is still in support I'm not ready to rule it out.

Issue #1657, opened 2024-01-17
How to convert DynamoDBEvent images to DynamoDBv2.Document


dscpinheiro:
### Discussed in https://github.com/aws/aws-lambda-dotnet/discussions/1654

<div type='discussions-op-text'>

<sup>Originally posted by **Dreamescaper** January 16, 2024</sup>
I have a lambda subscribed to dynamoDb events. I want to convert event's records to regular JSON, currently we use something like that:
```csharp
public async Task FunctionHandler(DynamoDBEvent input, ILambdaContext context)
{
        foreach (var record in input.Records)
        {
                var image = record.Dynamodb.NewImage;
                var document = Document.FromAttributeMap(image);
                var json = document.ToJson();
                /*   ....    */
        }
}
```

However, it doesn't work after the https://github.com/aws/aws-lambda-dotnet/pull/1648 is merged.
Could anyone suggest what is the easiest way to do same after the update?</div>

adam-knights:
We've just hit this as well. We can no longer do

```
var document = Document.FromAttributeMap(attributeValues);
return _dynamoDBContext.FromDocument<MyObject>(document);
```

I can understand wanting to break the dependency but its weird to have AttributeValue duplicated as different classes across the two libraries, is the plan to have it live in a shared model library?

ashovlin:
We've released version [3.1.0 of Amazon.Lambda.DynamoDBEvents](https://www.nuget.org/packages/Amazon.Lambda.DynamoDBEvents/3.1.0), which adds a `ToJson` and `ToJsonPretty` that can be used with `DynamoDBEvent`.

We still kept the event definition separate from the SDK definition in AWSSDK.DynamoDBv2, which we split in version 3.0.0 via #1648. This avoids some interaction with code relevant to the SDK but not to Lambda (request marshallers), and may reduce the package size for cases where one only needs to read the event without using the full DynamoDB SDK.

You can now use the `ToJson` method on either `OldImage` or `NewImage` to:
1. Convert to JSON
2. Convert it to the SDK's `Document`
3. Convert it to the SDK's object persistence classes. 

Note that the JSON conversion has the same limitations as the SDK: the sets (SS, NS, BS) will be converted to JSON arrays, and binary (B) will be converted to Base64 strings 

```
foreach (var record in dynamoEvent.Records)
{
    // Convert the event to a JSON string
    var json = record.Dynamodb.NewImage.ToJson();

    // Which you can convert to the mid-level document model
[...]

PR #1685, opened 2024-02-28
Add ToJson for DynamoDBEvent

ashovlin:
*Issue #, if available:* #1654, #1657, DOTNET-7345

*Description of changes:* 
### Before
In #1648, we removed the dependency on `AWSSDK.DynamoDBv2` from `Amazon.Lambda.DynamoDBEvents`. This allowed us to address nullability issues related to the SDK's marshalling code, which is likely not relevant when working with a Lambda event. 

However, if you were working with the SDK's higher level document or object persistence representation of the DynamoDB event, that code may no longer work since the `AWSSDK.DynamoDBv2` dependency may no longer be present and we separated the previously shared types.

### After
Now, you can convert either the `OldImage` or `NewImage` on each DynamoDB event record to json via the new `ToJson` or `ToJsonPretty` methods. This can be used with `Document.FromJson` in `AWSSDK.DynamoDBv2` to initialize the higher-level programming models.

```
foreach (var record in dynamoEvent.Records)
{
    // Convert the event to a JSON string
    var json = record.Dynamodb.NewImage.ToJson();

    // Which you can convert to the mid-level document model
    var document = Document.FromJson(json);
    
    // And then to the high-level object model using an IDynamoDBContext
    var myClass = context.FromDocument<T>(document);
}
```

### DIFF
--- a/Libraries/src/Amazon.Lambda.DynamoDBEvents/ExtensionMethods.cs
+++ b/Libraries/src/Amazon.Lambda.DynamoDBEvents/ExtensionMethods.cs
@@ -1,0 +1,163 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Text.Json;
+using static Amazon.Lambda.DynamoDBEvents.DynamoDBEvent;
+
+namespace Amazon.Lambda.DynamoDBEvents
+{
+    /// <summary>
+    /// Extension methods for working with <see cref="DynamoDBEvent"/>
+    /// </summary>
+    public static class ExtensionMethods
+    {
+        /// <summary>
+        /// Converts a dictionary representing a DynamoDB item to a JSON string.
+        /// This may be useful when casting a DynamoDB Lambda event to the AWS SDK's 
+        /// higher-level document and object persistence classes.
+        /// </summary>
+        /// <remarks></remarks>
+        /// <param name="item">Dictionary representing a DynamoDB item</param>
+        /// <returns>Unformatted JSON string representing the DynamoDB item</returns>
+        public static string ToJson(this Dictionary<string, AttributeValue> item)
+        {
+            return ToJson(item, false);
+        }
+
+        /// <summary>
+        /// Converts a dictionary representing a DynamoDB item to a JSON string.
+        /// This may be useful when casting a DynamoDB Lambda event to the AWS SDK's 
+        /// higher-level document and object persistence classes.
+        /// </summary>
+        /// <remarks></remarks>
+        /// <param name="item">Dictionary representing a DynamoDB item</param>
+        /// <returns>Formatted JSON string representing the DynamoDB item</returns>
+        public static string ToJsonPretty(this Dictionary<string, AttributeValue> item)
+        {
+            return ToJson(item, true);
+        }
+
+        /// <summary>
+        /// Internal entry point for converting a dictionary representing a DynamoDB item to a JSON string.
+        /// </summary>
+        /// <param name="item">Dictionary representing a DynamoDB item</param>
+        /// <param name="prettyPrint">Whether the resulting JSON should be formatted</param>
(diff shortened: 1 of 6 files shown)


ashovlin:
CI failure is unrelated to this change, see https://github.com/aws/aws-extensions-for-dotnet-cli/issues/302

Libraries/src/Amazon.Lambda.DynamoDBEvents/ExtensionMethods.cs
```diff
@@ -0,0 +1,163 @@
...
+        /// <param name="prettyPrint">Whether the resulting JSON should be formatted</param>
+        /// <returns>JSON string representing the DynamoDB item</returns>
+        private static string ToJson(Dictionary<string, AttributeValue> item, bool prettyPrint)
+        {
+            if (item == null || item.Count == 0)
+            {
+                return "{}";
+            }
+
+            var stream = new MemoryStream();
```
dscpinheiro, line 53: Should we dispose the stream and writer here? I wasn't familiar with `Utf8JsonWriter` and saw that the [Microsoft documentation](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/use-utf8jsonwriter) uses:
```
using var stream = new MemoryStream();
using var writer = new Utf8JsonWriter(stream, options);
```

Libraries/src/Amazon.Lambda.DynamoDBEvents/ExtensionMethods.cs
```diff
@@ -0,0 +1,163 @@
...
+                    writer.WriteStringValue(item);
+                }
+                writer.WriteEndArray();
+            }
+            else if (attribute.NS != null)
+            {
+                writer.WriteStartArray();
+                foreach (var item in attribute.NS)
+                {
+#if NETCOREAPP3_1  // WriteRawValue was added in .NET 6, but we need to write out Number values without quotes
```
dscpinheiro, line 141: In other places where we check for .NET 6 or later we use: `#if NET6_0_OR_GREATER`. Is there a reason you didn't use the same pattern here?

dscpinheiro:
Were you planning to update the package version (https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.DynamoDBEvents/Amazon.Lambda.DynamoDBEvents.csproj#L9) later?

I also think it'd be good to update the blueprints to use the new version, but I guess we'd need to coordinate that with the toolkit team.

Thanks again for taking a look.

Libraries/src/Amazon.Lambda.DynamoDBEvents/ExtensionMethods.cs
ashovlin: My thinking was when we drop 3.1 support, we'd be more likely to search for `NETCOREAPP3_1` during cleanup.

Libraries/src/Amazon.Lambda.DynamoDBEvents/ExtensionMethods.cs
ashovlin: Added `using`s, thanks. Also converted the other two spots to the C# 8 style.

ashovlin:
> Were you planning to update the package version (https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.DynamoDBEvents/Amazon.Lambda.DynamoDBEvents.csproj#L9) later?
> 
> I also think it'd be good to update the blueprints to use the new version, but I guess we'd need to coordinate that with the toolkit team.
> 
> Thanks again for taking a look.

Added the version bump for Amazon.Lambda.DynamoDBEvents. I think we'd have to bump the blueprints in a separate PR anyway since those have a package reference that isn't in NuGet yet. I'll add a note to our internal ticket for after this ships.

dscpinheiro (approved):

normj (approved):

Issue #839, opened 2021-03-23
DynamoDB events have milliseconds, not seconds, in their timestamps which crashes deserialization


cdegroot:
## Description
I'm creating a lambda to process DynamoDB events streamed through Kinesis, which means they're wrapped in base64, etcetera. During processing I pull them out and then deserialize them to DynamoDBv2 records:

```
            var serializer = new Amazon.Lambda.Serialization.Json.JsonSerializer();
            var dynamoEvent = serializer.Deserialize<Amazon.DynamoDBv2.Model.Record>(BaseObserver.GenerateStreamFromString(dynamoData));
```

## Reproduction Steps

This gist: https://gist.github.com/cdegroot/6e8e1957d9d08a595b3918dce3e16d84. Note that all I did was to add a couple of zeros to the test event.

### Logs

Stacktrace: 

```
  X TestDynamoParsing [8ms]
  Error Message:
   Test method Canary.EventStore.Tests.BaseObserverTests.TestDynamoParsing threw exception: 
Amazon.Lambda.Serialization.Json.JsonSerializerException: Error converting the Lambda event JSON payload to type Amazon.DynamoDBv2.Model.Record: Value to add was out of range. (Parameter 'value') ---> System.ArgumentOutOfRangeException: Value to add was out of range. (Parameter 'value')                                                                                    
  Stack Trace:
      at System.DateTime.Add(Double value, Int32 scale)
   at System.DateTime.AddSeconds(Double value)
   at Amazon.Lambda.Serialization.Json.JsonNumberToDateTimeDataConverter.ReadJson(JsonReader reader, Type objectType, Object existingValue, JsonSerializer serializer)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.DeserializeConvertable(JsonConverter converter, JsonReader reader, Type objectType, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.SetPropertyValue(JsonProperty property, JsonConverter propertyConverter, JsonContainerContract containerContract, JsonProperty containerProperty, JsonReader reader, Object target)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateObject(Object newObject, JsonReader reader, JsonObjectContract contract, JsonProperty member, String id)
[...]

PR #1710, opened 2024-03-29
Handle when Lambda events are sent unix epoch dates in milliseconds

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/839

*Description of changes:*
When Kinesis streams are used to handle DynamoDB events they send the approximate date a unix epoch in milliseconds. When coming from a DynamoDB stream event they are seconds which is the normal format.

Since there isn't a way to detect the source of the event the serializers are updated to look to see if the `long` in the event would be greater then the number of seconds for the year `5000`. If so then assume the value is in milliseconds. The other alternative I could have done is try to create the `DateTime` with the millisecond epoch and if I got an out of range exception fallback to creating the DateTime from milliseconds but I wanted to avoid control flow by exceptions.

### DIFF
--- a/Libraries/src/Amazon.Lambda.Serialization.Json/JsonNumberToDateTimeDataConverter.cs
+++ b/Libraries/src/Amazon.Lambda.Serialization.Json/JsonNumberToDateTimeDataConverter.cs
@@ -12,6 +12,7 @@ namespace Amazon.Lambda.Serialization.Json
     /// </summary>
     internal class JsonNumberToDateTimeDataConverter : JsonConverter
     {
+        private const long YEAR_5000_IN_SECONDS = 157753180800;
         private static readonly TypeInfo DATETIME_TYPEINFO = typeof(DateTime).GetTypeInfo();
         private static readonly DateTime EPOCH_DATETIME = new DateTime(1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc);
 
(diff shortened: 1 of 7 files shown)


Libraries/src/Amazon.Lambda.Serialization.Json/JsonNumberToDateTimeDataConverter.cs
```diff
@@ -12,6 +12,7 @@ namespace Amazon.Lambda.Serialization.Json
     /// </summary>
     internal class JsonNumberToDateTimeDataConverter : JsonConverter
     {
+        private const long YEAR_5000_IN_SECONDS = 157753180800;
```
ashovlin, line 15: When checking the math at first I assumed this was seconds-since-epoch, but appears to be since year 1. https://dotnetfiddle.net/0VesSV

Not a big deal functionally though, same principle applies.

ashovlin (approved):

Libraries/src/Amazon.Lambda.Serialization.Json/JsonNumberToDateTimeDataConverter.cs
normj: Added a comment to make it clearer.

ashishdhingra (approved):

PR #1723, opened 2024-04-10
chore: fix lambda images workflow

philasmar:
*Description of changes:*
* The workflow currently fails to create a PR. This change fixes the issue where the BRANCH variable was not properly set.
* Remove warnings related to ::set-output and Node16
* Change base for checkout and PR to `dev` branch instead of `main`. Workflow should always be triggered from `dev` branch.

### DIFF
--- a/.github/workflows/update-Dockerfiles.yml
+++ b/.github/workflows/update-Dockerfiles.yml
@@ -61,7 +61,9 @@ jobs:
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
+        with:
+          ref: 'dev'
 
       - name: Update .NET 6 AMD64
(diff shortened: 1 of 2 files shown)


.github/workflows/update-Dockerfiles.yml
```diff
@@ -127,7 +129,7 @@ jobs:
         uses: repo-sync/pull-request@v2
         with:
           source_branch: ${{ steps.commit-push.outputs.BRANCH }}
-          destination_branch: "master"
+          destination_branch: "dev"
```
ashovlin, line 132: Checking my understanding, this PR will now target `dev`, so when releasing the images we'll have to do the `dev` -> `master` sync that we don't today? But we'll minimize those master-only commits that skip dev?

philasmar: Correct. I am trying to force everything to go through `dev` first. This will align with our future efforts to automate dev->main syncs.

ashovlin (approved):

96malhar (approved):

PR #1730, opened 2024-04-15
[UA update] chore: Update User-Agent string for Lambda RuntimeSupport

96malhar:
*Issue #, if available:*
SDK-169882

*Description of changes:*
chore: Update User-Agent string


normj (changes requested):
We don't want change the user agent for RuntimeSupport.

Issue #1728, opened 2024-04-12
Upgrade of Amazon.Lambda.Serialization.SystemTextJson from 2.2.0 to 2.4.2 breaks deserialization of base64 to byte[]


brendonparker:
### Describe the bug

In upgrading from dotnet6 to dotnet8 I upgraded the Amazon.Lambda.Serialization.SystemTextJson from `2.2.0` to `2.4.2`

After the upgrade, my lambdas which deserialized payloads that were deserializing base64 strings to byte[] started to fail.

### Expected Behavior

Should be able to deserialize base64 string as a byte[]. Or at least document on how to configure around this. No breaking changes.

### Current Behavior

```
Amazon.Lambda.Serialization.SystemTextJson.JsonSerializerException: Error converting the Lambda event JSON payload to type Company.Product.CloudWatchLogEventWrapper: The JSON value could not be converted to byte[].
 ---> System.Text.Json.JsonException: The JSON value could not be converted to byte[].
   at Amazon.Lambda.Serialization.SystemTextJson.Converters.ByteArrayConverter.Read(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options)
   at System.Text.Json.Serialization.JsonConverter`1.TryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value, Boolean& isPopulatedValue)
   at System.Text.Json.Serialization.Metadata.JsonPropertyInfo`1.ReadJsonAndSetMember(Object obj, ReadStack& state, Utf8JsonReader& reader)
   at System.Text.Json.Serialization.Converters.ObjectDefaultConverter`1.OnTryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value)
   at System.Text.Json.Serialization.JsonConverter`1.TryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value, Boolean& isPopulatedValue)
   at System.Text.Json.Serialization.Metadata.JsonPropertyInfo`1.ReadJsonAndSetMember(Object obj, ReadStack& state, Utf8JsonReader& reader)
   at System.Text.Json.Serialization.Converters.ObjectDefaultConverter`1.OnTryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value)
[...]

PR #1736, opened 2024-04-20
1728 - Enhance/Fix ByteArrayConverter to support base64 deserialization of byte[]

brendonparker:
*Issue #, if available:* [1728](https://github.com/aws/aws-lambda-dotnet/issues/1728)

*Description of changes:*
Add support for deserializing a base64 `string` to a `byte[]`.

If the token is a string, assume it is base64 encoded string and get bytes.

Otherwise, continue to do what it did before.

I added xunit tests for both conditions. To ensure my desired behavior works, and the existing behavior also works. I'm not strong in xunit, so if there are tweaks needed, please advise. Also wasn't sure where the best home/project for those tests might be.

### DIFF
--- a/Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/Converters/ByteArrayConverter.cs
+++ b/Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/Converters/ByteArrayConverter.cs
@@ -18,6 +18,11 @@ namespace Amazon.Lambda.Serialization.SystemTextJson.Converters
                 return null;
             }
 
+            if (reader.TokenType == JsonTokenType.String)
+            {
+                return reader.GetBytesFromBase64();
+            }
+
             var byteList = new List<byte>();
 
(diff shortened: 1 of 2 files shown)


brendonparker:
One callout: serialization will still convert a byte[] to an array of numbers (ints) in json. No good way of supporting both write formats in one converter. 

But in this approach, at least reading/deserializing can be flexible.

normj (approved):
Thanks for the fix. I agree there isn't really anything we can do for writes without some pre knowledge of what the format is supposed to be.

If you want bump the version in the project file to `2.4.3` that will save us from creating another version bump BR. https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/Amazon.Lambda.Serialization.SystemTextJson.csproj#L12C10-L12C23

ashishdhingra (approved):

PR #1801, opened 2024-09-02
Adding ASP.NET Core IServer hosting for Websocket APIs

wiowou:
*Issue #, if available:*

*Description of changes:*
API Gateway Websocket APIs currently error out when using this library to handle requests since the `RequestContext.Http` object is null. This change allows this library to be used with Websocket APIs by setting the `Http.Path` to the `RouteKey` and the `Http.Method` fixed as `POST`. This approach requires the user to reference the same lambda for the `RouteKey`s which they want handled by the lambda server.

The changes to the `APIGatewayHttpApiV2ProxyFunction` class is minimal and adds a `APIGatewayWebsocketApiV2ProxyFunction` class and corresponding changes to the `Amazon.Lambda.AspNetCoreServer.Hosting` project to provide for a Websocket option.

All tests that were passing in the current master branch are passing in the PR.

### DIFF
--- a/Libraries/src/Amazon.Lambda.AspNetCoreServer.Hosting/Internal/LambdaRuntimeSupportServer.cs
+++ b/Libraries/src/Amazon.Lambda.AspNetCoreServer.Hosting/Internal/LambdaRuntimeSupportServer.cs
@@ -95,6 +95,47 @@ namespace Amazon.Lambda.AspNetCoreServer.Hosting.Internal
     }
 
     /// <summary>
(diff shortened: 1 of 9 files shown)


wiowou:
I plan to add tests, shortly.

wiowou:
I think I should have derived from `APIGatewayProxyFunction` and not `APIGatewayHttpApiV2ProxyFunction`. I will rework this.

wiowou:
I will open a different PR.

Issue #1778, opened 2024-07-18
Add validation during build time for Handler max length


matthewyost:
### Describe the feature

Include validation on the LambdaFunction attribute that notifies developer that the generated function handler is over the maximum character length (128) so that it can be fixed at build time.

### Use Case

When using the Amazon.Lambda.Annotations you can hit the limit for max length of a FunctionHandler string in Cloudformation and only recognize this after attempting to deploy it

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

Amazon.Lambda.Annotations version 1.5.0.0

### Targeted .NET Platform

.NET 6/8

### Operating System and version

All

bhoradc:
Hi @matthewyost,

Thanks for submitting the feature request. I will review it with the team. 

Regards, 
Chaitanya

ashishdhingra:
Also refer https://github.com/aws/aws-lambda-dotnet/issues/1642 where the last character is stripped from the handler name. It's an open issue with the Lambda runtime team and they might plan to fix it in near future. So, currently, handler name of `127` would be supported actually due to issue in runtime.

matthewyost:
While I understand that there is a boundary concern here (still valid), the request is more for the DevX around finding this in the IDE before we execute it in CloudFormation.  I spent a lot of time trying to determine exactly why this was happening only to figure out that it was being cause by the SourceGenerator inside the Annotations library.  

If we had the ability to have this lintable, that would be ideal.

ashishdhingra:
> While I understand that there is a boundary concern here (still valid), the request is more for the DevX around finding this in the IDE before we execute it in CloudFormation. I spent a lot of time trying to determine exactly why this was happening only to figure out that it was being cause by the SourceGenerator inside the Annotations library.
> 
> If we had the ability to have this lintable, that would be ideal.

@matthewyost Thanks for your reply. I added https://github.com/aws/aws-lambda-dotnet/issues/1778#issuecomment-2237014066 just for reference. It would still be a good idea to add some validation.

96malhar:
Hello! We have added a new validator that checks and imposes a character limit on the Lambda function handler. See PR #1812. This change was released in [Amazon.Lambda.Annotations v1.5.1](https://www.nuget.org/packages/Amazon.Lambda.Annotations/1.5.1)

Issue #1642, opened 2023-12-13
Documented handler max length of 128 characters isn't supported


SeijiSuenaga:
### Describe the bug

AWS documentation says the `Handler` property has a [max length 128 characters](https://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html#SSS-CreateFunction-request-Handler), but values exactly 128 characters long appear to be truncated to 127 characters in the `dotnet6` Lambda runtime, and therefore fail.

### Expected Behavior

A handler value exactly 128 characters long shouldn't be truncated and should invoke the handler method successfully.

### Current Behavior

A handler value exactly 128 characters long produces a failure:

```log
Amazon.Lambda.RuntimeSupport.ExceptionHandling.LambdaValidationException: Unable to find method 'HandleInvocationAsyn' in type 'An.Example.Namespace.For.The.GitHub.Issue.ExampleLambdaFunction' from assembly 'An.Example.Namespace.For.The.GitHub.Issue': Found no methods matching method name 'HandleInvocationAsyn'.
   at Amazon.Lambda.RuntimeSupport.Bootstrap.UserCodeLoader.FindCustomerMethod(Type type) in /src/Repo/Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/UserCodeLoader.cs:line 242
   at Amazon.Lambda.RuntimeSupport.Bootstrap.UserCodeLoader.Init(Action`1 customerLoggingAction) in /src/Repo/Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/UserCodeLoader.cs:line 116
   at Amazon.Lambda.RuntimeSupport.Bootstrap.UserCodeInitializer.InitializeAsync() in /src/Repo/Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/UserCodeInitializer.cs:line 46
   at Amazon.Lambda.RuntimeSupport.LambdaBootstrap.InitializeAsync() in /src/Repo/Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/LambdaBootstrap.cs:line 155
```

### Reproduction Steps

Create a Lambda with a handler specification that is exactly 128 characters long, such as:

`An.Example.Namespace.For.The.GitHub.Issue::An.Example.Namespace.For.The.GitHub.Issue.ExampleLambdaFunction.HandleInvocationAsync`

### Additional Information/Context

[...]

PR #1812, opened 2024-09-19
fix: Impose a 127 character limit on the Lambda function handler when the package type is set to zip

96malhar:
*Issue #, if available:*
* https://github.com/aws/aws-lambda-dotnet/issues/1778
* DOTNET-7659

*Description of changes:*
This PR imposes a 127 character limit on the Lambda function handler when the package type is set to `zip`.
The official AWS docs state a 128 character limit on the Lambda handler. However, there is an [open issue](https://github.com/aws/aws-lambda-dotnet/issues/1642) where the last character is stripped off when the handler is exactly 128 characters long. Hence, we are enforcing a 127 character limit.

**Note**: The character limit does not apply when the package type is set to `Image`. As part of our integration tests, we have several Lambda functions with the handler names exceeding 127 characters and are deployed sucessfully with an Image based deployment. [Example 1](https://github.com/aws/aws-lambda-dotnet/blob/e24c52fa59db72bc83824020d267321a81410c07/Libraries/test/TestServerlessApp/serverless.template#L243) and [Example 2](https://github.com/aws/aws-lambda-dotnet/blob/e24c52fa59db72bc83824020d267321a81410c07/Libraries/test/TestServerlessApp/serverless.template#L866)

### DIFF
--- a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Diagnostics/DiagnosticDescriptors.cs
+++ b/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Diagnostics/DiagnosticDescriptors.cs
@@ -141,7 +141,7 @@ namespace Amazon.Lambda.Annotations.SourceGenerator.Diagnostics
 
         public static readonly DiagnosticDescriptor MaximumHandlerLengthExceeded = new DiagnosticDescriptor(id: "AWSLambda0118",
             title: "Maximum Handler Length Exceeded",
-            messageFormat: "The handler string '{0}' exceeds the maximum length of 127 characters. Please trim down your project namespace to stay within the character limit.",
+            messageFormat: "The handler string '{0}' exceeds the maximum length of 127 characters. Please trim down your project namespace, class name or method name to stay within the character limit.",
             category: "AWSLambdaCSharpGenerator",
             DiagnosticSeverity.Error,
             isEnabledByDefault: true);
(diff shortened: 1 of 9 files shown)


Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -138,5 +138,12 @@ public static class DiagnosticDescriptors
             category: "AWSLambdaCSharpGenerator",
             DiagnosticSeverity.Error,
             isEnabledByDefault: true);
+
+        public static readonly DiagnosticDescriptor MaximumHandlerLengthExceeded = new DiagnosticDescriptor(id: "AWSLambda0118",
+            title: "Maximum Handler Length Exceeded",
+            messageFormat: "The handler string '{0}' exceeds the maximum length of 127 characters. Please trim down your project namespace to stay within the character limit.",
```
normj, line 144: Shouldn't this be "Please trim down your project namespace, class name or method name to ..."?

normj (changes requested):

philasmar (approved):
Approved, assuming you address Norm's comment.

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Diagnostics/DiagnosticDescriptors.cs
96malhar: Done

normj (approved):

Issue #1641, opened 2023-12-12
Lambda PowerShell Functions Do Not Log Debug Messages to CloudWatch Messages 


ashovlin:
### Describe the bug

_We already have internal case and ticket V1078993289 for this, but creating this so others can track a fix publicly._

### Expected Behavior

"Write-Debug" should print to CloudWatch messages when adding -Debug parameter or using $DebugPreference to get the message displayed on console.

### Current Behavior

Messages aren't logged.

### Possible Solution

This probably is happening because "Write-Debug" isn't included in these events https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.PowerShellHost/PowerShellFunctionHost.cs#L294

ashishdhingra:
The change has been released in [AWSLambdaPSCore 4.0.2.0](https://www.powershellgallery.com/packages/AWSLambdaPSCore/4.0.2.0).

Kindly note the following to enable debugging:
- Refer [$DebugPreference](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.4#debugpreference). In PowerShell code, you:
  - Either need to set `$DebugPreference = "Continue"` at the beginning of script; **OR**
  - Or include `-Debug` parameter while executing `Write-Debug` (e.g. `Write-Debug "Testing Lambda PowerShell Write-Debug" -Debug`)
- Also set environment variable `AWS_LAMBDA_HANDLER_LOG_LEVEL` with value `DEBUG` (refer blog post [Introducing the .NET 6 runtime for AWS Lambda](https://aws.amazon.com/blogs/compute/introducing-the-net-6-runtime-for-aws-lambda/)). This will enable Debug logs at Lambda level.

PR #1814, opened 2024-09-19
Fixes an issue where Write-Debug doesn't log debug messages to CloudWatch for a Lambda PowerShell function.

ashishdhingra:
*Issue #, if available:* #1641

*Description of changes:*
Fixes an issue where `Write-Debug` doesn't log debug messages to CloudWatch for a Lambda PowerShell function. Refer [PSDataStreams](https://learn.microsoft.com/en-us/dotnet/api/system.management.automation.psdatastreams?view=powershellsdk-7.4.0) for more information about various buffers.

___

### DIFF
--- a/Libraries/src/Amazon.Lambda.PowerShellHost/PowerShellFunctionHost.cs
+++ b/Libraries/src/Amazon.Lambda.PowerShellHost/PowerShellFunctionHost.cs
@@ -148,6 +148,9 @@ namespace Amazon.Lambda.PowerShellHost
             // Clear all previous PowerShell executions, variables and outputs
             this._ps.Commands?.Clear();
             this._ps.Streams.Verbose?.Clear();
+            this._ps.Streams.Debug?.Clear();
+            this._ps.Streams.Information?.Clear();
+            this._ps.Streams.Warning?.Clear();
             this._ps.Streams.Error?.Clear();
             this._ps.Runspace?.ResetRunspaceState();
             this._output.Clear();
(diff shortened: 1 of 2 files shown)


Libraries/src/Amazon.Lambda.PowerShellHost/PowerShellFunctionHost.cs
```diff
@@ -149,6 +149,7 @@ private IAsyncResult BeginInvoke(string input, ILambdaContext context)
             this._ps.Commands?.Clear();
             this._ps.Streams.Verbose?.Clear();
             this._ps.Streams.Error?.Clear();
+            this._ps.Streams.Debug?.Clear();
```
normj, line 152: Looks like we are missing clearing `Information` and `Warning` here. Can we add those?

Libraries/src/Amazon.Lambda.PowerShellHost/PowerShellFunctionHost.cs
```diff
@@ -295,6 +296,7 @@ private void SetupStreamHandlers()
             this._ps.Streams.Information.DataAdding += _loggerFactory("Information");
             this._ps.Streams.Warning.DataAdding += _loggerFactory("Warning");
             this._ps.Streams.Error.DataAdding += _loggerFactory("Error");
+            this._ps.Streams.Debug.DataAdding += _loggerFactory("Debug");
```
normj, line 299: nit: These lines are in order or log level. Can you put `Debug` in between `Verbose` and `Information`? Also the block above where you are clearing can we put those clears in the same order. I think it is easier for readers.

normj (changes requested):

Libraries/src/Amazon.Lambda.PowerShellHost/PowerShellFunctionHost.cs
ashishdhingra: Added

Libraries/src/Amazon.Lambda.PowerShellHost/PowerShellFunctionHost.cs
ashishdhingra: Done

dscpinheiro (approved):

normj (approved):

PR #1848, opened 2024-10-25
chore: replace Blueprints nuspec file with csproj file

philasmar:
*Description of changes:*
Remove nuspec files used for Blueprints and added csproj files that can be built using `dotnet build/pack` instead of relying on nuget.exe


normj (changes requested):
Can you the `package-blueprints` target in the build tools? I don't think it is used in the pipeline but I use this targets in this file to package and test the blueprints. https://github.com/aws/aws-lambda-dotnet/blob/master/buildtools/build.proj#L40

philasmar:
> Can you the `package-blueprints` target in the build tools? I don't think it is used in the pipeline but I use this targets in this file to package and test the blueprints. https://github.com/aws/aws-lambda-dotnet/blob/master/buildtools/build.proj#L40

Updated

normj (approved):

96malhar (approved):

Issue #1197, opened 2022-06-09
Can't invoke a Lambda function that takes an int as a parameter with `dotnet lambda invoke-function`


bjhogan:
### Describe the bug

There doesn't seem to be a way to invoke a Lambda function using `dotnet lambda invoke-function` when the parameter to the function is an int. 

Create a new Lambda function - 
```
dotnet new lambda.EmptyFunction --name TakesANumber
```

Change the `FunctionHandler` to  -

```
public string FunctionHandler(int input, ILambdaContext context)
{
    return $"you sent {input}";
}
```

Can't find a way to invoke this function using `dotnet lambda invoke-function --payload ...`

Have tried payloads with a variety of quotes, decimal places, etc. 

### Expected Behavior

Ability to invoke a function with a int payload.

### Current Behavior

Cannot invoke function with an int payload.

### Reproduction Steps

See description.

### AWS .NET SDK and/or Package version used

Amazon Lambda Tools for .NET Core applications (5.4.2)

### Targeted .NET Platform

.NET 6

### Operating System and version

Windows 10

ashishdhingra:
Reproducible. Sending just `12` (for example) via AWS Lambda console works.

ashishdhingra:
We need to implement fix similar to one in PR https://github.com/aws/aws-lambda-dotnet/pull/1851.

ashishdhingra:
Fix released in [Amazon.Lambda.Tools 5.11.1](https://www.nuget.org/packages/Amazon.Lambda.Tools/5.11.1)

Issue #1135, opened 2022-04-05
Escaping of string parameter different for `dotnet lambda-test-tool-6.0...` and `dotnet lambda invoke-function...`


bryanjhogan:
## Description
Invoking a Lambda function that takes a string as a parameter from the `lambda-test-tool` requires different escaping than invoking the real Lambda from `dotnet lambda invoke-function`. 

## Reproduction Steps
Create a new Lambda from the template `lambda.EmptyFunction`. 
Deploy to AWS Lambda. 

### Command Prompt 
Run `dotnet lambda invoke-function EmptyFunction --payload ""hello""`, or `dotnet lambda invoke-function EmptyFunction --payload '"hello"' `.
These will work. 

Try `dotnet lambda-test-tool-6.0 --no-ui --payload ""hello""`, or `dotnet lambda-test-tool-6.0 --no-ui --payload '"hello"'`
These will fail - "System.Exception: Error deserializing the input JSON to type String"

### PowerShell
Run `dotnet lambda invoke-function EmptyFunction --payload ""hello""`, or `dotnet lambda invoke-function EmptyFunction --payload '"hello"' `
These will work.

Try `dotnet lambda-test-tool-6.0 --no-ui --payload ""hello""`, or `dotnet lambda-test-tool-6.0 --no-ui --payload '"hello"' `
These will fail - "System.Exception: Error deserializing the input JSON to type String"

### What actually works for lambda-test-tool-6.0
PowerShell - `dotnet lambda-test-tool-6.0 --no-ui --payload '\"hello\"''`

Command Prompt - `dotnet lambda-test-tool-6.0 --no-ui --payload "\"hello\""`

### Logs

```
Captured Log information:

Request failed to execute
Error:
System.Exception: Error deserializing the input JSON to type String
   at Amazon.Lambda.TestTool.Runtime.LambdaExecutor.BuildParameters(ExecutionRequest request, ILambdaContext context) in C:\codebuild\tmp\output\src581676283\src\Tools\LambdaTestTool\src\Amazon.Lambda.TestTool\Runtime\LambdaExecutor.cs:line 215
   at Amazon.Lambda.TestTool.Runtime.LambdaExecutor.ExecuteAsync(ExecutionRequest request) in C:\codebuild\tmp\output\src581676283\src\Tools\LambdaTestTool\src\Amazon.Lambda.TestTool\Runtime\LambdaExecutor.cs:line 52
---------------- Inner 1 Exception ------------
[...]

PR #1851, opened 2024-10-28
Fixed an issue where delimited string and primitive values in payload were not parsed properly by Lambda Test Tool.

ashishdhingra:
*Issue #, if available:* #1135

*Description of changes:*
Fixed an issue where delimited JSON string payload was not parsed properly by Lambda Test Tool. This PR ports the string handling from .NET Lambda CLI tooling. .NET CLI Extensions tooling per logic [here](https://github.com/aws/aws-extensions-for-dotnet-cli/blob/2e5f84b59dfdb05eae42c76d7cb56b197549a6cc/src/Amazon.Lambda.Tools/Commands/InvokeFunctionCommand.cs#L84) delimits the raw string payload with `"` if it doesn't start with `{`. JSON strings are enclosed in double quotes:
```C#
if(!invokeRequest.Payload.StartsWith("{"))
{
    invokeRequest.Payload = "\"" + invokeRequest.Payload + "\"";
}
```
Per testing with `dotnet lambda invoke-function`:
- For `dotnet lambda-test-tool-6.0 --no-ui --payload ""hello""`, payload should be parsed as `hello`.
- For `dotnet lambda-test-tool-6.0 --no-ui --payload '"hello"'`, payload should be parsed as `'hello'`.

**EDIT:** Per discussion with @normj, the implementation in .NET Extensions for CLI would not handle primitive types and needs to be fixed (refer https://github.com/aws/aws-lambda-dotnet/issues/1197). I have implemented changes as suggested in this PR. Would work on .NET Extensions CLI tooling as part of issue https://github.com/aws/aws-lambda-dotnet/issues/1197.  
___


Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/TestToolStartup.cs
```diff
@@ -318,6 +318,12 @@ private static string DeterminePayload(LocalLambdaOptions localLambdaOptions, Co
                 }
             }
 
+            // Enclose simple string payload in double quotes so that it's parsed properly by Lambda JSON serializer.
+            if (!payload.StartsWith("{") && !payload.StartsWith("\""))
```
normj, line 322: I'm worried the logic we added in Amazon.Lambda.Tools isn't great logic to duplicate. The problem with the Amazon.Lambda.Tools logic is what if you want to send a bool or an int. This logic automatically converts `true` to `"true"` or `100` to `"100"`.

I wonder instead if we should attempt to parse the JSON using with System.Text.Json. If that parses with no problems then send as is. If it doesn't parse add surrounding quotes if it doesn't already start with quotes then use that value. We could retrofit Amazon.Lambda.Tools with the same logic. I wouldn't view that as a breaking change there and we would fix some scenarios that aren't working correct in Amazon.Lambda.Tools.

normj (changes requested):

Tools/LambdaTestTool/src/Amazon.Lambda.TestTool/TestToolStartup.cs
ashishdhingra: Implemented fix. Please review.

normj (approved):

philasmar (approved):

Issue #665, opened 2020-05-07
LambdaTestTool not populating LambdaContext


dude0001:
Is the mock lambda test tool supposed to populate values in the ILambdaContext in my function handler?

Specifically, in my config file, I have `function-timeout` set to 30 and `function-memory-size` set to 512. I am expecting those to be populated in the ILambdaContext that gets passed in as `RemainingTime` and `MemoryLimitInMB` respectively, but they are just getting their defaults zero values. 

This is problematic in my function as I set up a CancelationTokenSource that reserves a few seconds of `RemainingTime` to gracefully shutdown. I have to add some special logic to account for the 0 value that tool passes in that I don't really want in my code.

```
{
  "Information": [
    "This file provides default values for the deployment wizard inside Visual Studio and the AWS Lambda commands added to the .NET Core CLI.",
    "To learn more about the Lambda commands with the .NET Core CLI execute the following command at the command line in the project root directory.",

    "dotnet lambda help",

    "All the command line options for the Lambda command can be specified in this file."
  ],

  "profile": "default",
  "region": "us-east-2",
  "configuration": "Debug",
  "framework": "netcoreapp3.1",
  "function-runtime": "dotnetcore3.1",
  "function-memory-size": 512,
  "function-timeout": 30,
  "function-handler": "{assembly}::{namespace}.{class}::{method}"
}
```

![image](https://user-images.githubusercontent.com/148839/81244927-ac56ba80-8fe1-11ea-80f9-6882a2e734d4.png)

ryancormack:
I've made a start to looking at implementing this (I need/would like the AwsRequestId populated), but I'm not sure how we should handle certain scenarios.

Auto populating certaing properties, like the Request Id and not the other 'always available' properties would be odd.

So passing them all in as a config file seems like a sensible thing to do, but what should we do in scenarios when a user doesn't pass them in in the config? Should we default things like `InvokedFunctionArn` or the `FunctionVersion` to something?

@philasmar or @normj  Do you know if anything has start looking into this?

patricm-enbw:
This would be a great feature!

tim-hampton:
> This would be a great feature!

I agree.

Sonic198:
Will that be implemented? I would also like to see `RemainingTime` to be populated from `aws-lambda-tools-defaults.json`

normj:
[...]

PR #1852, opened 2024-10-29
Implement support for function-timeout configuration

unholyranger-work:
*Issue #665*

*Description of changes:*
The templates deploy a `function-timeout` field which should correlate to the number of seconds until the lambda function times out and exits. This is a requirement for functions that rely on the `ILambdaContext.RemainingTime` as a `CancellationTokenSoucre`. The remaining time is always set to `TimeSpan.Zero` currently making it harder to test, or requiring to break and manually set the value.


UnholyRanger:
Hey @96malhar and @philasmar, could you please review? Thanks

normj:
This is an interesting change for the test tool and wonder if would disrupt user's debugging in other ways. For example I can see users having `function-timeout` set to something small for production purposes like 10 seconds. But if we take this change as is and the value is used as the `CancellationTokenSource` then users would only have 10 seconds to debug.

Is the currently struggle that the value is `0` and that isn't an acceptable value for `CancellationTokenSource`. Wondering if instead of taking the `function-timeout` from the config file which is meant for deployment should the test tool set the `RemainingTime` to some high value so that the same code can be used for setting up the `CancellationTokenSource` but we would not interfere with debugging. Otherwise I think we have to come up some sort of opt-in mechanism to set the `RemainingTime` instead of always taking deployment config files `function-timeout` value.

unholyranger-work:
Thanks for the reply @normj!
The current struggle is that since we're not loading in the `ILambdaContext.RemainingTime` which is supposed to be equivalent to the lambda's configured `Timeout`, the value is always 0 seconds. For lambdas where we want to cancel before AWS terminates the run, we need to know how long we have to operate. In my example, we allow the full 15-minute runtime but give 2 minutes to gracefully shut down. So any runs that go long, can stop processing, upload logs, etc, and exit gracefully without being terminated.
In my view, passing the config value into their appropriate context properties brings the test tool closer to how Lambda functions work. For anyone using the remaining time as a cancellation token, when checking `cancellationToken.IsCancellationRequested` it is always true. In my case, I have a loop at the start that exits right away. For anyone with short time frames, anything is better than 0. I have to hit a breakpoint and manually update the value before continuing.

UnholyRanger:
Hey @normj @96malhar @philasmar,
I wanted to follow up to see if there is any additional feedback on this issue?

Thanks

Sonic198:
Is there any estimate when this could be deployed? I'm also waiting for this feature.

UnholyRanger:
@normj What do you think of the idea of having an additional configuration property that can override the function-timeout during debugging sessions. This way, they can either use the configured debug timeout, the function's real timeout, or the default 15 minutes. Either way, my argument that any of these are better than the existing 0 second timeout.

Also, any other recommendations such that we could get this pushed out?
Thanks

GarrettBeatty (approved):

dscpinheiro:
Closing this as the V2 test tool is out. Please let us know if that doesn't work for you.

PR #1876, opened 2024-11-19
Sync Dev from Master

normj:
*Description of changes:*
Sync dev branch with latest commits to master.


Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/LambdaBootstrap.cs
```diff
@@ -137,10 +151,41 @@ private LambdaBootstrap(HttpClient httpClient, LambdaBootstrapHandler handler, L
...
 
-            bool doStartInvokeLoop = _initializer == null || await InitializeAsync();
+                object registry = null;
+                try
+                {
+                    registry = SnapstartHelperCopySnapshotCallbacksIsolated.CopySnapshotCallbacks();
+                }
+                catch (TypeLoadException ex)
+                {
+                    Client.ConsoleLogger.FormattedWriteLine(
```
boblodgett, line 173: This doesn't appear to terminate or exit the call which means it will fall through where registry will be null. The following comment says `no expection in calling SnapStart hooks` but it looks like there could have been an exception still. Is this correct?

boblodgett (approved):

PR #1882, opened 2024-11-25
ci: onboard Amazon.Lambda.Templates to AutoVer

philasmar:
*Description of changes:*
* Onboard Amazon.Lambda.Templates to AutoVer so we can automatically version it using change files.

### DIFF
--- a/.autover/autover.json
+++ b/.autover/autover.json
@@ -127,6 +127,10 @@
         "Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Amazon.Lambda.TestTool.BlazorTester80-pack.csproj",
         "Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Amazon.Lambda.TestTool.BlazorTester90-pack.csproj"
       ]
+    },
+    {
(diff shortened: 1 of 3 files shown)


.autover/autover.json
```diff
@@ -127,6 +127,18 @@
         "Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Amazon.Lambda.TestTool.BlazorTester80-pack.csproj",
         "Tools/LambdaTestTool/src/Amazon.Lambda.TestTool.BlazorTester/Amazon.Lambda.TestTool.BlazorTester90-pack.csproj"
       ]
+    },
+    {
```
normj, line 131: I was going to a PR soon to remove the VS 2017 and 2019 blueprints. I don't think we should bother putting those here and just have `Amazon.Lambda.Templates` pointing to the VS 2022. If there is a new VS that comes out we can revisit this configuration.

normj (changes requested):

.autover/autover.json
philasmar: Fixed

normj (approved):

gcbeattyAWS (approved):

Issue #1854, opened 2024-10-31
AspNetCoreServer incorrectly sets IHttpRequestBodyDetectionFeature.CanHaveBody that leads to breaking change after net 7 upgrade


MadSciencist:
### Describe the bug

Hi,

It appears that the behavior of the APIGatewayProxyFunction is inconsistent with the behavior of Kestrel when handling requests with an empty body directed to endpoints with a parameter marked [FromBody(EmptyBodyBehavior = EmptyBodyBehavior.Allow)].

In the InvokeFeatures class, the CanHaveBody property is set based on requestFeature.Body != null. This is problematic because the IHttpRequestFeature.Body property is initialized with a new MemoryStream() value, therefor CanHaveBody is never false.

As a result, this change introduces a breaking behavior when upgrading from .NET 6.0 to .NET 7.0 in scenarios where requests with an empty body are sent from AWS Lambda. ASP.NET Core attempts to deserialize the request (due to the CanHaveBody property),  and because the request body is an empty stream, it causes the API to throw an error.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Endpoints with parameters marked with `[FromBody(EmptyBodyBehavior = EmptyBodyBehavior.Allow)]` attribute can handle request both with and without body.

### Current Behavior

AspNetCore tries to deserialize the body and throws `The input does not contain any JSON tokens. Expected the input to start with a valid JSON token, when isFinalBlock is true. Path: $ | LineNumber: 0 | BytePositionInLine: 0.`

### Reproduction Steps

Create `PUT` endpoint with optional body:

```csharp
    public IActionResult Test([FromBody(EmptyBodyBehavior = EmptyBodyBehavior.Allow)] Body request = default)
    {
        return Accepted();
    }

    public class Body
    {
        public string Prop { get; set; }
    }
```
and send requests with and without body.

### AWS .NET SDK and/or Package version used

Amazon.Lambda.AspNetCoreServer 9.0.1

### Targeted .NET Platform

NET 8.0

### Operating System and version

Windows 10

ashishdhingra:
[...]

PR #1885, opened 2024-11-26
Fixed issue "(EmptyBodyBehavior = EmptyBodyBehavior.Allow)" not being honored

normj:
*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/1854

*Description of changes:*
ASP.NET Core controllers that were configured with `(EmptyBodyBehavior = EmptyBodyBehavior.Allow)` should allow requests for PUT and POST to have no body. The controller parameter would be null in this case. This is controlled by our underlying implementation of ASP.NET Core's `IHttpRequestBodyDetectionFeature` interface. Our implementation check to see if the Body property was null or not. This didn't really do anything because we initialize the Body to an empty MemoryStream.

When looking at the docs for [IHttpRequestBodyDetectionFeature.CanHaveBody](https://github.com/dotnet/aspnetcore/blob/80fdd48f638c15b13f56c187b09fe89082c46663/src/Http/Http.Features/src/IHttpRequestBodyDetectionFeature.cs#L18) it says it should return true for HTTP 1.1 if the request body length is greater then 0. The use case of running ASP.NET Core through Lambda only supports HTTP 1.1. So the fix for the issue is instead of just checking null for the body also check if the length is greater then 0.


ashishdhingra (approved):

.autover/changes/7db9a194-0cc2-43e7-81da-e2f468d3fc90.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.AspNetCoreServer",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Fixed issue \u0027(EmptyBodyBehavior = EmptyBodyBehavior.Allow)\u0027 not being honored when request body was empty"
```
dscpinheiro, line 7: Does autover require 2 files? Also curious if `\u0027` is here intentionally.

normj: I think if I crafted the tool by hand I could have done it with one file with multiple items in the `Projects` array. I used the CLI which just does one project at a time. For this change the change log will end up the same whether I did it as one file or 2.

As for the `\u0027` I think that was just the escaping of the `'` I used on the CLI.

dscpinheiro (approved):

PR #1894, opened 2024-12-04
add packaging of runtime

gcbeattyAWS:
*Issue #, if available:*

*Description of changes:*

1. Add Amazon.Lambda.RuntimeSupport dll to the output nuget package
Screenshot 2024-12-04 154218
2. Add test to verify dll is inside the content folder.

*Testing*
1. So far only tested via checking that the files are in the nuget package manually. I still need to test that this dll works by running with the aspire proof of concept.

### DIFF
--- a/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
+++ b/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
@@ -15,9 +15,14 @@
   </PropertyGroup>
 	
   <ItemGroup>
+      <PackageReference Include="Amazon.Lambda.RuntimeSupport" Version="1.11.0" />
       <PackageReference Include="Blazored.Modal" Version="7.3.1" />
 	  <PackageReference Include="Microsoft.Extensions.FileProviders.Embedded" Version="8.0.11" /> 
-  </ItemGroup>	
+  </ItemGroup>
+
+  <ItemGroup>
+	  <None Include="$(OutputPath)Amazon.Lambda.RuntimeSupport.dll" Pack="true" PackagePath="content" />
+  </ItemGroup>
 	
   <ItemGroup>
     <EmbeddedResource Include="wwwroot\**" />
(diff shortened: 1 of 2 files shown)


Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
```diff
@@ -15,9 +15,14 @@
...
 	
   <ItemGroup>
+      <PackageReference Include="Amazon.Lambda.RuntimeSupport" Version="1.12.0" />
       <PackageReference Include="Blazored.Modal" Version="7.3.1" />
 	  <PackageReference Include="Microsoft.Extensions.FileProviders.Embedded" Version="8.0.11" /> 
-  </ItemGroup>	
+  </ItemGroup>
+
+  <ItemGroup>
+	  <None Include="$(OutputPath)Amazon.Lambda.RuntimeSupport.dll" Pack="true" PackagePath="content" />
```
gcbeattyAWS, line 24: wasnt sure if this needed to be `content` or `lib` folder. was seeing this warning when doing pack command

```
The assembly 'content\Amazon.Lambda.RuntimeSupport.dll' is not inside the 'lib' folder and hence it won't be added as a reference when the package is installed into a project. Move it into the 'lib' folder if it needs to be referenced.
```

but in our case in we are referencing the dll as part of a command line argument, so not sure if it really matters

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.IO;
+using System.IO.Compression;
+using Xunit;
+
+namespace Amazon.Lambda.TestTool.UnitTests;
+
+public class PackagingTests
+{
+    [Fact]
+    public void VerifyPackageContentsHasRuntimeSupport()
```
gcbeattyAWS, line 12: this test case seemed like overkill but i added it anyway

PR #1897, opened 2024-12-05
Add Amazon.Lambda.RuntimeSupport dll to the output nuget package

gcbeattyAWS:
*Issue #, if available:*

*Description of changes:*

1. Add all runtime versions of Amazon.Lambda.RuntimeSupport dll to the output nuget package. each dll is in its own folder in the content folder
3. Add test to verify dlsl are inside the content folder.

*Testing*
1. I copied the outputted dlls into the lampda poc project and verified that they work
image
image

image

### DIFF
--- a/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
+++ b/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
@@ -14,6 +14,7 @@
     <PackageId>Amazon.Lambda.TestTool</PackageId>
     <ToolCommandName>dotnet-lambda-test-tool</ToolCommandName>
     <Version>0.0.1-beta.1</Version>
+    <NoWarn>NU5100</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
@@ -29,8 +30,31 @@
     <PackageReference Include="BlazorMonaco" Version="3.2.0" />
   </ItemGroup>
 
+  <Target Name="GetRuntimeSupportTargetFrameworks">
+    <Exec Command="dotnet msbuild ..\..\..\..\Libraries\src\Amazon.Lambda.RuntimeSupport\Amazon.Lambda.RuntimeSupport.csproj --getProperty:TargetFrameworks" ConsoleToMSBuild="true">
+      <Output TaskParameter="ConsoleOutput" PropertyName="RuntimeSupportTargetFrameworks" />
+    </Exec>
+  </Target>
+
+  <Target Name="CopyRuntimeSupportFiles" DependsOnTargets="GetRuntimeSupportTargetFrameworks" BeforeTargets="_GetPackageFiles">
+    <ItemGroup>
+      <TempFrameworks Include="$(RuntimeSupportTargetFrameworks.Split(';'))" />
+
+      <TargetFrameworks Include="@(TempFrameworks)"
+                        Condition="'%(Identity)' != 'netstandard2.0'" />
+    </ItemGroup>
+
+    <Exec Command="dotnet publish &quot;$(MSBuildThisFileDirectory)..\..\..\..\Libraries\src\Amazon.Lambda.RuntimeSupport\Amazon.Lambda.RuntimeSupport.csproj&quot; -c $(Configuration) -f %(TargetFrameworks.Identity) /p:ExecutableOutputType=true" />
+
+    <ItemGroup>
+      <None Include="$(MSBuildThisFileDirectory)..\..\..\..\Libraries\src\Amazon.Lambda.RuntimeSupport\bin\$(Configuration)\%(TargetFrameworks.Identity)\publish\**\*.*">
+        <Pack>true</Pack>
+        <PackagePath>content\Amazon.Lambda.RuntimeSupport\%(TargetFrameworks.Identity)</PackagePath>
+      </None>
+    </ItemGroup>
+  </Target>
+
   <ItemGroup>
 	  <EmbeddedResource Include="Resources\**" />
   </ItemGroup>
-
 </Project>
--- a/Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
+++ b/Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
@@ -1,0 +1,152 @@
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.IO.Compression;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Amazon.Lambda.TestTool.UnitTests;
+
+public class PackagingTests
+{
+    private readonly ITestOutputHelper _output;
+    private readonly string[] _expectedFrameworks;
+
+    public PackagingTests(ITestOutputHelper output)
+    {
+        _output = output;
+        _expectedFrameworks = GetRuntimeSupportTargetFrameworks()
+            .Split([';'], StringSplitOptions.RemoveEmptyEntries)
+            .Where(f => f != "netstandard2.0")
+            .ToArray();
+    }
+
+    private string GetRuntimeSupportTargetFrameworks()
+    {
+        var solutionRoot = FindSolutionRoot();
+        var runtimeSupportPath = Path.Combine(solutionRoot, "Libraries", "src", "Amazon.Lambda.RuntimeSupport", "Amazon.Lambda.RuntimeSupport.csproj");
+
+        var process = new Process
+        {
+            StartInfo = new ProcessStartInfo
+            {
+                FileName = "dotnet",
+                Arguments = $"msbuild {runtimeSupportPath} --getProperty:TargetFrameworks",
+                RedirectStandardOutput = true,
+                RedirectStandardError = true,
+                UseShellExecute = false,
+                CreateNoWindow = true,
+            }
+        };
+
+        process.Start();
+        var output = process.StandardOutput.ReadToEnd();
+        var error = process.StandardError.ReadToEnd();
+        process.WaitForExit();
+
+        if (process.ExitCode != 0)
+        {
+            throw new Exception($"Failed to get TargetFrameworks: {error}");
+        }
+
+        return output.Trim();
+    }
+
+    [Fact]
+    public void VerifyPackageContentsHasRuntimeSupport()
+    {
+        var solutionRoot = FindSolutionRoot();
+        var projectPath = Path.Combine(solutionRoot, "Tools", "LambdaTestTool-v2", "src", "Amazon.Lambda.TestTool", "Amazon.Lambda.TestTool.csproj");
+
+        _output.WriteLine("\nPacking TestTool...");
+        var packProcess = new Process
+        {
+            StartInfo = new ProcessStartInfo
+            {
+                FileName = "dotnet",
+                Arguments = $"pack {projectPath} -c Release",
+                RedirectStandardOutput = true,
+                RedirectStandardError = true,
+                UseShellExecute = false,
+                CreateNoWindow = true,
+            }
+        };
+
+        packProcess.Start();
+        string packOutput = packProcess.StandardOutput.ReadToEnd();
+        string packError = packProcess.StandardError.ReadToEnd();
+        packProcess.WaitForExit();
+
+        _output.WriteLine("Pack Output:");
+        _output.WriteLine(packOutput);
+        if (!string.IsNullOrEmpty(packError))
+        {
+            _output.WriteLine("Pack Errors:");
+            _output.WriteLine(packError);
+        }
+
+        Assert.Equal(0, packProcess.ExitCode);
+
+        var packageDir = Path.Combine(Path.GetDirectoryName(projectPath), "bin", "Release");
+        _output.WriteLine($"Looking for package in: {packageDir}");
+
+        var packageFiles = Directory.GetFiles(packageDir, "*.nupkg", SearchOption.AllDirectories);
+        Assert.True(packageFiles.Length > 0, $"No .nupkg files found in {packageDir}");
+
+        var packagePath = packageFiles[0];
+        _output.WriteLine($"Found package: {packagePath}");
+
+        using var archive = ZipFile.OpenRead(packagePath);
+        // Verify each framework has its required files
+        foreach (var framework in _expectedFrameworks)
+        {
+            _output.WriteLine($"\nChecking framework: {framework}");
+
+            // Get all files for this framework
+            var frameworkFiles = archive.Entries
+                .Where(e => e.FullName.StartsWith($"content/Amazon.Lambda.RuntimeSupport/{framework}/"))
+                .Select(e => e.FullName)
+                .ToList();
+
+            // Verify essential files exist
+            var essentialFiles = new[]
+            {
+                $"content/Amazon.Lambda.RuntimeSupport/{framework}/Amazon.Lambda.Core.dll",
+                $"content/Amazon.Lambda.RuntimeSupport/{framework}/Amazon.Lambda.RuntimeSupport.dll",
+                $"content/Amazon.Lambda.RuntimeSupport/{framework}/Amazon.Lambda.RuntimeSupport.deps.json",
+                $"content/Amazon.Lambda.RuntimeSupport/{framework}/Amazon.Lambda.RuntimeSupport.runtimeconfig.json"
+            };
+
+            var missingFiles = essentialFiles.Where(f => !frameworkFiles.Contains(f)).ToList();
+
+            if (missingFiles.Any())
+            {
+                Assert.Fail($"The following essential files are missing for {framework}:\n" +
+                            string.Join("\n", missingFiles));
+            }
+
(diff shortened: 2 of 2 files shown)


Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
```diff
@@ -15,13 +15,17 @@
...
 	
   <ItemGroup>
+      <PackageReference Include="Amazon.Lambda.RuntimeSupport" Version="1.11.0" />
       <PackageReference Include="Blazored.Modal" Version="7.3.1" />
 	  <PackageReference Include="Microsoft.Extensions.FileProviders.Embedded" Version="8.0.11" /> 
-  </ItemGroup>	
+  </ItemGroup>
+
+  <ItemGroup>
+	  <None Include="$(OutputPath)Amazon.Lambda.RuntimeSupport.dll" Pack="true" PackagePath="content" Visible="false" />
```
gcbeattyAWS, line 24: wasnt sure if this needed to be `content` or `lib` folder. was seeing this warning when doing pack command

```
The assembly 'content\Amazon.Lambda.RuntimeSupport.dll' is not inside the 'lib' folder and hence it won't be added as a reference when the package is installed into a project. Move it into the 'lib' folder if it needs to be referenced.
```

but in our case in we are referencing the dll as part of a command line argument in the aspire proof of concept , so not sure if it really matters

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.IO;
+using System.IO.Compression;
+using Xunit;
+
+namespace Amazon.Lambda.TestTool.UnitTests;
+
+public class PackagingTests
+{
+    [Fact]
+    public void VerifyPackageContentsHasRuntimeSupport()
```
gcbeattyAWS, line 12: added this test case but wasnt sure it was overkill or not

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
gcbeattyAWS, line 24: i still need to test end to end with aspire poc if any more (lambda core dlls) are needed or anything else

philasmar: I'm assuming this is only adding the current target framework (net8.0) version of RuntimeSupport. I believe we need to add all targets of Runtimesupport. @normj is my assumption correct?

philasmar (commented):
I'm assuming this is only adding the current target framework (net8.0) version of RuntimeSupport. I believe we need to add all targets of Runtimesupport. @normj is my assumption correct?

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
gcbeattyAWS: i remember he mentioned when he updated the test tool v2 something about to not worry about the other versions. but ill let him confirm

normj: We will need to distribute each target framework for Lambda RuntimeSupport. That is how things work in Lambda. I think what you are thinking about is I meant with test tool v2 we only need to ship one version of the test tool NuGet package unlike the current v1 version were we ship separate versions of the package for each target framework. In the end we will have a single V2 test tool NuGet package with one .NET tool which is the Lambda emulator and then separate content folders for each version of RuntimeSupport with RuntimeSupport's dependencies.

gcbeattyAWS: Ahh ok let me make that change. thanks for the detailed explanation

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
```diff
@@ -23,8 +22,24 @@
   </ItemGroup>
 
   <ItemGroup>
-    <EmbeddedResource Include="wwwroot\**" />
-	  <EmbeddedResource Include="Resources\**" />
+    <ProjectReference Include="..\..\..\..\Libraries\src\Amazon.Lambda.RuntimeSupport\Amazon.Lambda.RuntimeSupport.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <RuntimeSupportFramework Include="net5.0;net6.0;net8.0;netstandard2.0" />
```
gcbeattyAWS, line 29: we havent updated runtime support in this branch to add .net9.0 which is why i havent added it here

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
gcbeattyAWS: ive updated it now, where each runtime support target version has its own folder with a dll in the content folder of the test tool

gcbeattyAWS: ive also updated the screenshot in the description

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
```diff
@@ -23,8 +22,24 @@
...
   </ItemGroup>
 
-</Project>
+  <Target Name="ListRuntimeSupportDlls" BeforeTargets="PrepareForBuild">
+    <ItemGroup>
+      <Content Include="$(MSBuildThisFileDirectory)..\..\..\..\Libraries\src\Amazon.Lambda.RuntimeSupport\bin\$(Configuration)\%(RuntimeSupportFramework.Identity)\Amazon.Lambda.RuntimeSupport.dll">
```
gcbeattyAWS, line 34: i havent been able to test end to end yet with aspire to see if any dependency dlls are also needed for runtime support. Unless someone happens to know off the top of their head if we need any other dlls?

gcbeattyAWS: possibly lambda core is needed too but im yet to confirm

normj (commented):
Instead of adding the project reference and manually copying the dll I wonder if we should have a customer target that runs before the `Pack` target and run some script that does a `dotnet publish` for each supported target. And then copy the output publish folder to a content folder in the test package for each target framework. That would cover all of the dependencies and would be the same as who the Runtime assembly and its dependencies are copied into the image.

Plus having the project dependency even though we don't use it is annoying because it means we have to keep building RuntimeSupport and its dependencies as part of our dev loop.

gcbeattyAWS:
> Instead of adding the project reference and manually copying the dll I wonder if we should have a customer target that runs before the `Pack` target and run some script that does a `dotnet publish` for each supported target. And then copy the output publish folder to a content folder in the test package for each target framework. That would cover all of the dependencies and would be the same as who the Runtime assembly and its dependencies are copied into the image.
> 
> Plus having the project dependency even though we don't use it is annoying because it means we have to keep building RuntimeSupport and its dependencies as part of our dev loop.

i have updated the latest logic to do this

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
gcbeattyAWS: I've updated the latest logic to now use dotnet publish and copy everything in the publish folder to the content folder. and i have updated the screenshot in the description

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
```diff
@@ -23,9 +22,25 @@
     <PackageReference Include="BlazorMonaco" Version="3.2.0" />
   </ItemGroup>
 
+  <Target Name="CopyRuntimeSupportFiles" BeforeTargets="_GetPackageFiles">
+    <Message Importance="high" Text="Publishing Amazon.Lambda.RuntimeSupport project..." />
+    
+    <ItemGroup>
+      <TargetFrameworks Include="netstandard2.0;net5.0;net6.0;net8.0" />
```
philasmar, line 29: You should drop net5 and add net9

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
```diff
@@ -23,9 +22,25 @@
     <PackageReference Include="BlazorMonaco" Version="3.2.0" />
   </ItemGroup>
 
+  <Target Name="CopyRuntimeSupportFiles" BeforeTargets="_GetPackageFiles">
+    <Message Importance="high" Text="Publishing Amazon.Lambda.RuntimeSupport project..." />
```
philasmar, line 26: nit: message is a bit misleading or non descriptive

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
```diff
@@ -0,0 +1,127 @@
...
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Amazon.Lambda.TestTool.UnitTests;
+
+public class PackagingTests
+{
+    private readonly ITestOutputHelper _output;
+    private static readonly string[] ExpectedFrameworks = new[]
+    {
```
philasmar, line 16: update frameworks here too

philasmar (changes requested):

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
gcbeattyAWS: yea i iwll remove the message

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
gcbeattyAWS: ive updated this in my local branch but waiting on https://github.com/aws/aws-lambda-dotnet/pull/1937/files to be merged first

gcbeattyAWS: fixed

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
gcbeattyAWS: fixed

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
gcbeattyAWS: done

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
gcbeattyAWS: i tested this end to end and this works now

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
```diff
@@ -1,14 +1,13 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
-
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
normj, line 1: Because of our unorthodox packaging of the assemblies I'm getting a lot of `NU5100` warnings. Can you add `<NoWarn>NU5100</NoWarn>` to the `PropertyGroup`.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
```diff
@@ -23,9 +22,23 @@
...
+    <ItemGroup>
+      <TargetFrameworks Include="netstandard2.0;net6.0;net8.0;net9.0" />
+    </ItemGroup>
+
+    <Exec Command="dotnet publish &quot;$(MSBuildThisFileDirectory)..\..\..\..\Libraries\src\Amazon.Lambda.RuntimeSupport\Amazon.Lambda.RuntimeSupport.csproj&quot; -c $(Configuration) -f %(TargetFrameworks.Identity) /p:ExecutableOutputType=true -v n" />
+
+    <ItemGroup>
+      <None Include="$(MSBuildThisFileDirectory)..\..\..\..\Libraries\src\Amazon.Lambda.RuntimeSupport\bin\$(Configuration)\%(TargetFrameworks.Identity)\publish\**\*.*">
+        <Pack>true</Pack>
+        <PackagePath>content\%(TargetFrameworks.Identity)</PackagePath>
```
normj, line 35: Can you change `content\%(TargetFrameworks.Identity)` to `content\Amazon.Lambda.RuntimeSupport\%(TargetFrameworks.Identity)` to give some name spacing to the content here.

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
normj: Would be great if instead of having this hardcoded list if you read the `TargetFrameworks` from the Amazon.Lambda.RuntimeSupport project file. That way we don't have to update this every time we add a new supported version.

You can skip .NET Standard 2.0.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
```diff
@@ -23,9 +22,23 @@
...
 
+  <Target Name="CopyRuntimeSupportFiles" BeforeTargets="_GetPackageFiles">
+    <ItemGroup>
+      <TargetFrameworks Include="netstandard2.0;net6.0;net8.0;net9.0" />
+    </ItemGroup>
+
+    <Exec Command="dotnet publish &quot;$(MSBuildThisFileDirectory)..\..\..\..\Libraries\src\Amazon.Lambda.RuntimeSupport\Amazon.Lambda.RuntimeSupport.csproj&quot; -c $(Configuration) -f %(TargetFrameworks.Identity) /p:ExecutableOutputType=true -v n" />
+
+    <ItemGroup>
+      <None Include="$(MSBuildThisFileDirectory)..\..\..\..\Libraries\src\Amazon.Lambda.RuntimeSupport\bin\$(Configuration)\%(TargetFrameworks.Identity)\publish\**\*.*">
```
normj, line 33: It would be great if we could skip `netstandard2.0` because it doesn't make any sense in this context. The `netstandard2.0` target is for class libraries not executables.

normj (changes requested):

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
gcbeattyAWS: done

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
gcbeattyAWS: done

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
gcbeattyAWS: ive updated the logic in the csproj and the unit test to dynamically get the targetframeworks and also skip netstandard2.0

normj (approved):
Nice job. Some impressive MSBuild work here.

philasmar (approved):

PR #1901, opened 2024-12-09
Add ApiGatewayHttpApiV2ProxyRequestTranslator and ApiGatewayProxyRequestTranslator

gcbeattyAWS:
DOTNET-7838

*Description of changes:*
1. Add functions to translate httpcontext into APIGatewayHttpApiV2ProxyRequest and APIGatewayProxyRequest 
2. Add unit tests for above classes

*Testing details*
1. Wrote unit tests
2. integration tests

### DIFF
--- a/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
+++ b/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
@@ -1,0 +1,196 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace Amazon.Lambda.TestTool.Extensions;
+
+using System.Web;
+using Amazon.Lambda.APIGatewayEvents;
+using Amazon.Lambda.TestTool.Models;
+using Amazon.Lambda.TestTool.Utilities;
+using static Amazon.Lambda.APIGatewayEvents.APIGatewayHttpApiV2ProxyRequest;
+
+/// <summary>
+/// Provides extension methods to translate an <see cref="HttpContext"/> to different types of API Gateway requests.
+/// </summary>
+public static class HttpContextExtensions
+{
+    /// <summary>
+    /// Translates an <see cref="HttpContext"/> to an <see cref="APIGatewayHttpApiV2ProxyRequest"/>.
+    /// </summary>
+    /// <param name="context">The <see cref="HttpContext"/> to be translated.</param>
+    /// <param name="apiGatewayRouteConfig">The configuration of the API Gateway route, including the HTTP method, path, and other metadata.</param>
+    /// <returns>An <see cref="APIGatewayHttpApiV2ProxyRequest"/> object representing the translated request.</returns>
+    public static APIGatewayHttpApiV2ProxyRequest ToApiGatewayHttpV2Request(
+        this HttpContext context,
+        ApiGatewayRouteConfig apiGatewayRouteConfig)
+    {
+        var request = context.Request;
+        var currentTime = DateTimeOffset.UtcNow;
+        var body = HttpRequestUtility.ReadRequestBody(request);
+        var contentLength = HttpRequestUtility.CalculateContentLength(request, body);
+
+        var pathParameters = RouteTemplateUtility.ExtractPathParameters(apiGatewayRouteConfig.Path, request.Path);
+
+        // Format 2.0 doesn't have multiValueHeaders or multiValueQueryStringParameters fields. Duplicate headers are combined with commas and included in the headers field.
+        var (_, allHeaders) = HttpRequestUtility.ExtractHeaders(request.Headers);
+        var headers = allHeaders.ToDictionary(
+            kvp => kvp.Key,
+            kvp => string.Join(", ", kvp.Value)
+        );
+
+        // Duplicate query strings are combined with commas and included in the queryStringParameters field.
+        var (_, allQueryParams) = HttpRequestUtility.ExtractQueryStringParameters(request.Query);
+        var queryStringParameters = allQueryParams.ToDictionary(
+            kvp => kvp.Key,
+            kvp => string.Join(",", kvp.Value)
+        );
+
+        string userAgent = request.Headers.UserAgent.ToString();
+
+        if (!headers.ContainsKey("content-length"))
+        {
+            headers["content-length"] = contentLength.ToString();
+        }
+
+        if (!headers.ContainsKey("content-type"))
+        {
+            headers["content-type"] = "text/plain; charset=utf-8";
+        }
+
+        var httpApiV2ProxyRequest = new APIGatewayHttpApiV2ProxyRequest
+        {
+            RouteKey = $"{request.Method} {apiGatewayRouteConfig.Path}",
+            RawPath = request.Path.Value, // this should be decoded value
+            Body = body,
+            IsBase64Encoded = false,
+            RequestContext = new ProxyRequestContext
+            {
+                Http = new HttpDescription
+                {
+                    Method = request.Method,
+                    Path = request.Path.Value, // this should be decoded value
+                    Protocol = !string.IsNullOrEmpty(request.Protocol) ? request.Protocol : "HTTP/1.1", // defaults to http 1.1 if not provided
+                    UserAgent = userAgent
+                },
+                Time = currentTime.ToString("dd/MMM/yyyy:HH:mm:ss") + " +0000",
+                TimeEpoch = currentTime.ToUnixTimeMilliseconds(),
+                RequestId = HttpRequestUtility.GenerateRequestId(),
+                RouteKey = $"{request.Method} {apiGatewayRouteConfig.Path}",
+            },
+            Version = "2.0"
+        };
+
+        if (request.Cookies.Any())
+        {
+            httpApiV2ProxyRequest.Cookies = request.Cookies.Select(c => $"{c.Key}={c.Value}").ToArray();
+        }
+
+        if (headers.Any())
+        {
+            httpApiV2ProxyRequest.Headers = headers;
+        }
+
+        httpApiV2ProxyRequest.RawQueryString = string.Empty; // default is empty string
+        if (queryStringParameters.Any())
+        {
+            // this should be decoded value
+            httpApiV2ProxyRequest.QueryStringParameters = queryStringParameters;
+
+            // this should be the url encoded value and not include the "?"
+            // e.g. key=%2b%2b%2b
+            httpApiV2ProxyRequest.RawQueryString = HttpUtility.UrlPathEncode(request.QueryString.Value?.Substring(1));
+
+        }
+
+        if (pathParameters.Any())
+        {
+            // this should be decoded value
+            httpApiV2ProxyRequest.PathParameters = pathParameters;
+        }
+
+        if (HttpRequestUtility.IsBinaryContent(request.ContentType))
+        {
+            // we already converted it when we read the body so we dont need to re-convert it
+            httpApiV2ProxyRequest.IsBase64Encoded = true;
+        }
+
+        return httpApiV2ProxyRequest;
+    }
+
+    /// <summary>
+    /// Translates an <see cref="HttpContext"/> to an <see cref="APIGatewayProxyRequest"/>.
+    /// </summary>
+    /// <param name="context">The <see cref="HttpContext"/> to be translated.</param>
+    /// <param name="apiGatewayRouteConfig">The configuration of the API Gateway route, including the HTTP method, path, and other metadata.</param>
+    /// <returns>An <see cref="APIGatewayProxyRequest"/> object representing the translated request.</returns>
+    public static APIGatewayProxyRequest ToApiGatewayRequest(
+        this HttpContext context,
+        ApiGatewayRouteConfig apiGatewayRouteConfig)
+    {
+        var request = context.Request;
+        var body = HttpRequestUtility.ReadRequestBody(request);
+        var contentLength = HttpRequestUtility.CalculateContentLength(request, body);
+
+        var pathParameters = RouteTemplateUtility.ExtractPathParameters(apiGatewayRouteConfig.Path, request.Path);
+
+        var (headers, multiValueHeaders) = HttpRequestUtility.ExtractHeaders(request.Headers);
+        var (queryStringParameters, multiValueQueryStringParameters) = HttpRequestUtility.ExtractQueryStringParameters(request.Query);
+
+        if (!headers.ContainsKey("content-length"))
+        {
+            headers["content-length"] = contentLength.ToString();
+            multiValueHeaders["content-length"] = new List<string> { contentLength.ToString() };
+        }
+
+        if (!headers.ContainsKey("content-type"))
+        {
+            headers["content-type"] = "text/plain; charset=utf-8";
+            multiValueHeaders["content-type"] = new List<string> { "text/plain; charset=utf-8" };
+        }
+
+        var proxyRequest = new APIGatewayProxyRequest
+        {
+            Resource = apiGatewayRouteConfig.Path,
+            Path = request.Path.Value,
+            HttpMethod = request.Method,
+            Body = body,
+            IsBase64Encoded = false
+        };
+
+        if (headers.Any())
+        {
+            proxyRequest.Headers = headers;
+        }
+
+        if (multiValueHeaders.Any())
+        {
+            proxyRequest.MultiValueHeaders = multiValueHeaders;
+        }
+
+        if (queryStringParameters.Any())
+        {
+            // this should be decoded value
+            proxyRequest.QueryStringParameters = queryStringParameters;
+        }
+
+        if (multiValueQueryStringParameters.Any())
+        {
+            // this should be decoded value
+            proxyRequest.MultiValueQueryStringParameters = multiValueQueryStringParameters;
+        }
+
+        if (pathParameters.Any())
+        {
+            // this should be decoded value
+            proxyRequest.PathParameters = pathParameters;
+        }
+
+        if (HttpRequestUtility.IsBinaryContent(request.ContentType))
+        {
+            // we already converted it when we read the body so we dont need to re-convert it
+            proxyRequest.IsBase64Encoded = true;
+        }
+
+        return proxyRequest;
+    }
+}
--- a/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
+++ b/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
@@ -3,6 +3,7 @@
 
 namespace Amazon.Lambda.TestTool.Extensions;
 
+using System.Text;
 using System.Web;
 using Amazon.Lambda.APIGatewayEvents;
 using Amazon.Lambda.TestTool.Models;
@@ -20,19 +21,20 @@ public static class HttpContextExtensions
     /// <param name="context">The <see cref="HttpContext"/> to be translated.</param>
     /// <param name="apiGatewayRouteConfig">The configuration of the API Gateway route, including the HTTP method, path, and other metadata.</param>
     /// <returns>An <see cref="APIGatewayHttpApiV2ProxyRequest"/> object representing the translated request.</returns>
-    public static APIGatewayHttpApiV2ProxyRequest ToApiGatewayHttpV2Request(
+    public static async Task<APIGatewayHttpApiV2ProxyRequest> ToApiGatewayHttpV2Request(
         this HttpContext context,
         ApiGatewayRouteConfig apiGatewayRouteConfig)
     {
         var request = context.Request;
         var currentTime = DateTimeOffset.UtcNow;
-        var body = HttpRequestUtility.ReadRequestBody(request);
+        var body = await HttpRequestUtility.ReadRequestBody(request);
         var contentLength = HttpRequestUtility.CalculateContentLength(request, body);
 
         var pathParameters = RouteTemplateUtility.ExtractPathParameters(apiGatewayRouteConfig.Path, request.Path);
 
         // Format 2.0 doesn't have multiValueHeaders or multiValueQueryStringParameters fields. Duplicate headers are combined with commas and included in the headers field.
-        var (_, allHeaders) = HttpRequestUtility.ExtractHeaders(request.Headers);
+        // 2.0 also lowercases all header keys
+        var (_, allHeaders) = HttpRequestUtility.ExtractHeaders(request.Headers, true);
         var headers = allHeaders.ToDictionary(
             kvp => kvp.Key,
             kvp => string.Join(", ", kvp.Value)
@@ -91,6 +93,7 @@ public static class HttpContextExtensions
         }
 
         httpApiV2ProxyRequest.RawQueryString = string.Empty; // default is empty string
+
         if (queryStringParameters.Any())
         {
             // this should be decoded value
@@ -123,12 +126,13 @@ public static class HttpContextExtensions
     /// <param name="context">The <see cref="HttpContext"/> to be translated.</param>
     /// <param name="apiGatewayRouteConfig">The configuration of the API Gateway route, including the HTTP method, path, and other metadata.</param>
     /// <returns>An <see cref="APIGatewayProxyRequest"/> object representing the translated request.</returns>
-    public static APIGatewayProxyRequest ToApiGatewayRequest(
+    public static async Task<APIGatewayProxyRequest> ToApiGatewayRequest(
         this HttpContext context,
-        ApiGatewayRouteConfig apiGatewayRouteConfig)
+        ApiGatewayRouteConfig apiGatewayRouteConfig,
+        ApiGatewayEmulatorMode emulatorMode)
     {
         var request = context.Request;
-        var body = HttpRequestUtility.ReadRequestBody(request);
+        var body = await HttpRequestUtility.ReadRequestBody(request);
         var contentLength = HttpRequestUtility.CalculateContentLength(request, body);
 
         var pathParameters = RouteTemplateUtility.ExtractPathParameters(apiGatewayRouteConfig.Path, request.Path);
@@ -136,22 +140,38 @@ public static class HttpContextExtensions
         var (headers, multiValueHeaders) = HttpRequestUtility.ExtractHeaders(request.Headers);
         var (queryStringParameters, multiValueQueryStringParameters) = HttpRequestUtility.ExtractQueryStringParameters(request.Query);
 
-        if (!headers.ContainsKey("content-length"))
+        if (!headers.ContainsKey("content-length") && emulatorMode != ApiGatewayEmulatorMode.Rest) // rest doesnt set content-length by default
         {
             headers["content-length"] = contentLength.ToString();
-            multiValueHeaders["content-length"] = new List<string> { contentLength.ToString() };
+            multiValueHeaders["content-length"] = [contentLength.ToString()];
         }
 
         if (!headers.ContainsKey("content-type"))
         {
             headers["content-type"] = "text/plain; charset=utf-8";
-            multiValueHeaders["content-type"] = new List<string> { "text/plain; charset=utf-8" };
+            multiValueHeaders["content-type"] = ["text/plain; charset=utf-8"];
+        }
+
+        // This is the decoded value
+        var path = request.Path.Value;
+
+        if (emulatorMode == ApiGatewayEmulatorMode.HttpV1 || emulatorMode == ApiGatewayEmulatorMode.Rest) // rest and httpv1 uses the encoded value for path an
+        {
+            path = request.Path.ToUriComponent();
+        }
+
+        if (emulatorMode == ApiGatewayEmulatorMode.Rest) // rest uses encoded value for the path params
+        {
+            var encodedPathParameters = pathParameters.ToDictionary(
+                    kvp => kvp.Key,
+                    kvp => Uri.EscapeUriString(kvp.Value)); // intentionally using EscapeURiString over EscapeDataString since EscapeURiString correctly handles reserved characters :/?#[]@!$&'()*+,;= in this case
+            pathParameters = encodedPathParameters;
         }
 
         var proxyRequest = new APIGatewayProxyRequest
         {
             Resource = apiGatewayRouteConfig.Path,
-            Path = request.Path.Value,
+            Path = path,
             HttpMethod = request.Method,
             Body = body,
             IsBase64Encoded = false
@@ -181,13 +201,11 @@ public static class HttpContextExtensions
 
         if (pathParameters.Any())
         {
-            // this should be decoded value
             proxyRequest.PathParameters = pathParameters;
         }
 
         if (HttpRequestUtility.IsBinaryContent(request.ContentType))
         {
-            // we already converted it when we read the body so we dont need to re-convert it
             proxyRequest.IsBase64Encoded = true;
         }
 
--- a/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
+++ b/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
@@ -1,4 +1,7 @@
-﻿using System.Text;
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Text;
 
 namespace Amazon.Lambda.TestTool.Utilities;
 
@@ -32,7 +35,7 @@ public static class HttpRequestUtility
     /// </summary>
     /// <param name="request">The HTTP request.</param>
     /// <returns>The body of the request as a string, or null if the body is empty.</returns>
-    public static string? ReadRequestBody(HttpRequest request)
+    public static async Task<string?> ReadRequestBody(HttpRequest request)
     {
         if (request.ContentLength == 0 || request.Body == null || !request.Body.CanRead)
         {
@@ -46,7 +49,7 @@ public static class HttpRequestUtility
 
         using (var memoryStream = new MemoryStream())
         {
-            request.Body.CopyTo(memoryStream);
+            await request.Body.CopyToAsync(memoryStream);
 
             // If the stream is empty, return null
             if (memoryStream.Length == 0)
@@ -67,7 +70,7 @@ public static class HttpRequestUtility
                 // For text data, read as string
                 using (var reader = new StreamReader(memoryStream))
                 {
-                    string content = reader.ReadToEnd();
+                    string content = await reader.ReadToEndAsync();
                     return string.IsNullOrWhiteSpace(content) ? null : content;
                 }
             }
@@ -80,6 +83,7 @@ public static class HttpRequestUtility
     /// Extracts headers from the request, separating them into single-value and multi-value dictionaries.
     /// </summary>
     /// <param name="headers">The request headers.</param>
+    /// <param name="lowerCaseKeyName">Whether to lowercase the key name or not.</param>
     /// <returns>A tuple containing single-value and multi-value header dictionaries.</returns>
     /// <example>
     /// For headers:
@@ -91,15 +95,16 @@ public static class HttpRequestUtility
     /// singleValueHeaders: { "Accept": "application/xhtml+xml", "X-Custom-Header": "value1" }
     /// multiValueHeaders: { "Accept": ["text/html", "application/xhtml+xml"], "X-Custom-Header": ["value1"] }
     /// </example>
-    public static (IDictionary<string, string>, IDictionary<string, IList<string>>) ExtractHeaders(IHeaderDictionary headers)
+    public static (IDictionary<string, string>, IDictionary<string, IList<string>>) ExtractHeaders(IHeaderDictionary headers, bool lowerCaseKeyName = false)
     {
-        var singleValueHeaders = new Dictionary<string, string>();
-        var multiValueHeaders = new Dictionary<string, IList<string>>();
+        var singleValueHeaders = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+        var multiValueHeaders = new Dictionary<string, IList<string>>(StringComparer.OrdinalIgnoreCase);
 
         foreach (var header in headers)
         {
-            singleValueHeaders[header.Key.ToLower()] = header.Value.Last() ?? "";
-            multiValueHeaders[header.Key.ToLower()] = [.. header.Value];
+            var key = lowerCaseKeyName ? header.Key.ToLower() : header.Key;
+            singleValueHeaders[key] = header.Value.Last() ?? "";
+            multiValueHeaders[key] = [.. header.Value];
         }
 
         return (singleValueHeaders, multiValueHeaders);
@@ -139,7 +144,7 @@ public static class HttpRequestUtility
     /// The generated ID is a 145character string consisting of lowercase letters and numbers, followed by an equals sign.
     public static string GenerateRequestId()
     {
-        return Guid.NewGuid().ToString("N").Substring(0, 8) + Guid.NewGuid().ToString("N").Substring(0, 7) + "=";
+        return $"{Guid.NewGuid().ToString("N").Substring(0, 8)}{Guid.NewGuid().ToString("N").Substring(0, 7)}=";
     }
 
     /// <summary>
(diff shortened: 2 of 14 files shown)


Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/ApiGatewayHttpApiV2ProxyRequestTranslator.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            var (headers, _) = _httpRequestUtility.ExtractHeaders(request.Headers);
+            var (queryStringParameters, _) = _httpRequestUtility.ExtractQueryStringParameters(request.Query);
+
+            var httpApiV2ProxyRequest = new APIGatewayHttpApiV2ProxyRequest
+            {
+                RouteKey = $"{request.Method} {resource}",
+                RawPath = request.Path,
+                RawQueryString = request.QueryString.Value,
+                Cookies = request.Cookies.Select(c => $"{c.Key}={c.Value}").ToArray(),
```
philasmar, line 43: I believe you need to properly encode the cookies.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/ApiGatewayHttpApiV2ProxyRequestTranslator.cs
```diff
@@ -0,0 +1,71 @@
...
+                Headers = headers,
+                QueryStringParameters = queryStringParameters,
+                PathParameters = pathParameters ?? new Dictionary<string, string>(),
```
philasmar, line 46: Could you please extract the path parameters within the translation layer?

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/ApiGatewayProxyRequestTranslator.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            var (headers, multiValueHeaders) = _httpRequestUtility.ExtractHeaders(request.Headers);
+            var (queryStringParameters, multiValueQueryStringParameters) = _httpRequestUtility.ExtractQueryStringParameters(request.Query);
+
+            var proxyRequest = new APIGatewayProxyRequest
+            {
+                Resource = resource,
+                Path = HttpUtility.UrlEncode(request.Path),
+                HttpMethod = request.Method,
+                Headers = headers,
```
philasmar, line 43: You should check if the cookies need special handling. We probably need to extract them, encode them, then set them back

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/ApiGatewayProxyRequestTranslator.cs
```diff
@@ -0,0 +1,61 @@
...
+                MultiValueHeaders = multiValueHeaders,
+                QueryStringParameters = queryStringParameters,
+                MultiValueQueryStringParameters = multiValueQueryStringParameters,
+                PathParameters = pathParameters ?? new Dictionary<string, string>(),
```
philasmar, line 47: same comment about pathparameters

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/ApiGatewayRequestTranslatorFactory.cs
```diff
@@ -0,0 +1,36 @@
+﻿namespace Amazon.Lambda.TestTool
+{
+
+    /// <summary>
+    /// Factory class for creating API Gateway request translators based on the specified API Gateway mode.
+    /// </summary>
+    public class ApiGatewayRequestTranslatorFactory : IApiGatewayRequestTranslatorFactory
```
philasmar, line 7: I think this is a bit overkill. The catch all api will decide which function to call based on the mode if we use extension methods.

philasmar (changes requested):

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/ApiGatewayHttpApiV2ProxyRequestTranslator.cs
gcbeattyAWS: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html it wasnt clear from the docs to me what to do for those. do youhave any real example we can look at easily that you know of?

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/ApiGatewayHttpApiV2ProxyRequestTranslator.cs
gcbeattyAWS: for the path parameter extraction, i think i originally had it inside the translate function. but because i would need to get the matching config too and stuff i took it out and opted for an input parameter.

i will update the code to use extensions to see how that looks first

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/IRouteConfigurationParser.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Amazon.Lambda.TestTool
+{
+    public interface IRouteConfigurationParser
```
gcbeattyAWS, line 3: assuming @philasmar will make the implementation of this interface

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/ApiGatewayHttpApiV2ProxyRequestTranslator.cs
gcbeattyAWS: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes looks like i should encode the cookie values to be safe

gcbeattyAWS: ive added encoding the cookie value in 86b3f5ecbcbd1c0dc1d7f25db6ef1cfb313c12b4

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/ApiGatewayProxyRequestTranslator.cs
gcbeattyAWS: ive updated this logic as well to be safe

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/ApiGatewayHttpApiV2ProxyRequestTranslator.cs
gcbeattyAWS: actually i just checked this again. 

when i logged the cookie in current production aws, i get

```
 headers: {
    'Content-Length': '0',
    Cookie: 'myCookie=special@value!++++',
```
when running with
```
$url = "https://myfunction.amazon.com/testfunction"
$cookieValue = "special@value!++++"

$headers = @{
    "Cookie" = "myCookie=$cookieValue"
}

$response = Invoke-WebRequest -Uri $url -Method GET -Headers $headers

# View the response status code
Write-Host "Status Code: $($response.StatusCode)"

# View the response content
Write-Host "Response Content:"
$response.Content

# Verify the HTTP version used
Write-Host "HTTP Version: $($response.Version)"

```

so it seems that we do not encode url encode it . so i am going to remove the cookie encoding then.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/ApiGatewayHttpApiV2ProxyRequestTranslator.cs
gcbeattyAWS: https://github.com/aws/aws-lambda-dotnet/pull/1902#discussion_r1878517289 from this comment, i will need route config in order to extract the path parameters. so either ill need to re-get it, or just have another abstraction on top of the translation layer which takes in the route config. we can discuss more

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/HttpContextExtensions.cs
```diff
@@ -0,0 +1,113 @@
...
+    /// Translates an <see cref="HttpContext"/> to an <see cref="APIGatewayHttpApiV2ProxyRequest"/>.
+    /// </summary>
+    /// <param name="context">The <see cref="HttpContext"/> to be translated.</param>
+    /// <returns>An <see cref="APIGatewayHttpApiV2ProxyRequest"/> object representing the translated request.</returns>
+    public static APIGatewayHttpApiV2ProxyRequest ToApiGatewayHttpV2Request(
+        this HttpContext context)
+    {
+        var request = context.Request;
+
+        var matchedConfig = _routeConfigurationParser.GetRouteConfig(request.Method, request.Path);
```
gcbeattyAWS, line 36: this should be removed from here or put in some other level of abstraction. will sync with phil today

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ExceptionExtensions.cs
```diff
@@ -1,29 +1,29 @@
-using Amazon.Lambda.TestTool.Models;
```
gcbeattyAWS, line 1: whitespace changes i think are due to phils code not having CRLF line endings. can review the pr with "ignore" whitespace flag setting enabled in meantime

normj (commented):
I don't see any tests for special characters in headers, resource paths or query strings. Have we done a comparison yet with API Gateway? We should have tests for those cases.

gcbeattyAWS:
> I don't see any tests for special characters in headers, resource paths or query strings. Have we done a comparison yet with API Gateway? We should have tests for those cases. That includes the `RawPath` property as well.

so i did some tests with special characters in the header by
1. deploying an api gateway
2. deploying a lambda
3. attaching the lambda to the api gatway
4. invoking the api gateway via command line (curl), and sending a non url encoded header value of `special+++`
5. logging the `event` (apigatewayrequest) object in the lambda itself, and it prints out `special+++` (not encoded), so i assumed no encoding is required for that - but let me double check on this and get back to you

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// Determines whether the specified content type represents binary content.
+    /// </summary>
+    /// <param name="contentType">The content type to check.</param>
+    /// <returns>True if the content type represents binary content; otherwise, false.</returns>
+    public static bool IsBinaryContent(string? contentType)
+    {
+        if (string.IsNullOrEmpty(contentType))
+            return false;
+
+        return contentType.StartsWith("image/", StringComparison.OrdinalIgnoreCase) ||
```
normj, line 18: We have this list of content types we use for response in the ASP.NET Core bridge library. https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.AspNetCoreServer/AbstractAspNetCoreFunction.cs#L59

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
```diff
@@ -0,0 +1,106 @@
...
+        var httpApiV2ProxyRequest = new APIGatewayHttpApiV2ProxyRequest
+        {
+            RouteKey = $"{request.Method} {apiGatewayRouteConfig.Path}",
+            RawPath = request.Path,
+            RawQueryString = request.QueryString.Value,
+            Cookies = request.Cookies.Select(c => $"{c.Key}={c.Value}").ToArray(),
+            Headers = headers,
+            QueryStringParameters = queryStringParameters,
+            PathParameters = pathParameters ?? new Dictionary<string, string>(),
+            Body = HttpRequestUtility.ReadRequestBody(request),
```
normj, line 43: Can you take note that when we get these ToAPIGatewayXXX() methods hooked up to the main code path that is asking for the `APIGatewayHttpApiV2ProxyRequest`. When we convert the `APIGatewayHttpApiV2ProxyRequest` to the MemoryStream to Lambda we should check the size of the stream is no more then 6MB. If so throw back the same error that users would get from API Gateway.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
```diff
@@ -0,0 +1,106 @@
...
+        var proxyRequest = new APIGatewayProxyRequest
+        {
+            Resource = apiGatewayRouteConfig.Path,
+            Path = HttpUtility.UrlEncode(request.Path),
+            HttpMethod = request.Method,
+            Headers = headers,
+            MultiValueHeaders = multiValueHeaders,
+            QueryStringParameters = queryStringParameters,
+            MultiValueQueryStringParameters = multiValueQueryStringParameters,
+            PathParameters = pathParameters ?? new Dictionary<string, string>(),
```
normj, line 93: Why are we initializing this to an empty collection if we don't have any. Is this what API Gateway does?

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
normj: We should also think about how this can be configurable. Probably via environment variables that we can have users configure in Aspire.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// multiValueHeaders: { "Accept": ["text/html", "application/xhtml+xml"], "X-Custom-Header": ["value1"] }
+    /// </example>
+    public static (IDictionary<string, string>, IDictionary<string, IList<string>>) ExtractHeaders(IHeaderDictionary headers)
+    {
+        var singleValueHeaders = new Dictionary<string, string>();
+        var multiValueHeaders = new Dictionary<string, IList<string>>();
+
+        foreach (var header in headers)
+        {
+            singleValueHeaders[header.Key] = header.Value.Last() ?? "";
```
normj, line 59: My memory for API Gateway is that in HTTP V2 where there is just a single headers collection the headers are comma delimited. But we need to verify that. I'm sure it isn't just get the last value.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// multiValueParams: { "param1": ["value1"], "param2": ["value2", "value3"] }
+    /// </example>
+    public static (IDictionary<string, string>, IDictionary<string, IList<string>>) ExtractQueryStringParameters(IQueryCollection query)
+    {
+        var singleValueParams = new Dictionary<string, string>();
+        var multiValueParams = new Dictionary<string, IList<string>>();
+
+        foreach (var param in query)
+        {
+            singleValueParams[param.Key] = param.Value.Last() ?? "";
```
normj, line 85: Same comment as above for ExtractHeaders

normj (changes requested):

normj:
> 4\. invoking the api gateway via command line (curl), and sending a non url encoded header value of `special+++`

@gcbeattyAWS Also check more characters then `+` like spaces and other white space and Unicode characters. resource path and query strings can also be special. Especially when you through `/`, `&` and `?' in the values for resource path segments and query string values. In the ASP.NET Core Lambda bridge library where I have to do all of this in reverse I had to do some encoding handling. https://github.com/aws/aws-lambda-dotnet/blob/3596a58ac6d532695745866a7157896a63f6979c/Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/Utilities.cs#L148

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
gcbeattyAWS: yeah you are right on this https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html 

```
Format 2.0 doesn't have multiValueHeaders or multiValueQueryStringParameters fields. Duplicate headers are combined with commas and included in the headers field. Duplicate query strings are combined with commas and included in the queryStringParameters field.
```

i think i got confused when reading the other docs for v1 https://aws.amazon.com/blogs/compute/support-for-multi-value-parameters-in-amazon-api-gateway/

```
Before this change, API Gateway used to retain only the last value and drop everything else for a multi-valued parameter. You can see the original behavior in the queryStringParameters parameter in the above input, where only the “fish” value is retained.
```

I will update accordingly

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
gcbeattyAWS: good catch. i just did a test
`Invoke-WebRequest -Uri https://myapi-gateway-url.amazonaws.com/testfunction -Method GET`
where it echos the api gateway request object and i see the value is `null` when there are no path parameters. 
```
pathParameters: null,
  stageVariables: null,
  body: null,
  isBase64Encoded: false
}
```

i will update the code

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
gcbeattyAWS: ive updated in fecf91d to handle this

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
gcbeattyAWS: ive updated in fecf91d to handle this

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
gcbeattyAWS: ive updated in fecf91d to handle this

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
gcbeattyAWS: yeah i will connect with phil on where he wants to do this logic at

gcbeattyAWS:
> > 4. invoking the api gateway via command line (curl), and sending a non url encoded header value of `special+++`
> 
> @gcbeattyAWS Also check more characters then `+` like spaces and other white space and Unicode characters. resource path and query strings can also be special. Especially when you through `/`, `&` and `?' in the values for resource path segments and query string values. In the ASP.NET Core Lambda bridge library where I have to do all of this in reverse I had to do some encoding handling.
> 
> https://github.com/aws/aws-lambda-dotnet/blob/3596a58ac6d532695745866a7157896a63f6979c/Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/Utilities.cs#L148

so i have updated the logic based on my findings:1


1. For query strings. `QueryStringParameters` should be the decoded values (this already happens automatically when accessing Request.QueryString.  `RawQueryString` should be the encoded query string without the `?`.
2. Headers do not seem to have any additional logic of decoding that i have seen.
3. The path and path parameters should be decoded (this is also happening automatically now by accessing via request.Path)
4.  The api gatway cannot handle `%2f` in the urls ("/") character

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
```diff
@@ -0,0 +1,161 @@
...
+                    Protocol = request.Protocol
+                },
+                RouteKey = $"{request.Method} {apiGatewayRouteConfig.Path}"
+            },
+            Version = "2.0"
+        };
+
+        if (request.Cookies.Any())
+        {
+            httpApiV2ProxyRequest.Cookies = request.Cookies.Select(c => $"{c.Key}={c.Value}").ToArray();
```
gcbeattyAWS, line 65: headers and cookies are just sent as-is (from my testing)

gcbeattyAWS:
pushing what i have so far

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/HttpContextExtensionsTests.cs
```diff
@@ -0,0 +1,278 @@
...
+
+            CompareHeaders(expected.Headers, actual.Headers);
+            CompareDictionaries(expected.QueryStringParameters, actual.QueryStringParameters);
+            CompareDictionaries(expected.PathParameters, actual.PathParameters);
+            CompareStringArrays(expected.Cookies, actual.Cookies);
+
+            CompareRequestContexts(expected.RequestContext, actual.RequestContext);
+        }
+
+        private void CompareHeaders(IDictionary<string, string> expected, IDictionary<string, string> actual)
```
gcbeattyAWS, line 139: this will go in helper class or similar

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
```diff
@@ -0,0 +1,196 @@
...
+        var body = HttpRequestUtility.ReadRequestBody(request);
+        var contentLength = HttpRequestUtility.CalculateContentLength(request, body);
+
+        var pathParameters = RouteTemplateUtility.ExtractPathParameters(apiGatewayRouteConfig.Path, request.Path);
+
+        // Format 2.0 doesn't have multiValueHeaders or multiValueQueryStringParameters fields. Duplicate headers are combined with commas and included in the headers field.
+        var (_, allHeaders) = HttpRequestUtility.ExtractHeaders(request.Headers);
+        var headers = allHeaders.ToDictionary(
+            kvp => kvp.Key,
+            kvp => string.Join(", ", kvp.Value)
```
philasmar, line 38: nit: small inconsistency between headers and query params. Here you are joining on ", " and below on ","

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
```diff
@@ -0,0 +1,196 @@
...
+        var queryStringParameters = allQueryParams.ToDictionary(
+            kvp => kvp.Key,
+            kvp => string.Join(",", kvp.Value)
+        );
+
+        string userAgent = request.Headers.UserAgent.ToString();
+
+        if (!headers.ContainsKey("content-length"))
+        {
+            headers["content-length"] = contentLength.ToString();
```
philasmar, line 52: Shouldn't we always update the content-length since we are updating the body?

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
```diff
@@ -0,0 +1,196 @@
...
+            RawPath = request.Path.Value, // this should be decoded value
+            Body = body,
+            IsBase64Encoded = false,
+            RequestContext = new ProxyRequestContext
+            {
+                Http = new HttpDescription
+                {
+                    Method = request.Method,
+                    Path = request.Path.Value, // this should be decoded value
+                    Protocol = !string.IsNullOrEmpty(request.Protocol) ? request.Protocol : "HTTP/1.1", // defaults to http 1.1 if not provided
```
philasmar, line 72: should we make the "HTTP/1.1" assumption? This is more likely to get updated and we wouldn't know to update it

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
```diff
@@ -0,0 +1,196 @@
...
+        var request = context.Request;
+        var body = HttpRequestUtility.ReadRequestBody(request);
+        var contentLength = HttpRequestUtility.CalculateContentLength(request, body);
+
+        var pathParameters = RouteTemplateUtility.ExtractPathParameters(apiGatewayRouteConfig.Path, request.Path);
+
+        var (headers, multiValueHeaders) = HttpRequestUtility.ExtractHeaders(request.Headers);
+        var (queryStringParameters, multiValueQueryStringParameters) = HttpRequestUtility.ExtractQueryStringParameters(request.Query);
+
+        if (!headers.ContainsKey("content-length"))
```
philasmar, line 139: Shouldn't we always update the content-length since we are updating the body?

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
philasmar, line 1: Add amazon copyright header

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        // Check if the content is binary
+        bool isBinary = HttpRequestUtility.IsBinaryContent(request.ContentType);
+
+        request.Body.Position = 0;
+
+        using (var memoryStream = new MemoryStream())
+        {
+            request.Body.CopyTo(memoryStream);
```
philasmar, line 49: use async version

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
```diff
@@ -0,0 +1,178 @@
...
+                // For binary data, convert to Base64 string
+                byte[] bytes = memoryStream.ToArray();
+                return Convert.ToBase64String(bytes);
+            }
+            else
+            {
+                // For text data, read as string
+                using (var reader = new StreamReader(memoryStream))
+                {
+                    string content = reader.ReadToEnd();
```
philasmar, line 70: use async version

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
```diff
@@ -0,0 +1,178 @@
...
+    /// <param name="headers">The request headers.</param>
+    /// <returns>A tuple containing single-value and multi-value header dictionaries.</returns>
+    /// <example>
+    /// For headers:
+    /// Accept: text/html
+    /// Accept: application/xhtml+xml
+    /// X-Custom-Header: value1
+    /// 
+    /// The method will return:
+    /// singleValueHeaders: { "Accept": "application/xhtml+xml", "X-Custom-Header": "value1" }
```
philasmar, line 91: Why are we taking the last occurrence of Accept? Is that a valid assumption?

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
```diff
@@ -0,0 +1,178 @@
...
+    /// <summary>
+    /// Extracts query string parameters from the request, separating them into single-value and multi-value dictionaries.
+    /// </summary>
+    /// <param name="query">The query string collection.</param>
+    /// <returns>A tuple containing single-value and multi-value query parameter dictionaries.</returns>
+    /// <example>
+    /// For query string: ?param1=value1&amp;param2=value2&amp;param2=value3
+    /// 
+    /// The method will return:
+    /// singleValueParams: { "param1": "value1", "param2": "value3" }
```
philasmar, line 117: same question here

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
```diff
@@ -0,0 +1,178 @@
...
+
+    /// <summary>
+    /// Generates a random API Gateway request ID for HTTP API v1 and v2.
+    /// </summary>
+    /// <returns>A string representing a random request ID in the format used by API Gateway for HTTP APIs.</returns>
+    /// <remarks>
+    /// The generated ID is a 145character string consisting of lowercase letters and numbers, followed by an equals sign.
+    public static string GenerateRequestId()
+    {
+        return Guid.NewGuid().ToString("N").Substring(0, 8) + Guid.NewGuid().ToString("N").Substring(0, 7) + "=";
```
philasmar, line 142:
```suggestion
        return $"{Guid.NewGuid().ToString("N").Substring(0, 8)} {Guid.NewGuid().ToString("N").Substring(0, 7)}=";
```

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/RouteTemplateUtility.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// <example>
+    /// Using this method:
+    /// <code>
+    /// var routeTemplate = "/users/{id}/orders/{orderId}";
+    /// var actualPath = "/users/123/orders/456";
+    /// var parameters = RouteTemplateUtility.ExtractPathParameters(routeTemplate, actualPath);
+    /// // parameters will contain: { {"id", "123"}, {"orderId", "456"} }
+    /// </code>
+    /// </example>
+    public static Dictionary<string, string> ExtractPathParameters(string routeTemplate, string actualPath)
```
philasmar, line 25: This does not take into account the greedy variable {proxy+}

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/RouteTemplateUtility.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// <returns>A dictionary of default values for the template parameters.</returns>
+    /// <example>
+    /// Using this method:
+    /// <code>
+    /// var template = TemplateParser.Parse("/api/{version=v1}/users/{id}");
+    /// var defaults = RouteTemplateUtility.GetDefaults(template);
+    /// // defaults will contain: { {"version", "v1"} }
+    /// </code>
+    /// </example>
+    public static RouteValueDictionary GetDefaults(RouteTemplate parsedTemplate)
```
philasmar, line 52: This is not needed

philasmar (changes requested):

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
gcbeattyAWS: this is intentional. from my testing the headers to have a space in them but query string params do not

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
gcbeattyAWS: So when api gateway receives a request  with headers

````
Accept: a
Accept: b
```

iand single value header would be `Accept: b`

However, in .net multi value headers are always sent to api gateway like

```
Accept: a, b
```

and singleValue header would be `Accept: a, b`. 


so depending what we want to do we can either do the intended way by api gateway, or do `, '.join(header.value) for the singelvalue header due to the .net limitation

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
gcbeattyAWS: yea thats a good point. it doesnt hurt to always update it

gcbeattyAWS: actually i just checked this. api gateway doesnt set this if the request doesnt contain it already

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
gcbeattyAWS: so i checked this an apparently rest does not set this by default

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
gcbeattyAWS: ive put this here so it matches api gateway

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
gcbeattyAWS: looks like rest api doesnt set this automatically by default

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/HttpContextExtensionsTests.cs
```diff
@@ -0,0 +1,325 @@
...
+        {
+            var route = testCase.ApiGatewayRouteConfig?.Path ?? "/test";
+            var routeKey = testCase.ApiGatewayRouteConfig?.HttpMethod ?? "POST";
+            await _fixture.ApiGatewayHelper.AddRouteToHttpApi(_fixture.ReturnFullEventHttpApiV2Id, _fixture.ReturnFullEventLambdaFunctionArn, "2.0", route, routeKey);
+            await RunApiGatewayTest<APIGatewayHttpApiV2ProxyRequest>(testCase, _fixture.ReturnFullEventHttpApiV2Url, _fixture.ReturnFullEventHttpApiV2Id,
+                async (context, config) => await context.ToApiGatewayHttpV2Request(config), ApiGatewayEmulatorMode.HttpV2);
+        }
+
+        [Fact]
+        public async Task BinaryContentHttpV1()
```
gcbeattyAWS, line 62: we still need an integration test for the rest api for binary content. it was taking too long to setup though because i think need to create a new lambda, api gateway with binary types configured and all that. i can create a backlog item to add this in next pr

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
```diff
@@ -0,0 +1,214 @@
...
+        if (emulatorMode == ApiGatewayEmulatorMode.HttpV1 || emulatorMode == ApiGatewayEmulatorMode.Rest) // rest and httpv1 uses the encoded value for path an
+        {
+            path = request.Path.ToUriComponent();
+        }
+
+        if (emulatorMode == ApiGatewayEmulatorMode.Rest) // rest uses encoded value for the path params
+        {
+            var encodedPathParameters = pathParameters.ToDictionary(
+                    kvp => kvp.Key,
+                    kvp => Uri.EscapeUriString(kvp.Value)); // intentionally using EscapeURiString over EscapeDataString since EscapeURiString correctly handles reserved characters :/?#[]@!$&'()*+,;= in this case
```
gcbeattyAWS, line 167: for example. escapedatastring encodes * as %20 which is different than how api gateway rest api does it (it keeps * as *)

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Extensions/HttpContextExtensionsTests.cs
```diff
@@ -0,0 +1,248 @@
...
+            Assert.Equal(new List<string> { "text/plain; charset=utf-8" }, result.MultiValueHeaders["content-type"]);
+            Assert.Null(result.QueryStringParameters);
+            Assert.Null(result.MultiValueQueryStringParameters);
+            Assert.Null(result.PathParameters);
+        }
+
+        [Theory]
+        [InlineData(ApiGatewayEmulatorMode.Rest)]
+        [InlineData(ApiGatewayEmulatorMode.HttpV1)]
+        public async Task ToApiGateway_MultiValueHeader(ApiGatewayEmulatorMode emulatorMode)
```
gcbeattyAWS, line 110: i manually tested these unit tests

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayIntegrationTestFixture.cs
```diff
@@ -16,17 +16,39 @@ public class ApiGatewayIntegrationTestFixture : IAsyncLifetime
         public ApiGatewayTestHelper ApiGatewayTestHelper { get; private set; }
 
         public string StackName { get; private set; }
-        public string RestApiId { get; private set; }
```
gcbeattyAWS, line 19: this is just renaming

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/RouteTemplateUtility.cs
gcbeattyAWS: this should be fixed now

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
gcbeattyAWS: i checked on this. it needs to be the original content-length which is why i dont set it. my integration test failed when i made it always be set

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/RouteTemplateUtility.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// 1. Converts AWS-style {proxy+} to ASP.NET Core style {*proxy}
+    /// 2. Handles AWS ignoring constraignts by temporarily renaming parameters
+    ///    (e.g., {abc:int} becomes {__aws_param__abc__int})
+    /// </remarks>
+    private static string PreprocessRouteTemplate(string template)
+    {
+        // Convert AWS-style {proxy+} to ASP.NET Core style {*proxy}
+        template = Regex.Replace(template, @"\{(\w+)\+\}", "{*$1}");
+
+        // Handle AWS-style "constraints" by replacing them with temporary parameter names
```
normj, line 81: My understanding is API Gateway does not support constraints like `int` in `{abs:int}`. So if a user gave us a resource path that including constraints when configuring the lambda functions for the emulator it seems like we should throw an error for bad user configuration.

normj: A heads up, as someone that has RegEx dyslexia I generally push back on having anything but the simplest of regexes. I do appreciate these were well commented.

gcbeattyAWS: good question - i actually tried creating a path in api gateway called `{variable:int}` and this was a valid path, with the variable be named the literal `variable:int`. so this code basically makes it so that `variable:int` is correctly parsed from the path as the path parameter key. if we directly used the microsoft template parser, without any modifications, it would think the path parameter key is just `variable` that has an `int` constraint, which is not correct.

gcbeattyAWS: i can update the comment to make it more clear if that helps - aws does not support using constraints, but they let you name the path parameter with same naming convention

normj: Interesting API Gateway allows you to have a variable with a semicolon.

normj (approved):

philasmar (approved):

PR #1902, opened 2024-12-10
feat: add api gateway emulator skeleton

philasmar:
*Issue #, if available:*
DOTNET-7837

*Description of changes:*
* Added an API Gateway Emulator process
* Added new command line arguments for the emulator
* Added a service to process Lambda configuration through environment variables
* Added the ability to match incoming HTTP requests with existing Lambda configurations
* Added a catch-all HTTP endpoint
* Updated base Spectre Command to support Cancellation tokens in order to facilitate unit tests


Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/RunCommand.cs
```diff
@@ -8,49 +8,32 @@
 
 namespace Amazon.Lambda.TestTool.Commands;
```
gcbeattyAWS, line 9: nit: namespace at top to be consistent with others

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/CancellableAsyncCommand.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Runtime.InteropServices;
+
+namespace Spectre.Console.Cli;
+
```
gcbeattyAWS, line 4: can you add docs for these?

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,98 @@
...
+            if (!matcher.TryMatch(path, routeValueDictionary))
+                continue;
+            
+            if (!routeConfig.HttpMethod.Equals(httpMethod, StringComparison.InvariantCultureIgnoreCase))
+                continue;
+
+            return routeConfig;
+        }
+
+        return null;
```
gcbeattyAWS, line 81: should we throw exception here instead of returning null? is there any use case where null would be valid?

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        var app = builder.Build();
+
+        app.UseHttpsRedirection();
+
+        app.MapHealthChecks("/health");
+
+        app.Map("/{**catchAll}", (HttpContext context, IApiGatewayRouteConfigService routeConfigService) =>
+        {
+            var routeConfig = routeConfigService.GetRouteConfig(context.Request.Method, context.Request.Path);
```
gcbeattyAWS, line 51: since routeconfig is already here, thats the reason i didnt try to access it again my translator function originally. we can discuss if we want to pass it in or if i just call the function that you made

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/IApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,18 @@
...
+/// </summary>
+public interface IApiGatewayRouteConfigService
+{
+    /// <summary>
+    /// A method to match an HTTP Method and HTTP Path with an existing <see cref="ApiGatewayRouteConfig"/>.
+    /// </summary>
+    /// <param name="httpMethod">An HTTP Method</param>
+    /// <param name="path">An HTTP Path</param>
+    /// <returns>An <see cref="ApiGatewayRouteConfig"/> corresponding to Lambda function with an API Gateway HTTP Method and Path.</returns>
+    ApiGatewayRouteConfig? GetRouteConfig(string httpMethod, string path);
```
gcbeattyAWS, line 17: i was assuming there would be another function ExtractPathParameters, but i guess i can create a different interface for this. so we should sync on that

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,98 @@
+using System.Collections;
+using System.Text.Json;
+using Amazon.Lambda.TestTool.Models;
+using Amazon.Lambda.TestTool.Services.IO;
+using Microsoft.AspNetCore.Routing.Template;
+
+namespace Amazon.Lambda.TestTool.Services;
+
+/// <inheritdoc cref="IApiGatewayRouteConfigService"/>
+public class ApiGatewayRouteConfigService : IApiGatewayRouteConfigService
```
gcbeattyAWS, line 10: can you add test cases for this class? and any other ones that youve added

gcbeattyAWS (changes requested):

gcbeattyAWS:
can you double check the line file endings. i was getting warnings in my ide that they were not set as CRLF

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/IApiGatewayRouteConfigService.cs
gcbeattyAWS: we talked offline. i will create this function

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs
gcbeattyAWS: we talked offline on this. i will pass routeconfig into my function

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs
```diff
@@ -0,0 +1,78 @@
...
+/// </summary>
+public class ApiGatewayEmulatorProcess
+{
+    /// <summary>
+    /// The service provider that will contain all the registered services.
+    /// </summary>
+    public required IServiceProvider Services { get; init; }
+
+    /// <summary>
+    /// The API Gatewat emulator task that was started.
```
normj, line 19: typo `Gatewat`

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs
```diff
@@ -0,0 +1,78 @@
...
+        builder.WebHost.UseUrls(serviceUrl);
+        builder.WebHost.SuppressStatusMessages(true);
+
+        builder.Services.AddHealthChecks();
+
+        var app = builder.Build();
+
+        app.UseHttpsRedirection();
+
+        app.MapHealthChecks("/health");
```
normj, line 47: If we want/need this we have to use a less common resource path to not interfere with any `/health` resource paths being emulated by the Lambda functions. Something like `/__lambda_test_tool_apigateway_health__`. But given we are web app for running just locally do really need this?

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        app.Map("/{**catchAll}", (HttpContext context, IApiGatewayRouteConfigService routeConfigService) =>
+        {
+            var routeConfig = routeConfigService.GetRouteConfig(context.Request.Method, context.Request.Path);
+            if (routeConfig == null)
+            {
+                return Results.NotFound("Route not found");
```
normj, line 54: Testing API Gateway I see if the is no mapped route it returns a `403` with a response body of `{"message":"Missing Authentication Token"}`. Also the header `x-amzn-errortype` is set to `MissingAuthenticationTokenException`.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,98 @@
...
+{
+    private readonly List<ApiGatewayRouteConfig> _routeConfigs = new();
+
+    /// <summary>
+    /// Constructs an instance of <see cref="ApiGatewayRouteConfigService"/>
+    /// which loads and parses environment variables that match a specific prefix.
+    /// </summary>
+    /// <param name="environmentManager">A service to access environment variables.</param>
+    public ApiGatewayRouteConfigService(
```
normj, line 19: Take in a ILogger and use debug logging through here. Getting the config wrong is likely going to be a place of problems and we will want some mechanism to asks users to enable logging so we can see what is going on.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,98 @@
...
+            if (!(key.Equals(Constants.LambdaConfigEnvironmentVariablePrefix) || 
+                key.StartsWith($"{Constants.LambdaConfigEnvironmentVariablePrefix}_")))
+                continue;
+
+            var jsonValue = entry.Value?.ToString();
+            if (string.IsNullOrEmpty(jsonValue))
+                continue;
+            try
+            {
+                var config = JsonSerializer.Deserialize<ApiGatewayRouteConfig>(jsonValue);
```
normj, line 38: Can we check if the jsonValue starts with `[` and if so deserialize as a list instead of using exceptions for control flow?

normj (changes requested):

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/CancellableAsyncCommand.cs
philasmar: Done

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: From a separation of concerns perspective, this function is trying to match routes. If a route doesn't match, I don't want it to blow up. It should indicate that the route doesn't match, which I am using null for. It is the responsibility of the caller to throw an exception if needed.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: Done. I added wrappers for Environment and Directory. I am not going to add tests for those. They are 1:1 mapping of the system classes.

philasmar:
> can you double check the line file endings. i was getting warnings in my ide that they were not set as CRLF

Since we are working on different platforms we're bound to run into some line ending issues. I think what we did with the deploy tool was update our git config to retrieve files as LF. (something like that)

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Constants.cs
```diff
@@ -1,23 +1,94 @@
...
 {
+    /// <summary>
+    /// The name of the dotnet CLI tool
+    /// </summary>
     public const string ToolName = "lambda-test-tool";
+
+    /// <summary>
+    /// The default port used by the Lambda Test Tool for the Lambda Runtime API and the Web Interface.
+    /// </summary>
     public const int DefaultPort = 5050;
```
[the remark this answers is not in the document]
philasmar: Done

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs
philasmar: The health check is needed purely for testing. I am using it to make sure that the process went up before I can cancel it. I renamed it to the name you suggested.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs
philasmar: Let's sync up on this

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: Done

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,135 @@
...
+            }
+
+            var jsonValue = entry.Value?.ToString();
+            logger.LogDebug("Environment variable value: {VariableValue}", jsonValue);
+            if (string.IsNullOrEmpty(jsonValue))
+                continue;
+
+            try
+            {
+                if (jsonValue.StartsWith('['))
```
gcbeattyAWS, line 51: can you double check if this string needs to have whitespace trimmed in the beginning?

```
 [
 //rest of json
]
```

i believe whitespace is technically ignored in json, so if theres a leading whitespace then this would fail 
```

Comment (philasmar) [reply Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs:51]: Good catch. Fixed.

File: Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs (hunk 9 above)
Comment (philasmar) [reply Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs]: Fixed

File: Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs [hunk 14]
```diff
@@ -0,0 +1,80 @@
...
+        app.UseHttpsRedirection();
+
+        app.MapHealthChecks("/__lambda_test_tool_apigateway_health__");
+
+        app.Map("/{**catchAll}", (HttpContext context, IApiGatewayRouteConfigService routeConfigService) =>
+        {
+            var routeConfig = routeConfigService.GetRouteConfig(context.Request.Method, context.Request.Path);
+            if (routeConfig == null)
+            {
+                context.Response.StatusCode = StatusCodes.Status403Forbidden;
```
normj, line 54: Use an ILogger that we failed to find a configured Lambda route for the specified method and path. I would log at the information level.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                // TODO: Translate to APIGatewayProxyRequest
+            }
+
+            return Results.Ok();
+        });
+
+        var runTask = app.RunAsync(cancellationToken);
+        
+        return new ApiGatewayEmulatorProcess
```
normj, line 73: At this point do a `LogInformation` that the API Gateway emulator is available at the serviceUrl endpoint. Imagine this is what non-Aspire users will learn where to send requests to.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    /// <inheritdoc />
+    public ApiGatewayRouteConfig? GetRouteConfig(string httpMethod, string path)
+    {
+        foreach (var routeConfig in _routeConfigs)
+        {
+            _logger.LogDebug("Checking if '{Path}' matches '{Template}'.", path, routeConfig.Path);
+            var template = TemplateParser.Parse(routeConfig.Path);
```
normj, line 94: I don't see any tests for handling API Gateway's style of using `{proxy+}` to indicate wild card character. I really want to know if Microsoft's `TemplateParser` is handling that or if we need to roll our own solution. Have you done any side by side comparisons with API Gateway and this library to see if they resolve the same.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,135 @@
...
+            
+            _logger.LogDebug("{HttpMethod} {Path} matches the existing configuration {TemplateMethod} {Template}.", httpMethod, path, routeConfig.HttpMethod, routeConfig.Path);
+
+            return routeConfig;
+        }
+
+        return null;
+    }
+
+    private RouteValueDictionary GetDefaults(RouteTemplate parsedTemplate)
```
normj, line 121: Are defaults even a think in API Gateway routing? I thought with API Gateway routing you can just have the name of the variable. I could be wrong but be sure we are not just implementing a feature of the Microsoft library and not API Gateway.

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Services/ApiGatewayRouteConfigServiceTests.cs
```diff
@@ -0,0 +1,158 @@
+using System.Text.Json;
+using Amazon.Lambda.TestTool.Models;
+using Amazon.Lambda.TestTool.Services;
+using Amazon.Lambda.TestTool.Services.IO;
+using Microsoft.Extensions.Logging;
+using Moq;
+
+namespace Amazon.Lambda.TestTool.UnitTests.Services;
+
+public class ApiGatewayRouteConfigServiceTests
```
normj, line 10: You need tests with `{proxy+}` especially when there are other routes. And you need tests with some overlap in the routes. For example a test with routeA -> `/foo/{value}` and routeB -> `/foo/bar` and you test with requests to `/foo`, `/foo/bar` and `/foo/fred`. Similar idea with `{proxy+}` and make sure we mimic API Gateway's behavior when choosing to use the proxy endpoint versus a more direct path. Another example routeA `/{proxy+}` and routeB `/phil`. Requests going to `/norm` and `/phil` and `/phil/dotnet`. Don't expect the Microsoft template library to match API Gateway's behavior.

normj (changes requested):

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs
philasmar: This is a bit redundant since the dotnet process will print the following to the console:
```
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: **http://localhost:5051**
dbug: Microsoft.AspNetCore.Hosting.Diagnostics[13]
      Loaded hosting startup assembly Amazon.Lambda.TestTool
info: Microsoft.Hosting.Lifetime[0]
      Application started. Press Ctrl+C to shut down.
info: Microsoft.Hosting.Lifetime[0]
      Hosting environment: Development
info: Microsoft.Hosting.Lifetime[0]
      Content root path: /Users/asmarp/CodeBase/aws-lambda-dotnet/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool
dbug: Microsoft.Extensions.Hosting.Internal.Host[2]
      Hosting started
```

philasmar: Also, at this point in time, the process may not successfully start. So if I was to print the service URL and then the task failed, it would be kind of weird. I'd prefer if we rely on the default .NET logging for this.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs
philasmar: Done

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: TemplateParser supports a catch-all variable similar to the greedy variable in API gateway. I have updated the implementation to account for it as well as add sorting based on the route priority logic from API Gateway. I have added comments to explain everything in the code.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: You are right. The function was not doing anything for our use case. I removed it.

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Services/ApiGatewayRouteConfigServiceTests.cs
philasmar: I have updated the implementation to account for it as well as add sorting based on the route priority logic from API Gateway. I have added comments to explain everything in the code.
I also added 1 big test that lists a bunch of different routes with overlap and tests which function was matched. This should cover everything.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs
normj: The ASP.NET Core log line below doesn't give users context that this is the API Gateway emulator versus Lambda runtime API.
```
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: **http://localhost:5051**
```

I believe in the published version of the tool we will default all of the Microsoft logging default level to `Error` so that our console window is focused on giving logging messages focused on our tool. When users see the console of our tool that should be thinking this is a log message for a lambda emulator tool not a web server.

So we need somewhere in the output that says basically the API Gateway Emulator endpoint is here. If this isn't that right spot that is fine but needs to be somewhere. 

Imagine a user installing who just does `dotnet install lambda-test-tool` and `dotnet lambda-test-tool`. There mind set should only be how to I test my Lambda function and what is informative to me as a Lambda developer debugging. Not how to a debug some webserver, that is our implementation detail.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,221 @@
...
+    /// 4. ANY /{proxy+}
+    /// 
+    /// For more info: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-routes.html
+    /// </summary>
+    private void UpdateRouteConfigMetadataAndSorting()
+    {
+        _logger.LogDebug("Updating the metadata needed to properly sort the Lambda config");
+        foreach (var routeConfig in _routeConfigs)
+        {
+            if (routeConfig.Path.Contains("{proxy+}"))
```
normj, line 117: What if you have a resource path like `/resource/{id}/subsegment/{proxy+}` so you have both a proxy and a variable?

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Services/ApiGatewayRouteConfigServiceTests.cs
```diff
@@ -0,0 +1,256 @@
...
+
+        // Assert
+        Assert.NotNull(result1);
+        Assert.NotNull(result2);
+        Assert.Equal("Function1", result1.LambdaResourceName);
+        Assert.Equal("Function2", result2.LambdaResourceName);
+    }
+
+    [Fact]
+    public void ProperlySortRouteConfigs()
```
normj, line 160: Do you verify with an API Gateway deployed with routes like this that returns behaves the same?

philasmar: Yes the values I have in the test have been verified in API Gateway

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    /// <summary>
+    /// Applies some validity checks for Lambda route configuration.
+    /// </summary>
+    /// <param name="routeConfig">Lambda route configuration</param>
+    /// <returns>true if route is valid, false if not</returns>
+    private bool IsRouteConfigValid(ApiGatewayRouteConfig routeConfig)
+    {
+        var occurrences = routeConfig.Path.Split("{proxy+}").Length - 1;
```
normj, line 121: Validate that path, httpmethod and Lambda name are set.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,430 @@
...
+                {
+                    literalMatches++;
+                    if (!encounteredGreedy) matchedSegmentsBeforeGreedy++;
+                    routeTemplateIndex++;
+                    requestPathIndex++;
+                }
+            }
+        }
+
+        // If there are leftover route segments
```
normj, line 324: Can you explain what would be an example route and request path that would go into this section. It would be good to add examples in the comment block to help readers. I'm struggling to read this and understand in what scenario would the route ever be a matched if there are more route segments then were in the request path.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,430 @@
...
+            else
+            {
+                // Literal not matched
+                matched = false;
+                routeTemplateIndex++;
+            }
+        }
+
+        // If request has leftover segments that aren't matched
+        if (matched && requestPathIndex < requestSegments.Length)
```
normj, line 364: If the `variableCount` doesn't match the variables found in the request path should that be an false match.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: This is now handled

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/ApiGatewayEmulatorProcess.cs
philasmar: I added an informational message when the application starts using:
```
app.Lifetime.ApplicationStarted.Register(() =>
{
    app.Logger.LogInformation("The API Gateway Emulator is available at: {ServiceUrl}", serviceUrl);
});
```

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Models/ApiGatewayRouteType.cs
```diff
@@ -0,0 +1,22 @@
+namespace Amazon.Lambda.TestTool.Models;
+
+/// <summary>
+/// The type of API Gateway Route. This is used to determine the priority of the route when there is route overlap.
+/// </summary>
+public enum ApiGatewayRouteType
```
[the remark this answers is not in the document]
philasmar: Deleted

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: Done

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: I simplified this code paths as it was redundant. This was a leftover from my tweaking when I was working on it. I also added comments and examples explaining the code paths as much as possible.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: If we enter these code paths, then matched will be false and variableCount won't matter.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,452 @@
...
+    /// <summary>
+    /// Applies some validity checks for Lambda route configuration.
+    /// </summary>
+    /// <param name="routeConfig">Lambda route configuration</param>
+    /// <returns>true if route is valid, false if not</returns>
+    private bool IsRouteConfigValid(ApiGatewayRouteConfig routeConfig)
+    {
+        if (string.IsNullOrEmpty(routeConfig.LambdaResourceName))
+        {
+            _logger.LogDebug("The Lambda resource name cannot be empty for the route config {Method} {Path}.",
```
normj, line 123: Change `LogDebug` to `LogError` since the input they gave us was wrong.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,452 @@
...
+                routeConfig.HttpMethod, routeConfig.Path);
+            return false;
+        }
+
+        if (string.IsNullOrEmpty(routeConfig.HttpMethod))
+        {
+            _logger.LogDebug("The HTTP Method cannot be empty for the route config with the Lambda resource name {Lambda}.",
```
normj, line 130: Change `LogDebug` to `LogError` since the input they gave us was wrong.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,452 @@
...
+                routeConfig.LambdaResourceName);
+            return false;
+        }
+
+        if (string.IsNullOrEmpty(routeConfig.Path))
+        {
+            _logger.LogDebug("The HTTP Path cannot be empty for the route config with the Lambda resource name {Lambda}.",
```
normj, line 137: Change `LogDebug` to `LogError` since the input they gave us was wrong.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,452 @@
...
+                routeConfig.LambdaResourceName);
+            return false;
+        }
+        
+        var occurrences = routeConfig.Path.Split("{proxy+}").Length - 1;
+        if (occurrences > 1)
+        {
+            _logger.LogDebug("The route config {Method} {Path} cannot have multiple greedy variables {{proxy+}}.",
```
normj, line 145: Change `LogDebug` to `LogError` since the input they gave us was wrong.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -0,0 +1,452 @@
...
+        if (occurrences > 1)
+        {
+            _logger.LogDebug("The route config {Method} {Path} cannot have multiple greedy variables {{proxy+}}.", 
+                routeConfig.HttpMethod, routeConfig.Path);
+            return false;
+        }
+
+        if (occurrences == 1 && !routeConfig.Path.EndsWith("/{proxy+}"))
+        {
+            _logger.LogDebug("The route config {Method} {Path} uses a greedy variable {{proxy+}} but does not end with it.",
```
normj, line 152: Change `LogDebug` to `LogError` since the input they gave us was wrong.

normj (approved):
I ask that you change some `LogDebug` to `LogError` because they represent invalid user data. Other then that I'm good. I don't need to rereview for the `LogError` change.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: Done

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: Done

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: Done

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: Done

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: Done

PR #1903, opened 2024-12-10
Api gateway response parsing

gcbeattyAWS:
*Issue #, if available:*
DOTNET-7839

*Description of changes:*
1. Converts api gateway response to httpresonse


normj (commented):
API Gateway sets common headers on the response. Like a request id. We should mimic those as much as possible.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+    /// <param name="response">The HttpResponse object to modify.</param>
+    /// <param name="headers">The single-value headers to set.</param>
+    /// <param name="multiValueHeaders">The multi-value headers to set.</param>
+    private static void SetResponseHeaders(HttpResponse response, IDictionary<string, string>? headers, IDictionary<string, IList<string>>? multiValueHeaders = null)
+    {
+        if (headers != null)
+        {
+            foreach (var header in headers)
+            {
+                response.Headers[header.Key] = header.Value;
```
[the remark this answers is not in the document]
gcbeattyAWS: yeah you are right. i just checked by making a function that returns

```
return {
        statusCode: 200,
        headers: {
            'Content-Type': 'application/json',
            'myheader': 'test,shouldhavesecondvalue',
        },
```

and _curling_ it. and the response header only shows `myheader: shouldhavesecondvalue`. in my current code, i beleive it would've been  'test,shouldhavesecondvalue which is not correct. It should be the last header value for that key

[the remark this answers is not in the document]
gcbeattyAWS: lambda code

```
 return {
        statusCode: 200,
        headers: {
            'Content-Type': 'application/json',
            'myheader': 'test,shouldhavesecondvalue',
            'anotherheader': 'firstvalue',
            'anotherheader': 'secondvalue'
        },
        multiValueHeaders: {
            "headername": ["headervalue", "headervalue2"]
        },
        body: JSON.stringify({
            event
        }),    };
};
```

V1 Response headers
```
Myheader: test,shouldhavesecondvalue
Anotherheader: secondvalue
Headername: headervalue
Headername: headervalue2
```

v2 response headers

```
Anotherheader: secondvalue
Myheader: test,shouldhavesecondvalue
```

so i think my code is correct then no?

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+    /// <param name="apiResponse">The API Gateway proxy response to convert.</param>
+    /// <returns>An HttpResponse object representing the API Gateway response.</returns>
+    public static HttpResponse ToHttpResponse(this APIGatewayProxyResponse apiResponse)
+    {
+        var httpContext = new DefaultHttpContext();
+        var response = httpContext.Response;
+
+        response.StatusCode = apiResponse.StatusCode;
+
+        SetResponseHeaders(response, apiResponse.Headers, apiResponse.MultiValueHeaders);
```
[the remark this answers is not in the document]
gcbeattyAWS: i updated to set content length. for the default content-type, theres actually a couple of cases for how that can get set. ive written test cases for those and it may be easier to read them than me typing :D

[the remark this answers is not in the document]
gcbeattyAWS: so i dug more into this. this was tripping me up for a couple of hours. the default content type actually depends on what kind of api gateway is being used. If content-type is not specified and a REST api is being used, the content type is application/.json. but if its an http api being used (for v1) its text/plain. but then a http (v2) can sometimes infer the application type is json...

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Extensions/ApiGatewayResponseExtensionsTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            Assert.Equal("application/json", httpResponse.ContentType);
+
+            httpResponse.Body.Seek(0, SeekOrigin.Begin);
+            using var reader = new StreamReader(httpResponse.Body);
+            var bodyContent = reader.ReadToEnd();
+            Assert.Equal(jsonBody, bodyContent);
+        }
+
+        [Fact]
+        public void ToHttpResponse_APIGatewayHttpApiV2ProxyResponse_HandlesNonJsonResponse()
```
gcbeattyAWS, line 90: im not actually convinced this test case is right since i have not been able to return a non json response from a lambda (lol). Even if i return this `{fdasfdsa}`, it thinks the string is valid json. 

The docs are not really clear either
https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html

With the 2.0 format version, API Gateway can infer the response format for you. API Gateway makes the following assumptions if your Lambda function returns valid JSON and doesn't return a statusCode:

isBase64Encoded is false.

statusCode is 200.

content-type is application/json.

body is the function's response.


from the examples they have, they even consider a regular string to be valid. so im thinking i need to update the logic with the json validator/remove it (see other comment)

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,276 @@
...
+        {
+            response.ContentType = "text/plain; charset=utf-8"; // api gateway v2 defaults to this content type if none is provided
+        }
+
+        if (statusCode != 0)
+        {
+            response.StatusCode = statusCode;
+        }
+        // v2 tries to automatically make some assumptions if the body is valid json
+        else if (IsValidJson(body))
```
gcbeattyAWS, line 224: this logic here - i havent been able to find a case where returning something from a lambda is considered _not_ json. even doing `callback(null, 1) or `callback(null, "fdasfdasfads") in the lambda, still thinks this response type is json. so im thinking maybe we should actually just check "is body present and if it is, set as json"

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayResponseExtensionsSpecialCases.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            await _helper.AssertResponsesEqual(actualResponse, httpTestResponse);
+            Assert.Equal(200, (int)actualResponse.StatusCode);
+            Assert.Equal("application/json", actualResponse.Content.Headers.ContentType.ToString());
+            var content = await actualResponse.Content.ReadAsStringAsync();
+            Assert.Equal("{\"key\" : \"value\"}", content);
+        }
+
+        [Fact]
+        public async Task V2_HandlesNonJsonResponse()
```
gcbeattyAWS, line 95: this test case actually fails as of now - in reality it still returns "{key}". ive kept it here to comment on how i dont really understand api gateways behavior in this case

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayTestHelper.cs
```diff
@@ -0,0 +1,297 @@
...
+            var actualContent = await actualResponse.Content.ReadAsStringAsync();
+            Assert.Equal(expectedContent, actualContent);
+
+            Assert.Equal(httpTestResponse.StatusCode, (int)actualResponse.StatusCode);
+
+            var headersToIgnore = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+            {
+                "Date",
+                "Apigw-Requestid",
+                "X-Amzn-Trace-Id",
```
gcbeattyAWS, line 232: im planning on setting some of these. i will add in the next couple of commits

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        return response;
+    }
+
+    /// <summary>
+    /// Converts an APIGatewayHttpApiV2ProxyResponse to an HttpResponse.
+    /// </summary>
+    /// <param name="apiResponse">The API Gateway HTTP API v2 proxy response to convert.</param>
+    /// <returns>An HttpResponse object representing the API Gateway response.</returns>
+    public static HttpResponse ToHttpResponse(this APIGatewayHttpApiV2ProxyResponse apiResponse)
```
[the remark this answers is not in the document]
gcbeattyAWS: so the 6mb part is actually a little tricky. the reason being is that the lambda will return something like

```
{
  "StatusCode": 200,
  "Headers": {
    "Content-Type": "application/json",
    "Custom-Header": "Some Value"
  },
  "Body": "This is the response body",
  "IsBase64Encoded": false
}

```

and then from our end we will have the already deserialized object of APIGatewayHttpApiV2ProxyResponse with that contents. so we need to check the size of the actual serialized APIGatewayHttpApiV2ProxyResponse  object represented in json form right? if thats the case then that gets tricky, because then how do we handle null values?

if we include them theres no way to know if the user sent back

```
Cookies = null
```

in the response (which would be part of the payload size) or if they did not set it (and the payload size would actually be less(\).

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,326 @@
...
+        return response;
+    }
+    /// <summary>
+    /// Sets the response headers on the HttpResponse, including default API Gateway headers based on the emulator mode.
+    /// </summary>
+    /// <param name="response">The HttpResponse to set headers on.</param>
+    /// <param name="headers">The single-value headers to set.</param>
+    /// <param name="multiValueHeaders">The multi-value headers to set.</param>
+    /// <param name="emulatorMode">The API Gateway emulator mode determining which default headers to include.</param>
+    private static void SetResponseHeaders(HttpResponse response, IDictionary<string, string>? headers, IDictionary<string, IList<string>>? multiValueHeaders = null, ApiGatewayEmulatorMode emulatorMode = ApiGatewayEmulatorMode.HttpV2)
```
normj, line 61: Having a default for `emulatorMode` seems odd to me. Defaulting the value to HttpV2 seems random.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,326 @@
...
+    /// </summary>
+    /// <param name="response">The HttpResponse to set the body on.</param>
+    /// <param name="body">The body content.</param>
+    /// <param name="isBase64Encoded">Whether the body is Base64 encoded.</param>
+    private static void SetResponseBody(HttpResponse response, string? body, bool isBase64Encoded, ApiGatewayEmulatorMode apiGatewayEmulator)
+    {
+        if (!string.IsNullOrEmpty(body))
+        {
+            byte[] bodyBytes;
+            if (isBase64Encoded && ApiGatewayEmulatorMode.Rest != apiGatewayEmulator)
```
normj, line 181: Why not base 64 decided for `Rest`? That seems valid for a Rest Lambda function to return the response base 64 encoded.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,326 @@
...
+        else
+        {
+            if (emulatorMode == ApiGatewayEmulatorMode.Rest) // rest api text for this message/error code is slightly different
+            {
+                response.StatusCode = 502;
+                response.ContentType = "application/json";
+                var errorBytes = Encoding.UTF8.GetBytes("{\"message\": \"Internal server error\"}");
+                response.Body = new MemoryStream(errorBytes);
+                response.ContentLength = errorBytes.Length;
+            } else
```
normj, line 244: nit: Please put the `else` on a separate line.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,326 @@
...
+        if (statusCode != 0)
+        {
+            response.StatusCode = statusCode;
+        }
+        // v2 tries to automatically make some assumptions if the body is valid json
+        else if (IsValidJson(body))
+        {
+            // API Gateway 2.0 format version assumptions
+            response.StatusCode = 200;
+            response.ContentType = "application/json";
```
normj, line 278: Should you check to see if content type was already set from the headers collection coming back from the Lambda function?

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayResponseExtensionsJsonInference.cs
```diff
@@ -0,0 +1,115 @@
...
+                console.log(event);
+                callback(null, event.body);
+            };";
+            _lambdaArn = await _helper.CreateLambdaFunctionAsync(_roleArn, lambdaCode);
+
+            // Create HTTP API (v2)
+            (_httpApiV2Id, _httpApiV2Url) = await _helper.CreateHttpApi(_lambdaArn, "2.0");
+
+            // Wait for the API Gateway to propagate
+            await Task.Delay(10000); // Wait for 10 seconds
```
normj, line 50: Instead of a sleep can we ping the endpoint till we get a Http StatusCode 200 or it has been 10 seconds?

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayTestHelper.cs
```diff
@@ -0,0 +1,300 @@
...
+            IAmazonIdentityManagementService iamClient)
+        {
+            _apiGatewayV1Client = apiGatewayV1Client;
+            _apiGatewayV2Client = apiGatewayV2Client;
+            _lambdaClient = lambdaClient;
+            _iamClient = iamClient;
+            _httpClient = new HttpClient();
+        }
+
+        public async Task<string> CreateLambdaFunctionAsync(string roleArn, string lambdaCode)
```
normj, line 44: Please create the function with a CloudFormation template so it is easier to tear down when we have orphan test data when tests crash.

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Extensions/ApiGatewayResponseExtensionsTests.cs
normj: To be clear you are saying you can't return a non-json response when the status code is not set correct? I have definitely seen non-JSON response with Http Api V2 but that I'm sure has been with the status code set.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: when i tested with the Rest api gateway - the response doesnt actually get decoded from base 64. it does get decoded when i use httpv1 or httpv2 though

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Extensions/ApiGatewayResponseExtensionsTests.cs
gcbeattyAWS: i mean i cant figure out how to return a non json response from a lamda function. if i do something like

```

return {}
```

this is valid json. but even doing 

```
return "abc"
```

from the lambda, api gateway thinks this is valid json.

and i have also tried (in javascript) doing

```
callback(1234)
```

do send the direct value, but the content type is still set as application/json

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: ill add an integration test case to confirm what happens in that scenario and update accordingly

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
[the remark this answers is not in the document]
gcbeattyAWS: also from my testing, this size check happens at the lambda layer (at least the 6mb one does) so we may want to have this in another spot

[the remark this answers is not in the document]
normj: I agree we will need to add the check elsewhere.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
normj: Not sure what you mean. It is up to the Lambda function itself to base 64 encode what it sets to the `Body` property of the `APIGatewayProxyResponse`. There is nothing at the service layer that automatically encodes the body response.

gcbeattyAWS: right. but what i mean was i made a lambda function with 

```
export const handler = async (event, context) => {
    return {
        statusCode: 200,
        body: "dGVzdA==",
        isBase64Encoded: true
    };
};
```

and connected an api gateway to it (both rest and http). when i invoke the api gateway url for the http one, the response is `test`. when i invoke the rest api, the response is `dGVzdA==`. so it seems like in rest api case, api gateway does not automatically decode the response

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: ive updated my test case  locally. so in the case of inferring the json response, it doesnt use whats in the headers.

```
  [Fact]
  public async Task V2_TestCanInferJsonType2()
  {

      var testResponse = new APIGatewayHttpApiV2ProxyResponse
      {
          Body = "{\"key\" : \"value\"}",
          Headers = new Dictionary<string, string>
              {
                  { "Content-Type", "application/xml" },
              },
      };

      var httpTestResponse = testResponse.ToHttpResponse();
      var actualResponse = await _httpClient.PostAsync(_httpApiV2Url, new StringContent("{\"key\" : \"value\"}"));

      await _helper.AssertResponsesEqual(actualResponse, httpTestResponse);
      Assert.Equal(200, (int)actualResponse.StatusCode);
      Assert.Equal("application/json", actualResponse.Content.Headers.ContentType.ToString());
      var content = await actualResponse.Content.ReadAsStringAsync();
      Assert.Equal("{\"key\" : \"value\"}", content);
  }
```

this test case passed when i ran it

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayTestHelper.cs
gcbeattyAWS: i would like to update it to use cloudformation in a separate PR, after i get the main functionality in.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: fixed in latest revision

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: fixed in latest revision

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayResponseExtensionsJsonInference.cs
gcbeattyAWS: fixed in latest revision

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: ive added `V2_TestJsonInfersAlwaysUsesApplicationJson`. i believe my current logic is right. the reason being is for this scenario, since status code is 0, api gateway just uses the whole payload as the body response and doesnt consider any lambda provided headers.

gcbeattyAWS:
@normj we can discuss more on 12/18 on the application/json default logic when the status code is 0 for http v2. i feel that is the last big thing functionality wise missing. so far i have been able to test when the status code is 0 and the body is valid json. but i have not been able to figure out how to verify what happens when the status code is 0 and the body is _not_ valid json

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+{
+
+    private const string InternalServerErrorMessage = "{\"message\":\"Internal Server Error\"}";
+
+    /// <summary>
+    /// Converts an APIGatewayProxyResponse to an HttpResponse.
+    /// </summary>
+    /// <param name="apiResponse">The API Gateway proxy response to convert.</param>
+    /// <returns>An HttpResponse representing the API Gateway response.</returns>
+    public static HttpResponse ToHttpResponse(this APIGatewayProxyResponse apiResponse, ApiGatewayEmulatorMode emulatorMode)
```
philasmar, line 27: Returning an `HttpResponse` will make it hard for us to call this extension method from the minimal API and then return the new `HttpResponse`. The `HttpResponse` is tightly coupled with the `HttpContext` and cannot be updated as it does not have a setter. We will need to update the existing `HttpResponse` in-place.
The method should look like:
```
public static void ApplyToHttpResponse(
    this APIGatewayProxyResponse apiResponse, 
    HttpContext context, 
    ApiGatewayEmulatorMode emulatorMode)
```

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+    {
+        var httpContext = new DefaultHttpContext();
+        var response = httpContext.Response;
```
philasmar, line 30: Do not do this. Use the existing HttpContext.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+
+        return response;
+    }
+
+    /// <summary>
+    /// Converts an APIGatewayHttpApiV2ProxyResponse to an HttpResponse.
+    /// </summary>
+    /// <param name="apiResponse">The API Gateway HTTP API v2 proxy response to convert.</param>
+    /// <returns>An HttpResponse representing the API Gateway response.</returns>
+    public static HttpResponse ToHttpResponse(this APIGatewayHttpApiV2ProxyResponse apiResponse)
```
philasmar, line 44: similar comment to above

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+    /// <summary>
+    /// Sets the response headers on the HttpResponse, including default API Gateway headers based on the emulator mode.
+    /// </summary>
+    /// <param name="response">The HttpResponse to set headers on.</param>
+    /// <param name="headers">The single-value headers to set.</param>
+    /// <param name="multiValueHeaders">The multi-value headers to set.</param>
+    /// <param name="emulatorMode">The API Gateway emulator mode determining which default headers to include.</param>
+    private static void SetResponseHeaders(HttpResponse response, IDictionary<string, string>? headers, ApiGatewayEmulatorMode emulatorMode, IDictionary<string, IList<string>>? multiValueHeaders = null)
+    {
+        var processedHeaders = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
philasmar, line 64: If you;re using the HttpResponse from the API, you might need to clear the existing headers first

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+                response.Headers[header.Key] = new StringValues(header.Value.ToArray());
+                processedHeaders.Add(header.Key);
+            }
+        }
+
+        if (headers != null)
+        {
+            foreach (var header in headers)
+            {
+                if (!processedHeaders.Contains(header.Key))
```
philasmar, line 87: Why do you need to use a separate set to track existing keys when you can just use: 
```
response.Headers.ContainsKey(header.Key)
```

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+        var guid2 = Guid.NewGuid().ToString("N");
+        return $"Root=1-{timestamp}-{guid1.Substring(0, 12)}{guid2.Substring(0, 12)};Parent={Guid.NewGuid().ToString("N").Substring(0, 16)};Sampled=0;Lineage=1:{Guid.NewGuid().ToString("N").Substring(0, 8)}:0";
+    }
+
+
+    /// <summary>
+    /// Generates a random API Gateway request ID for HTTP API v1 and v2.
+    /// </summary>
+    /// <returns>A string representing a random request ID in the format used by API Gateway for HTTP APIs.</returns>
+    /// <remarks>
```
philasmar, line 152: you haven't closed this tag

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+    /// The generated ID is a 145character string consisting of lowercase letters and numbers, followed by an equals sign.
+    private static string GenerateRequestId()
+    {
+        return Guid.NewGuid().ToString("N").Substring(0, 8) + Guid.NewGuid().ToString("N").Substring(0, 7) + "=";
```
philasmar, line 156: use string interpolation

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    /// <summary>
+    /// Sets the response body on the HttpResponse.
+    /// </summary>
+    /// <param name="response">The HttpResponse to set the body on.</param>
+    /// <param name="body">The body content.</param>
+    /// <param name="isBase64Encoded">Whether the body is Base64 encoded.</param>
```
philasmar, line 164: ApiGatewayEmulatorMode apiGatewayEmulator does not have a doc param

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+    private static void SetResponseBody(HttpResponse response, string? body, bool isBase64Encoded, ApiGatewayEmulatorMode apiGatewayEmulator)
+    {
+        if (!string.IsNullOrEmpty(body))
```
philasmar, line 167: Since we will update the existing HttpResponse, the body and content length might already have values. You should add an else to clear them out if `body` is null or empty.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Sets the content type and status code for API Gateway v1 responses.
+    /// </summary>
+    /// <param name="response">The HttpResponse to set the content type and status code on.</param>
+    /// <param name="headers">The single-value headers.</param>
+    /// <param name="multiValueHeaders">The multi-value headers.</param>
+    /// <param name="statusCode">The status code to set.</param>
```
philasmar, line 190: ApiGatewayEmulatorMode emulatorMode does not have a docs param

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+    private static void SetContentTypeAndStatusCode(HttpResponse response, IDictionary<string, string>? headers, IDictionary<string, IList<string>>? multiValueHeaders, int statusCode, ApiGatewayEmulatorMode emulatorMode)
+    {
+        string? contentType = null;
+
+        if (headers != null && headers.TryGetValue("Content-Type", out var headerContentType))
```
philasmar, line 195: nit: you could just do:
```
if (headers != null && headers.TryGetValue("Content-Type", out contentType))
```

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+        {
+            contentType = headerContentType;
+        }
+        else if (multiValueHeaders != null && multiValueHeaders.TryGetValue("Content-Type", out var multiValueContentType))
+        {
+            contentType = multiValueContentType[0];
```
philasmar, line 201: update to:
```
contentType = multiValueContentType.FirstOrDefault();
```

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+        else
+        {
+            if (emulatorMode == ApiGatewayEmulatorMode.HttpV1)
+            {
+                response.ContentType = "text/plain; charset=utf-8";
+            } 
+            else if (emulatorMode == ApiGatewayEmulatorMode.Rest)
+            {
+                response.ContentType = "application/json";
+            }
```
philasmar, line 217: for the sake of completion, add an else and handle the fact that we could mistakenly make this call for V2. Maybe throw an exception as this is dev error and needs to be updated.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+            {
+                var obj = JsonSerializer.Deserialize<object>(strInput);
+                return true;
+            }
+            catch (JsonException)
+            {
+                return false;
+            }
+        }
+        // a regular string is consisered json in api gateway.
```
philasmar, line 301: nit: typo in "consisered"

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
...
+    private static bool IsValidJson(string? strInput)
+    {
+        if (string.IsNullOrWhiteSpace(strInput)) { return false; }
+        strInput = strInput.Trim();
+        if ((strInput.StartsWith("{") && strInput.EndsWith("}")) ||
+            (strInput.StartsWith("[") && strInput.EndsWith("]")))
+        {
+            try
+            {
+                var obj = JsonSerializer.Deserialize<object>(strInput);
```
philasmar, line 293: obj is never used. You don't need to store the result.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,317 @@
+﻿namespace Amazon.Lambda.TestTool.Extensions;
+
+using Amazon.Lambda.APIGatewayEvents;
+using Amazon.Lambda.TestTool.Models;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Primitives;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection.PortableExecutable;
```
philasmar, line 10: This is not needed

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ServiceCollectionExtensions.cs
philasmar, line 1: Since you didn't make modifications to this file, please reset it so we don't have the EOL changes

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/Amazon.Lambda.TestTool.IntegrationTests.csproj
philasmar, line 1: I am seeing a lot of warnings from this project. Please fix them.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,326 @@
+﻿namespace Amazon.Lambda.TestTool.Extensions;
```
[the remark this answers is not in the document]
gcbeattyAWS: i dont see the amazon copyright header in other files though?

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayTestHelper.cs
gcbeattyAWS: actually, i figured out how to do this. i will push an update using cloudformation

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: i added response.clear up above

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: fixed

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: fixed

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayTestHelper.cs
gcbeattyAWS: ive updated it to use cloudformation to create everything. i reran the integration tests too and they all pass (except for the 1 i have intentionally failing right now related to the v2 json assumptions)

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: fixed

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/Amazon.Lambda.TestTool.IntegrationTests.csproj
gcbeattyAWS: ill go through these now and update. was waiting till the end since they are mostly warnings

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: didnt get what you meant by this

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -1,31 +1,31 @@
-using Amazon.Lambda.TestTool.Services;
```
gcbeattyAWS, line 1: ill remove this file from the pr once i rebase at the end to squash everything

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayResponseExtensionsJsonInference.cs
```diff
@@ -0,0 +1,101 @@
...
+            var content = await actualResponse.Content.ReadAsStringAsync();
+
+            var responsePayload = JsonSerializer.Deserialize<Dictionary<string, object>>(content);
+            Assert.Equal("value", responsePayload["key"].ToString());
+        }
+
+        [Fact]
+        public async Task V2_HandlesNonJsonResponse()
+        {
+            var testResponse = new APIGatewayHttpApiV2ProxyResponse
```
gcbeattyAWS, line 85: this case needs to be reviewed because its currently failing.it seems lambda functions always return valid json

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: i updated this logic to just be straightforward - with the results from my manual testing. that if status code is not set, then api gateway http 2.0 will default to application/json

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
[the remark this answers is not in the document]
gcbeattyAWS: fixed

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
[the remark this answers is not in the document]
gcbeattyAWS: fixed

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/Amazon.Lambda.TestTool.IntegrationTests.csproj
gcbeattyAWS: fixed

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
```diff
@@ -0,0 +1,273 @@
...
+    /// <param name="response">The <see cref="HttpResponse"/> to set the body on.</param>
+    /// <param name="body">The body content.</param>
+    /// <param name="isBase64Encoded">Whether the body is Base64 encoded.</param>
+    /// <param name="apiGatewayEmulator">The <see cref="ApiGatewayEmulatorMode"/> being used.</param>
+    private static void SetResponseBody(HttpResponse response, string? body, bool isBase64Encoded, ApiGatewayEmulatorMode apiGatewayEmulator)
+    {
+        if (!string.IsNullOrEmpty(body))
+        {
+            byte[] bodyBytes;
+            if (isBase64Encoded && ApiGatewayEmulatorMode.Rest != apiGatewayEmulator) // rest api gateway doesnt automatically decode the response
```
normj, line 160: If `isBase64Encoded` is `true` we should always decode the body. I think what you are seeing is that in API Gateway you need to configure your REST API for binary media. And then it will decode it. From our emulator's point of view we should assume that they have correctly configured binary media and so we should decode base 64 encoded bodies.

In CloudFormation you can configure binary media by adding the following resource:
```
    "TheApi": {
      "Type": "AWS::Serverless::Api",
      "Properties": {
        "StageName": "Prod",
        "BinaryMediaTypes": [
          "*~1*"
        ]
      }
    },
```

Then on the event configuration for the serviceless function reference the API:
```
        ],
        "PackageType": "Zip",
        "Events": {
          "RootGet": {
            "Type": "Api",
            "Properties": {
              "Path": "/",
              "Method": "GET",
              "RestApiId" : { "Ref" : "TheApi" }
            }
          }
``

File: Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayResponseExtensionsTestsManual.cs [hunk 34]
```diff
@@ -0,0 +1,66 @@
...
+
+using Amazon.Lambda.APIGatewayEvents;
+using Microsoft.AspNetCore.Http;
+using System.Text.Json;
+using Amazon.Lambda.TestTool.Extensions;
+
+namespace Amazon.Lambda.TestTool.IntegrationTests
+{
+    [Collection("ApiGateway Integration Tests")]
+    public class ApiGatewayResponseExtensionsTestsManual
```
normj, line 12: What does the `Manual` suffix mean here? This test will run automatically as part of integration test runs.

gcbeattyAWS: ill rename this to something better. its not part of the ApiGatewayResponeTestCases list and the test cases are separate. maybe `AdditionalTests`

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: i guess what you meant by this (based on your other comment with the binary type for rest api), is that if the user configured custom response headers for this error code that it might be in the response already? ill have to check that

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: i see. that certainly makes more sense. ill test it out to see what happens

gcbeattyAWS: yeah i just tested this and you are right. i will update this logic  and add a test case

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayResponseExtensionsTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        private readonly ApiGatewayIntegrationTestFixture _fixture;
+
+        public ApiGatewayResponseExtensionsTests(ApiGatewayIntegrationTestFixture fixture)
+        {
+            _fixture = fixture;
+        }
+
+        [Theory]
+        [MemberData(nameof(ApiGatewayResponseTestCases.V1TestCases), MemberType = typeof(ApiGatewayResponseTestCases))]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "xUnit1026:Theory methods should use all of their parameters")]
```
gcbeattyAWS, line 23: i added this suppression so in vscode we can easily see what test is being executed by having test name parameter here

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: ive fixed this now and have also added an integration and unit test

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/ApiGatewayResponseExtensions.cs
gcbeattyAWS: so i was thinking about this more with the way the v2 infers things. i think it actually depends on what language the function is written in. Because for .net statusCode is an int, so default value is 0 and its not possible to be null, which is why when you were testing before you got internal server error. but when i was testing in javascript, i excluded the  statuscode from the response object completely, which api gateway was then able to interpret.

i wonder if we should be updating APIGatewayHttpApiV2ProxyResponse.StatusCode to be nullable/check that logic? or should we just actually remove the logic i have here for when the status code is 0, and i can default to internal server error then.

gcbeattyAWS: i removed the json inference check from here. i think we should add the lambda -> api gateay translation inference in another pr

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayResponseExtensionsAdditionalTests.cs
```diff
@@ -0,0 +1,84 @@
...
+        private readonly HttpClient _httpClient;
+
+        public ApiGatewayResponseExtensionsAdditionalTests(ApiGatewayIntegrationTestFixture fixture)
+        {
+            _fixture = fixture;
+            _httpClient = new HttpClient();
+        }
+
+        //[Fact]
+        //public async Task V2_SetsContentTypeApplicationJsonWhenNoStatusProvided()
```
gcbeattyAWS, line 26: i commented this out for now so it can be referenced for future when we create the lambda -> api gateway response layer

normj (approved):

philasmar (approved):

PR #1911, opened 2024-12-19
fix: greedy variable can be any value

philasmar:
*Description of changes:*
* Previously, I hard coded the greedy variable as {proxy+}. In reality, the greedy variable can have any value and not "proxy" explicitly. This PR updates that behavior.

### DIFF
--- a/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
+++ b/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
@@ -3,6 +3,7 @@
 
 using System.Collections;
 using System.Text.Json;
+using System.Text.RegularExpressions;
 using Amazon.Lambda.TestTool.Models;
 using Amazon.Lambda.TestTool.Services.IO;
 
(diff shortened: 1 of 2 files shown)


Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Services/ApiGatewayRouteConfigServiceTests.cs
```diff
@@ -169,7 +171,7 @@ public void CatchAllRouteConfig()
             {
                 LambdaResourceName = "F1",
                 HttpMethod = "ANY",
-                Path = "/{proxy+}"
```
gcbeattyAWS, line 172: can you add a test case to see what happens with `{proxy+}+}` (is this even valid?)

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
```diff
@@ -142,15 +142,20 @@ private bool IsRouteConfigValid(ApiGatewayRouteConfig routeConfig)
             return false;
         }
 
-        var occurrences = routeConfig.Path.Split("{proxy+}").Length - 1;
```
gcbeattyAWS, line 145: for validating the path i think you may be able to actually just use the regex that aws uses 

```
^(\{[\w.:-]+\+?\}|[a-zA-Z0-9.:_-]+)$
```

image

would that work?

normj (approved):
Approving assuming you address @gcbeattyAWS comments.

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Services/ApiGatewayRouteConfigServiceTests.cs
philasmar: Done

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/ApiGatewayRouteConfigService.cs
philasmar: Done

gcbeattyAWS (approved):

PR #1919, opened 2024-12-26
feat: add multiple function support

philasmar:
*Issue #, if available:*
DOTNET-7864

*Description of changes:*
* Added a drop down that displays the default function as well as any other registered lambda functions.
* Switching the Lambda function also switches the underlying DataStore which updates the entire UI accordingly.
* Updated the Events Modal and leveraged bootstrap instead of a 3rd party library (Blazored.Modal)
* Added other UI updates to enhance the user experience

No function registered or default function registered:
image

Multiple functions registered:
image

Active event UI update:
image

History/Queued events list UI update:
image

Event modal UI update:
image


gcbeattyAWS:
`No function registered or default function registered` - in this case for default function registered - does it still have a name or anything or no?  i was thinking it still may make sense to have that lambda functions dropdown displayed with the default lambda selected, even if there is only one there. same thing with no function registered - wondering if we should tell the user "No functions registered and you need to register one.. " or something like that.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Components/Pages/Home.razor
```diff
@@ -1,104 +1,133 @@
...
+@using Amazon.Lambda.TestTool.SampleRequests
+@using Amazon.Lambda.TestTool.Services
 
 <PageTitle>Lambda Function Tester</PageTitle>
 
-<h3>Lambda Function Tester</h3>
+<div class="d-flex align-items-center mb-2">
+    <h3 class="me-auto">Lambda Function Tester</h3>
+    <div class="form-floating @(_availableLambdaFunctions.Count <= 1 ? "d-none" : "")" style="min-width: 300px;">
+        <select class="form-select" id="selectedFunctionName" @bind="SelectedFunctionName">
```
gcbeattyAWS, line 12: `No function registered or default function registered` - in this case for default function registered - Is the picture in the PR for this correct? Based on the code it seems like it will still show the drop down with "default lambda" as the value, which i think should be the correct behavior. But i just wanted to double check.

For no function registered - wondering if we should still show the dropdown menu, but then have an error/message saying a function needs to be registered or something to let the user know?


same thing with no function registered - wondering if we should tell the user "No functions registered and you need to register one.. " (or maybe have the drop down there but empty to show that it wont work?) or something like that.

normj (approved):
Tool is looking really good. 

One thing I am noticing but not necessary caused by this change but if you run the tool outside of an environment like an IDE that has the environment variable `ASPNETCORE_ENVIRONMENT` set to `Development` I get errors running the tool. Basically if I do `dotnet run` in the project directory I get errors.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Components/Pages/Home.razor
philasmar: Functionally, having the default function registered or no function registered, they are the same from a UI perspective. I don't want to show the dropdown because it serves no purpose. I want it to behave similar to how the old test tool behaves. As soon as another function is registered, the drop down will pop-up and allow the users to switch. In the code, `@(_availableLambdaFunctions.Count <= 1 ? "d-none" : "")` the drop down will disappear if _availableLambdaFunctions has 1 function or less. This is the default case, since at least the default function will always get registered.

philasmar:
> Tool is looking really good. My approval is from a basic scan and running the tool. I can do a deeper review of the tool's current state when I get back but don't want you to be held up while people are out of the office.
> 
> One thing I am noticing but not necessary caused by this change but if you run the tool outside of an environment like an IDE that has the environment variable `ASPNETCORE_ENVIRONMENT` set to `Development` I get errors running the tool. Basically if I do `dotnet run` in the project directory I get errors.

Yes I have noticed this behavior when `ASPNETCORE_ENVIRONMENT` is not set. This is unrelated to my change and has been there since we started working on the tool, but I was planning on looking into it separately.

gcbeattyAWS (approved):

PR #1921, opened 2024-12-30
feat: use offcanvas for function switching

philasmar:
*Issue #, if available:*
DOTNET-7864

*Description of changes:*
* Added offcanvas control to display available functions
* Removed the lambda function drop down in favor of offcanvas for a more modern UI
* The home page now shows the selected function name as the title
* Clicking on the function name or the switch lambda button will open the offcanvas
* The button to switch functions now starts glowing when a new function is added. When you click on it, the glow goes away.

No or default function registered:
image

Registering more functions shows a new button to switch the selected Lambda
image

Clicking on Lambda name or switch lambda button:
image

### DIFF
--- a/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Components/Layout/MainLayout.razor
+++ b/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Components/Layout/MainLayout.razor
@@ -1,5 +1,4 @@
 ﻿@using Amazon.Lambda.TestTool.Services
-@using Amazon.Lambda.TestTool.Utilities
 @inherits LayoutComponentBase
 
 @inject IJSRuntime Js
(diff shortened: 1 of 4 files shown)


normj:
I'm concerned about discoverability for switching functions. When running both a the test tool and a function hitting the test tool it took me a while to notice where the switch functionality was at. And I knew to look for it. A user with little context might have a harder time.

If changes the user experience from a one gesture of changing a drop down box to clicking about on the far right of the screen and then go all the way over to the left side of the screen.

My comments are subjective and I agree it looks nice but I not sure it makes the user experience better.

gcbeattyAWS (approved):
dropdown vs offcanvas doesnt matter too much to me tbh. Im fine if we want to merge this change since we can always get feedback and change it later if needed

normj (approved):

PR #1925, opened 2025-01-08
Fix test tool hanging when restarting debugging session.

normj:
*Description of changes:*
In the test tool if a user starts debugging their Lambda function, then stops debugging without letting the function finish it puts the test tool in hanging situation when the debug session is restarted. This happens because the first debug session never told the test tool that it was done processing the event so the test tool thinks the event is still active and executing. When the debug session is restarted and the `GetNextInvocation` is called it blocks waiting for the event to finish.

In the real Lambda environment the `GetNextInvocation` would never be called till the event is done processing. So when in the `GetNextInvocation` there should never be an active event still executing. For the test tool we will assume that if there is an active event with its status as executing that this is a debug restart scenario and resend the event to be debugged again.

### DIFF
--- a/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/LambdaRuntimeAPI.cs
+++ b/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/LambdaRuntimeAPI.cs
@@ -92,10 +92,24 @@ public class LambdaRuntimeApi
         var runtimeDataStore = _runtimeApiDataStoreManager.GetLambdaRuntimeDataStore(functionName);
 
         EventContainer? activeEvent;
-        while (!runtimeDataStore.TryActivateEvent(out activeEvent))
+
+        // A Lambda function should never call to get the next event till it was done
+        // processing the active event and there is no more active event. If there
+        // is an active event still executing that most likely means the previous debug session was
+        // killed leaving the event active. In that case resend the active event
+        // to restart debugging the event.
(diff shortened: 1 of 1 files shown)


gcbeattyAWS (approved):
just a general question. how does the lambda function/runtime know whether it should be calling either the GetNextInvocationDefaultFunction url vs calling the GetNextInvocation url? i assume if multiple lambdas are running at the same time they will use the url with their function name. in what case would the other url be called?

```
        app.MapGet("/2018-06-01/runtime/invocation/next", GetNextInvocationDefaultFunction);
        app.MapGet("/{functionName}/2018-06-01/runtime/invocation/next", GetNextInvocation);
 ```

normj:
> just a general question. how does the lambda function/runtime know whether it should be calling either the GetNextInvocationDefaultFunction url vs calling the GetNextInvocation url? i assume if multiple lambdas are running at the same time they will use the url with their function name. in what case would the other url be called?
> 
> ```
>         app.MapGet("/2018-06-01/runtime/invocation/next", GetNextInvocationDefaultFunction);
>         app.MapGet("/{functionName}/2018-06-01/runtime/invocation/next", GetNextInvocation);
> ```

In the Lambda function if the `AWS_LAMBDA_RUNTIME_API` environment variable is set to `localhost:<port>` it will go to `GetNextInvocationDefaultFunction`. That is how we direct users in the current test tool which isn't designed for multiple functions. If you set `AWS_LAMBDA_RUNTIME_API` to `localhost:<port>/<function-name>` then it will go to `GetNextInvocation`. When we get to the Aspire integration we will set the environment this way.

PR #1942, opened 2025-01-25
Add port environment variables for Aspire integration

normj:
*Description of changes:*
In Aspire we don't know the ports that it will allocate for the test tool till after the command line parameters have been set for the executable. To work around this I added environment variables that can be set by Aspire for the executable to pick up and use for ports.

I had trouble loading the test tool project for debugging due to the recent commit adding the wwwroot folder for packaging. To get around that I have the property only be set in Release mode allowing the project to be loaded in visual studio and `dotnet pack` which is run in release mode to find the static content. Not a perfect solution because it means VS will have issues in release mode. Unblocks for now but open to better solution.

I also addressed the remaining warnings in the project.

### DIFF
--- a/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
+++ b/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
@@ -168,9 +168,11 @@ public static class HttpContextExtensions
 
         if (emulatorMode == ApiGatewayEmulatorMode.Rest) // rest uses encoded value for the path params
         {
+#pragma warning disable SYSLIB0013 // Type or member is obsolete
             var encodedPathParameters = pathParameters.ToDictionary(
                     kvp => kvp.Key,
-                    kvp => Uri.EscapeUriString(kvp.Value)); // intentionally using EscapeURiString over EscapeDataString since EscapeURiString correctly handles reserved characters :/?#[]@!$&'()*+,;= in this case
+                    kvp => Uri.EscapeUriString(kvp.Value)); // intentionally using EscapeUriString over EscapeDataString since EscapeUriString correctly handles reserved characters :/?#[]@!$&'()*+,;= in this case
+#pragma warning restore SYSLIB0013 // Type or member is obsolete
             pathParameters = encodedPathParameters;
         }
 
--- a/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Program.cs
+++ b/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Program.cs
@@ -19,17 +19,4 @@ app.Configure(config =>
     config.SetApplicationName(Constants.ToolName);
 });
 
-var arguments = new List<string>(args);
-
-if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable("LAMBDA_RUNTIME_API_PORT")))
-{
-    arguments.Add("--port");
-    arguments.Add(Environment.GetEnvironmentVariable("LAMBDA_RUNTIME_API_PORT")!);
-}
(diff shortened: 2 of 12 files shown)


Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Utilities/HttpRequestUtility.cs
```diff
@@ -107,7 +107,7 @@ public static (IDictionary<string, string>, IDictionary<string, IList<string>>)
         {
             var key = lowerCaseKeyName ? header.Key.ToLower() : header.Key;
             singleValueHeaders[key] = header.Value.Last() ?? "";
-            multiValueHeaders[key] = [.. header.Value];
+            multiValueHeaders[key] = [.. header.Value!];
```
normj, line 110: I think the null warnings happening in this file were a false positive. They warnings only show up when doing `dotnet pack` not in VS. And the `Value` is a struct so it can't be null.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Program.cs
```diff
@@ -19,4 +19,17 @@
     config.SetApplicationName(Constants.ToolName);
 });
 
-return await app.RunAsync(args);
+var arguments = new List<string>(args);
```
philasmar, line 22: I prefer that you not modify the arguments directly and instead add your logic to https://github.com/aws/aws-lambda-dotnet/blob/feature/lambdatesttool-v2/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/RunCommand.cs or https://github.com/aws/aws-lambda-dotnet/blob/feature/lambdatesttool-v2/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/Settings/RunCommandSettings.cs

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Extensions/HttpContextExtensions.cs
```diff
@@ -168,9 +168,11 @@ public static async Task<APIGatewayProxyRequest> ToApiGatewayRequest(
 
         if (emulatorMode == ApiGatewayEmulatorMode.Rest) // rest uses encoded value for the path params
         {
+#pragma warning disable SYSLIB0013 // Type or member is obsolete
             var encodedPathParameters = pathParameters.ToDictionary(
                     kvp => kvp.Key,
-                    kvp => Uri.EscapeUriString(kvp.Value)); // intentionally using EscapeURiString over EscapeDataString since EscapeURiString correctly handles reserved characters :/?#[]@!$&'()*+,;= in this case
+                    kvp => Uri.EscapeUriString(kvp.Value)); // intentionally using EscapeUriString over EscapeDataString since EscapeUriString correctly handles reserved characters :/?#[]@!$&'()*+,;= in this case
```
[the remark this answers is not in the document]
normj: @gcbeattyAWS Can you answer this question since this was your change?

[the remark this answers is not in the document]
gcbeattyAWS: so i tried looking for an alternative function that was not deprecated and produced the same output that we needed - but i could not find any. i think i tried 5 or 6 different ways and spent a couple hours on it. If we want to ensure the function never gets dropped I think we could just re-implement/copy their source code for it no?

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Program.cs
normj: Done

gcbeattyAWS (approved):

philasmar (approved):

PR #1954, opened 2025-01-29
fix: fix unit and integ tests hanging

philasmar:
*Issue #, if available:*
DOTNET-7945

*Description of changes:*
* I have updated the tests to atomically get the next available port for the processes we are spinning up.
* Added DirectoryHelpers to create temp working directories for the tests that modify the project in place.
* Skipped the tests that are hanging to unblock the feature branch. Will work on the hanging test as a follow up.

Note: The unit tests are now passing in the GitHub CI and Release Pipeline. However, the integration tests are still failing. Will work on that as a follow up.

### DIFF
--- a/Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Helpers/TestHelpers.cs
+++ b/Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Helpers/TestHelpers.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+using System.Collections.Concurrent;
+
 namespace Amazon.Lambda.TestTool.UnitTests.Helpers;
 
 internal static class TestHelpers
@@ -37,5 +39,18 @@ internal static class TestHelpers
         {
             return await client.GetAsync(url);
         }
+    }
+
+    private static int _maxLambdaRuntimePort = 6000;
+    private static int _maxApiGatewayPort = 9000;
+
+    public static int GetNextLambdaRuntimePort()
+    {
+        return Interlocked.Increment(ref _maxLambdaRuntimePort);
+    }
+
+    public static int GetNextApiGatewayPort()
+    {
+        return Interlocked.Increment(ref _maxApiGatewayPort);
     }
 }
--- a/Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
+++ b/Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
@@ -56,7 +56,11 @@ public class PackagingTests : IDisposable
         return output.Trim();
     }
 
+#if DEBUG
+    [Fact]
+#else
     [Fact(Skip = "Skipping this test as it is not working properly.")]
+#endif
     public void VerifyPackageContentsHasRuntimeSupport()
     {
         var projectPath = Path.Combine(_workingDirectory, "Tools", "LambdaTestTool-v2", "src", "Amazon.Lambda.TestTool", "Amazon.Lambda.TestTool.csproj");
--- a/Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
+++ b/Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
@@ -1,22 +1,22 @@
-using System;
 using System.Diagnostics;
-using System.IO;
 using System.IO.Compression;
-using System.Collections.Generic;
-using System.Linq;
+using Amazon.Lambda.TestTool.UnitTests.Utilities;
 using Xunit;
 using Xunit.Abstractions;
 
 namespace Amazon.Lambda.TestTool.UnitTests;
 
-public class PackagingTests
+public class PackagingTests : IDisposable
 {
     private readonly ITestOutputHelper _output;
     private readonly string[] _expectedFrameworks;
+    private readonly string _workingDirectory;
 
     public PackagingTests(ITestOutputHelper output)
     {
         _output = output;
+        var solutionRoot = FindSolutionRoot();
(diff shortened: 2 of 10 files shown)


Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Processes/ApiGatewayEmulatorProcessTests.cs
```diff
@@ -21,7 +21,7 @@ public async Task RouteNotFound(ApiGatewayEmulatorMode mode, HttpStatusCode stat
         // Arrange
         var cancellationSource = new CancellationTokenSource();
         cancellationSource.CancelAfter(5000);
-        var settings = new RunCommandSettings { ApiGatewayEmulatorPort = 9003,  ApiGatewayEmulatorMode = mode, NoLaunchWindow = true};
+        var settings = new RunCommandSettings { ApiGatewayEmulatorPort = 9004,  ApiGatewayEmulatorMode = mode, NoLaunchWindow = true};
```
gcbeattyAWS, line 24: one other idea for the ports is that we can automatically find any open ports and use those. not sure if thats related to your issue or not though

buildtools/ci.buildspec.yml
```diff
@@ -12,5 +12,6 @@ phases:
       - curl -sSL https://dot.net/v1/dotnet-install.sh | bash /dev/stdin --channel 9.0
   build:
     commands:
+      - dotnet test -c Release --logger "console;verbosity=detailed" ./Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Amazon.Lambda.TestTool.UnitTests.csproj
```
gcbeattyAWS, line 15: this is a temporary change right?

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
```diff
@@ -77,7 +78,7 @@ public void VerifyPackageContentsHasRuntimeSupport()
         packProcess.Start();
         string packOutput = packProcess.StandardOutput.ReadToEnd();
         string packError = packProcess.StandardError.ReadToEnd();
-        packProcess.WaitForExit();
+        packProcess.WaitForExit(int.MaxValue);
```
gcbeattyAWS, line 81: do we want to make this something lower? like 5 minutes?

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
```diff
@@ -44,8 +44,10 @@ private string GetRuntimeSupportTargetFrameworks()
         process.Start();
         var output = process.StandardOutput.ReadToEnd();
         var error = process.StandardError.ReadToEnd();
-        process.WaitForExit();
+        process.WaitForExit(int.MaxValue);
```
gcbeattyAWS, line 47: same here

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Helpers/TestHelpers.cs
```diff
@@ -38,4 +40,17 @@ internal static async Task<HttpResponseMessage> SendRequest(string url)
             return await client.GetAsync(url);
         }
     }
+
+    private static int _maxLambdaRuntimePort = 6000;
+    private static int _maxApiGatewayPort = 9000;
+
+    public static int GetNextLambdaRuntimePort()
```
gcbeattyAWS, line 47: can we update this to something like https://github.com/aws/aws-lambda-dotnet/pull/1955/files#diff-16dac3e7db3ba4a144e7ac3063ba31b74a53037cb85cc1a5e35410efb0d6393dR311? the reason being is depending on the developers machine/codebuild machine, the above ports may not actually be free. I think we should use `TcpListener` for the OS to give us free ports

gcbeattyAWS (changes requested):

buildtools/ci.buildspec.yml
philasmar: No it's not, I'm separating the way we run the test tool tests from the other lambda ones. This is how the DevEx Pipeline runs them. I will do the same for Integ tests in followup PRs.

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
philasmar: The int.MaxValue is used to workaround a microsoft bug. We don't actually intend to wait that long. We use it in other repos. Sometimes the process hangs, and when we pass `int.MaxValue` the process stops hanging.

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/PackagingTests.cs
philasmar: The int.MaxValue is used to workaround a microsoft bug. We don't actually intend to wait that long. We use it in other repos. Sometimes the process hangs, and when we pass `int.MaxValue` the process stops hanging.

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Helpers/TestHelpers.cs
philasmar: The point of the change I'm making is so we can atomically check what the next port is that other tests are not using. I believe the TcpListener is a bit overkill as these port ranges are always free in the CI. Even on our machines, these ports are usually not used.

gcbeattyAWS (approved):

normj (approved):
Approve but I recommend when add the skips for tests for now still allow the test to run in debug mode.

```
#if !DEBUG
[Fact(Skip = "Skipping this test as it is not working properly.")]
#endif
```

philasmar:
> Approve but I recommend when add the skips for tests for now still allow the test to run in debug mode.
> 
> ```
> #if !DEBUG
> [Fact(Skip = "Skipping this test as it is not working properly.")]
> #endif
> ```

Fixed

PR #1964, opened 2025-01-31
fix: fix skipped tests

philasmar:
*Issue #, if available:*
DOTNET-7945

*Description of changes:*
* Moved common utilities to the Common project
* Fixed previously skipped tests
* PackagingTests unit tests were previously hanging due to the MSBUILD Target in the test tool project which was building and publishing RuntimeSupport during the test run. I changed this behavior to only build and publish when the TestTool project is being built. When the test is ran, the needed binaries are copied over to the NuGet package. This resolves the hanging.
* To support the previous behavior, I created Targets for running all the tests in the CI as well as the unit and integ tests separately. GitHub CI and Release Pipeline will now trigger the build.proj instead of calling dotnet test directly.
* In some tests, the executed Lambda function was writing to Standard Error which was causing `dotnet msbuild` to fail with a -1 exit code. I had to ignore Console.Error in those tests.
* For integration tests, with the help of @GarrettBeatty, I moved the tests from testing an external project in the repo to only writing the function handler method and passing it to the Bootstrap similar to other tests.

Note: The tests are still flaky as there are other underlying issues such as us setting Environment Variables for tests which is global for all tests. However, this PR fixes the definitive failure of the tests that were skipped.


normj (approved):

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayEmulatorProcessTests.cs
```diff
@@ -15,171 +18,180 @@
...
+        _cancellationTokenSource.CancelAfter(TimeSpan.FromSeconds(30));
+        var consoleError = Console.Error;
         try
         {
-            StartTestToolProcess(ApiGatewayEmulatorMode.HttpV2, config, lambdaPort, apiGatewayPort, cancellationTokenSource);
+            Console.SetError(TextWriter.Null);
+            await StartTestToolProcessAsync(ApiGatewayEmulatorMode.HttpV2, "testfunction", lambdaPort, apiGatewayPort, _cancellationTokenSource);
             await WaitForGatewayHealthCheck(apiGatewayPort);
-            await StartLambdaProcess(config, lambdaPort);
+            var handler = (APIGatewayHttpApiV2ProxyRequest request, ILambdaContext context) =>
```
GarrettBeatty, line 44: shouldnt we delete the whole testapps folder then if we are adding all of the handlers in here?

GarrettBeatty (changes requested):

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.IntegrationTests/ApiGatewayEmulatorProcessTests.cs
philasmar: I deleted them

GarrettBeatty (approved):

PR #1969, opened 2025-02-05
Add version to test tool

normj:
*Description of changes:*
Add a `--version` switch that if set prints the tool's version as a JSON document. This allows other tooling like the Aspire integration to shell out to the command and programmatically check the installed version. JSON document was chosen to allow the possibility of expanding on the info returned.

### DIFF
--- a/.autover/changes/efa829ed-59af-46eb-a5e5-ef471698c82d.json
+++ b/.autover/changes/efa829ed-59af-46eb-a5e5-ef471698c82d.json
@@ -4,7 +4,8 @@
       "Name": "Amazon.Lambda.TestTool",
       "Type": "Patch",
       "ChangelogMessages": [
-        "Add --version switch to allow integrators to check what version is installed"
+        "Breaking change: Switch to use commands to invoke the tool. For example to run the Lambda emulator use the command 'dotnet lambda-test-tool start --lambda-emulator-port 5050'",
+		"Add new info command to get metadata about the tool. For example getting the version number of the tool."
       ]
     }
   ]
--- a/.autover/changes/efa829ed-59af-46eb-a5e5-ef471698c82d.json
+++ b/.autover/changes/efa829ed-59af-46eb-a5e5-ef471698c82d.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.TestTool",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Add --version switch to allow integrators to check what version is installed"
+      ]
+    }
+  ]
+}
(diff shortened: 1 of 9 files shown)


GarrettBeatty (approved):

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/RunCommand.cs
```diff
@@ -28,6 +29,12 @@ public override async Task<int> ExecuteAsync(CommandContext context, RunCommandS
     {
         try
         {
+            if (settings.PrintVersionInfo)
```
GarrettBeatty, line 32: can we add a unit test to https://github.com/aws/aws-lambda-dotnet/blob/8c1705f523c9d75db87bc326f5ee3d5ec6de39be/Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Commands/RunCommandTests.cs#L17? 

Also as a side note, do we want to enforce unit test coverage for future PRs as a check? What do we do for other projects that we have?

normj: I was being reluctant adding a mocking layer around stdout so I could reliable unit test this. Turns out we are already mocking stdout with `IToolInteractiveService` so I added the unit test.

normj:
I renamed the `--version` switch to `--tool-info` because I though the name seemed more broad then return just the version number. I also added the `install-path` of the tooling. I though that might come in handy when we work on setting up launchsettings.json file.

.autover/changes/efa829ed-59af-46eb-a5e5-ef471698c82d.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.TestTool",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Add --version switch to allow integrators to check what version is installed"
```
philasmar, line 7: you missed updating the changelog

philasmar (commented):
I'm ok with the change, but instead of us putting everything new thing in RunCommand. We should create another command for info. So the command would be `dotnet-lambda-test-tool info`. That way, we can potentially add further command line arguments to format the info we are looking for.

GarrettBeatty (approved):

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/Settings/RunCommandSettings.cs
```diff
@@ -36,22 +36,6 @@ public sealed class RunCommandSettings : CommandSettings
     [Description("Disable auto launching the test tool's web interface in a browser.")]
     public bool NoLaunchWindow { get; set; }
 
-    /// <summary>
```
normj, line 39: These were leftover from the old version of the tooling and are not used in the new version.

philasmar (approved):

Issue #1977, opened 2025-02-12
Error with code generated with Amazon.Lambda.Annotations (event property must have both add and remove accessors)


romulomrossi-caylent:
### Describe the bug

I'm trying to implement a Lambda function that a Step Function will trigger. I'm using the LambdaFunction decorator, as well as the LambdaStartup.

<img width="2273" alt="Image" src="https://github.com/user-attachments/assets/f8bac3e3-b007-4bf6-bc39-64fc4e9f5491" />

But I'm getting the following errors from the dotnet build:

`/FacStats/Amazon.Lambda.Annotations.SourceGenerator/Amazon.Lambda.Annotations.SourceGenerator.Generator/FacStatsProcessor_ProcessFacilityStats_Generated.g.cs(48,94): error CS0065: 'FacStatsProcessor_ProcessFacilityStats_Generated.': event property must have both add and remove accessors
`

`/FacStats/Amazon.Lambda.Annotations.SourceGenerator/Amazon.Lambda.Annotations.SourceGenerator.Generator/FacStatsProcessor_ProcessFacilityStats_Generated.g.cs(48,50): error CS1994: The 'async' modifier can only be used in methods that have a body. 
`

`/FacStats/Amazon.Lambda.Annotations.SourceGenerator/Amazon.Lambda.Annotations.SourceGenerator.Generator/FacStatsProcessor_ProcessFacilityStats_Generated.g.cs(56,63): error CS0102: The type 'FacStatsProcessor_ProcessFacilityStats_Generated' already contains a definition for '' `

Thank you in advance for any help.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Generated code to be successfully built.

### Current Behavior

Build fails with the errors in the description.

### Reproduction Steps

Just try to implement a handler with only the LambdaHandler decorator. I tested it, and it works if I add the [HttpApi()] decorator, but I don't want it to receive any API requests.

### AWS .NET SDK and/or Package version used

    <PackageReference Include="Amazon.Lambda.RuntimeSupport" Version="1.12.0" />
    <PackageReference Include="Amazon.Lambda.Core" Version="2.5.0" />
    <PackageReference Include="Amazon.Lambda.Serialization.SystemTextJson" Version="2.4.4" />
    <PackageReference Include="Amazon.Lambda.Annotations" Version="1.6.1" />

### Targeted .NET Platform

.NET 8

### Operating System and version

[...]

PR #1984, opened 2025-02-17
fix: implemented change to always use `__` as prefix and suffix for parameter names generated for a LambdaFunction.

ashishdhingra:
*Issue #, if available:* #1977

*Description of changes:*
Implemented change to always use `__` as prefix and suffix for parameter names generated for a LambdaFunction.

**Debugging Steps:**
Referred article https://nicksnettravels.builttoroam.com/debug-code-gen/ to debug the project (needed to understand the code flow for source generator):
  - Added below reference of `Amazon.Lambda.Annotations.SourceGenerator` project in the target application project:
    ```XML
    <ItemGroup>
      <ProjectReference 
			Include="..\aws-lambda-dotnet\Libraries\src\Amazon.Lambda.Annotations.SourceGenerator.csproj"
			OutputItemType="Analyzer"
			ReferenceOutputAssembly="false" />
    </ItemGroup>
    ```
  - Uncommented the below code in [Generator](https://github.com/aws/aws-lambda-dotnet/blob/d10b8d73b41c01ccb60715ac61bb5488f0fa547a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs#L42) constructor:
    ```C#
    #if DEBUG
      if (!Debugger.IsAttached)
      {
          Debugger.Launch();
      }
    #endif 
    ```
  We would need to clean solution, close Visual Studio, reload project solution. It would prompt to select Visual Studio instance for debugging. If selecting current instance, it would have dialog looking for change in source files as `Generator()` constructor is debugged.
  Thereafter, set breakpoint in [Execute()](https://github.com/aws/aws-lambda-dotnet/blob/d10b8d73b41c01ccb60715ac61bb5488f0fa547a/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs#L52) method and expand `Dependencies` > `Analyzers` > `Amazon.Lambda.Annotations.SourceGenerator`. It will try to inspect target application project and execute source generator logic.

This PR also includes the following changes:
- Executed command `autover change --project-name "Amazon.Lambda.Annotations" -m "fix: implemented change to always use __ as prefix and suffix for parameter names generated for a LambdaFunction."` to create the change file using **ONLY** `Amazon.Lambda.Annotations` project. Per @normj, there is a single NuGet package with the both Amazon.Lambda.Annotations.dll for runtime and Amazon.Lambda.Annotaitons.SourceGenerator.dll run by the Roslyn compiler. The csproj file for [Amazon.Lambda.Annotations.SourceGenerator](https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Amazon.Lambda.Annotations.SourceGenerator.csproj#L33). Refer existing [commit](https://github.com/aws/aws-lambda-dotnet/commit/be26107eb2c2e9b39467212a2f2b9dcc6323fc85#diff-0b3a12c7e2cb3ce5d863c3531b486d9f6808a7a5f506219b0dd68748fff7e382).

- Updated integration test snapshots which would be changed as a result of this change.
- Updated existing test to include the above scenario.
___


Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/GeneratedMethodModelBuilder.cs
```diff
@@ -215,6 +215,11 @@ private static IList<ParameterModel> BuildParameters(IMethodSymbol lambdaMethodS
                     }
                     else
                     {
+                        // Somehow, Roslyn doesn't include @ character when using with reserved keyword. Add a check just in case Roslyn fix is implemented in later versions.
```
GarrettBeatty, line 218: i didnt get what this comment means. why do we have to have this if statement?

ashishdhingra: For this customer's scenario, they use `@event` as parameter name. When code reaches here, Roslyn doesn't return its name as `@event`, it returns `event`. This causes issue as mentioned by customer. This condition will handle scenario if Microsoft decides to fix it and return parameter name as set by user. `@` character can only be present at the beginning of the identifier.

GarrettBeatty: `@ character can only be present at the beginning of the identifier.` - ah okay. this is what i didnt understand at first. Makes sense then

GarrettBeatty (approved):

PR #1989, opened 2025-02-21
Add Runtime API Endpoint Command Line Arg to Runtime Support

philasmar:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/LambdaBootstrapBuilder.cs
+++ b/Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/LambdaBootstrapBuilder.cs
@@ -19,6 +19,7 @@ using System.Net.Http;
 using System.Threading.Tasks;
 
 using Amazon.Lambda.Core;
+using Amazon.Lambda.RuntimeSupport.Bootstrap;
 
 namespace Amazon.Lambda.RuntimeSupport
 {
@@ -31,6 +32,7 @@ namespace Amazon.Lambda.RuntimeSupport
         private HandlerWrapper _handlerWrapper;
         private HttpClient _httpClient;
         private LambdaBootstrapInitializer _lambdaBootstrapInitializer;
+        private LambdaBootstrapOptions _lambdaBootstrapOptions = new LambdaBootstrapOptions();
 
         private LambdaBootstrapBuilder(HandlerWrapper handlerWrapper)
         {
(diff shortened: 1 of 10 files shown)


Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/LambdaBootstrapOptions.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace Amazon.Lambda.RuntimeSupport.Bootstrap
+{
+    public class LambdaBootstrapOptions
+    {
+        public string RuntimeApiEndpoint { get; set; }
```
normj, line 8: Be sure to add documentation that this should only be set for testing environments. Not when deployed to Lambda.

Libraries/src/Amazon.Lambda.RuntimeSupport/Program.cs
```diff
@@ -60,5 +63,46 @@ private static System.Reflection.Assembly ResolveSnapshotRestoreAssembly(Assembl
             return null;
         }
 #endif
+
+        private static LambdaBootstrapOptions ParseCommandLineArguments(string[] args)
+        {
+            var arguments = new Dictionary<string, string>();
+
+            string currentKey = null;
+            for (int i = 1; i < args.Length; i++)
```
normj, line 72: Probably personal preference but I find this logic a bit confusing especially keeping state outside of the loop and waiting for value to get resolved next iteration of the look. What about this something like this which is untested.

```
private static LambdaBootstrapOptions ParseCommandLineArguments(string[] args)
{
    var option = new LambdaBootstrapOptions();

    if (args.Length > 1)
    {
        var arguments = new Dictionary<string, string>();
        for (int i = 1; i < args.Length; i++)
        {
            if (!string.IsNullOrEmpty(args[i]))
                throw new Exception();

            string key = args[i];
            int valuePos = i + 1;
            string value = null;
            if (args.Length < valuePos && !args[valuePos].StartsWith("-"))                
            {
                value = args[valuePos];
                i++;
            }

            arguments[key] = value;
        }

        if (arguments.TryGetValue("--runtime-api-endpoint", out var argument))
            option.RuntimeApiEndpoint = argument;

    }

    return option;
}
```

normj (commented):
I assume the Lambd1 project is just there for demonstration purposes as a draft PR.

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/LambdaBootstrapBuilder.cs
```diff
@@ -449,14 +451,28 @@ public LambdaBootstrapBuilder UseBootstrapHandler(LambdaBootstrapInitializer lam
             return this;
         }
 
+        /// <summary>
+        /// Configure Lambda Bootstrap options
+        /// This is for testing purposes and should not be used in production.
```
GarrettBeatty, line 456: is there any way to prevent this from being used in production besides just having a comment?

philasmar: I don't think we want to prevent people from using it. If someone has a use case and this fits it, then why not. However, for Lambda's production environment, the value is passed using env vars. So we're adding a note for people to see that this is meant for testing purposes. I don't know if Norm has different thoughts on it.

PR #2041, opened 2025-04-10
Add support for optional SQS client region in event source test tool

fennekin23:
*Description of changes:*
There are some cases when service URL needs to be overridden for SDK client. This will result in setting `RegionEndpoint` property to `null` in SDK client config and will cause `NullReferenceException` in SQS event source background service.

With this change the region is read from SQS queue ARN rather than from SQS client config, which allows to have custom service URL configured for SQS client.


normj (commented):
The change looks good but can you fix the whitespace change issue so it doesn't look like you changed the entire file.

fennekin23:
> The change looks good but can you fix the whitespace change issue so it doesn't look like you changed the entire file.

Sorry, missed this.
Should be fixed now.

.autover/changes/d0cfdbc8-e15b-49ae-b0b4-ec882dec15f2.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.TestTool",
+      "Type": "Minor",
```
normj, line 5: Can you make this a `Patch` instead of `Minor`?

normj (changes requested):

philasmar (approved):

normj:
PR was released in version `0.10.1` of Amazon.Lambda.TestTool. Currently the `Aspire.Hosting.AWS` package is going to default to installing `0.10.0` but if you manually update the installed version of `Amazon.Lambda.TestTool` then `Aspire.Hosting.AWS` will recognize a newer version is installed and use that. To manually install run the following command on the command line.
```
dotnet tool update -g Amazon.Lambda.TestTool
```

Most likely in the next release of `Aspire.Hosting.AWS` we will bump up its default version to the latest version.

PR #2045, opened 2025-04-15
Update lambda test tool tests to use snapshots

GarrettBeatty:
*Issue #, if available:*

*Description of changes:*
We previously had integration tests for lambda test tool which did the following:
1. Call an extension method (toApiGatewayRequest/toHttpResponse,etc)
2. Deploy and Invoke a real api gateway and compare the actual api gateway request/response with the expected request/response

This process was overkill because the real api gateway request and response will not change. We can instead use "snapshots" (static files) to validate that our extension methods produce the correct request and response. This will reduce the amount of time it takes for our tests to run too.

**The tests now take 5 seconds to run instead of 15 minutes** after this change.

*Code Changes*
The actual changes/steps that I have done in this PR are:
1. I created snapshots automatically by running the existing tests, but then additionally calling a `saveSnapshot` method that i have created. This produced the snapshot json files you see.
2. I then modified the tests to read these snapshot files instead of invoking the real api gateway
3. I then deleted all of the code responsible for creating api gateways, cloudformation stacks, etc.
4. After I did this, I realized that our integration tests and unit tests for extension methods are essentially identical because they are testing the same thing (validating the extension method works by comparing it to some static value). Because of this I moved the "integration" tests for these functions into the unit test project, and replaced the existing unit tests with them (because they are now duplicative).
5. I also removed the additional `Assertions` in the `ApiGatewayResponseTestCases` class because these assertions are duplicate/not needed. The reason that they are not needed is because they are checking things that are already checked via reading the snapshot file.


Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/SnapshotHelper/ApiGatewayTestSnapshot.cs
```diff
@@ -0,0 +1,70 @@
+using System.Net;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Amazon.Lambda.TestTool.UnitTests.SnapshotHelper;
+
+public class HttpResponseMessageConverter : JsonConverter<HttpResponseMessage>
```
GarrettBeatty, line 7: have to use a cusotm json serializer here instead of default jsonconverter because HttpResponseMessage is cyclic in nature with some of its attributes so json cannot be easily created

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/SnapshotHelper/ApiGatewayTestSnapshot.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    public override HttpResponseMessage Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        using var document = JsonDocument.ParseValue(ref reader);
+        var element = document.RootElement;
+
+        var response = new HttpResponseMessage((HttpStatusCode)element.GetProperty("StatusCode").GetInt32());
```
GarrettBeatty, line 14: we only care about status code, content, and headers in our test

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Extensions/ApiGatewayResponseTestCases.cs
```diff
@@ -0,0 +1,367 @@
...
+
+using System.Text;
+using System.Text.Json;
+using Amazon.Lambda.APIGatewayEvents;
+
+namespace Amazon.Lambda.TestTool.UnitTests.Extensions;
+
+public static class ApiGatewayResponseTestCases
+{
+    public static IEnumerable<object[]> V1TestCases()
```
GarrettBeatty, line 12: these are all the same. the only minor change that i did (as mentioned in the pr description) is remove the unneeded "Assertions" functions

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/Extensions/HttpContextExtensionsTests.cs
```diff
@@ -1,250 +1,280 @@
...
+        {
+            _snapshots = new SnapshotTestHelper();
+        }
+
         [Theory]
-        [MemberData(nameof(HttpContextTestCases.V1TestCases), MemberType = typeof(HttpContextTestCases))]
+        [MemberData(nameof(V1TestCases), MemberType = typeof(HttpContextTestCases))]
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "xUnit1026:Theory methods should use all of their parameters")]
-        public async Task ToApiGatewayRequestRest_ConvertsCorrectly(string testName, HttpContextTestCase testCase)
+        public Task IntegrationTest_APIGatewayV1_REST(string testName, HttpContextTestCase testCase)
```
GarrettBeatty, line 27: reminder for self to fix this

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/ApiGatewayTestHelper.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            // Convert response to HttpResponse (simulates what API Gateway would do)
+            var convertedResponse = await ConvertToHttpResponseAsync(response);
+
+            // Load the expected response from snapshot
+            var expectedResponse = await _snapshots.LoadSnapshot<HttpResponseMessage>(snapshotName);
+
+            // Compare the responses
+            await AssertResponsesEqual(expectedResponse, convertedResponse);
+        }
```
philasmar, line 50: nit: these 2 functions are very close, they can be combined into a single function

Tools/LambdaTestTool-v2/tests/Amazon.Lambda.TestTool.UnitTests/SnapshotHelper/ApiGatewayTestSnapshot.cs
philasmar, line 7: add docs explaining what this is for and what is is doing

GarrettBeatty: added docs in latest commit

philasmar (approved):

normj (approved):

PR #2087, opened 2025-06-12
feat: add .NET 10 preview Lambda container image

philasmar:
*Issue #, if available:*
DOTNET-8071

*Description of changes:*
* Moved self-mutating step from each OCI pipeline to a parent pipeline.  This will allow us to add new pipelines in the future with a PR instead of requiring an MCM. 
image
* Added .NET 10 images
* Updated the GitHub action to add .NET 10 as well as fix a bug in the script which didn't work for .NET preview versions
* Updated Docker images to the latest .NET patches 8.0.17, 9.0.6, 10.0.0-preview.5
* Removed .NET 6 images
* .NET 10 images use the bootstrap.sh version with the SSL certificate updates which were required for .NET 8/9
* Moved .NET version configurations to code instead of environment variables. This will allow us to add/remove .NET versions using a PR instead of an MCM. 


GarrettBeatty (commented):
is there a different branch for when it gets deployed into staging vs prod  account?

PR #2095, opened 2025-06-24
feat: add ability to save requests and toggle ui elements

philasmar:
*Issue #, if available:*
DOTNET-8176

*Description of changes:*
* Added a button to save requests to a specified path provided via a command line argument
* Users will be unable to save requests if the path is not supplied via the command line argument
* Added a settings modal to allow users to:
  * Show/hide sample requests 
  * Show/hide saved requests 
  * Show/hide requests list
  * Delete saved requests

### DIFF
--- a/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/Settings/RunCommandSettings.cs
+++ b/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/Settings/RunCommandSettings.cs
@@ -78,6 +78,11 @@ public sealed class RunCommandSettings : CommandSettings
         if (string.IsNullOrEmpty(ConfigStoragePath))
             return ValidationResult.Success();
 
+        if (!Path.IsPathFullyQualified(ConfigStoragePath))
+        {
+            ConfigStoragePath = Path.Combine(Environment.CurrentDirectory, ConfigStoragePath);
+        }
+
         return !Path.IsPathFullyQualified(ConfigStoragePath)
             ? ValidationResult.Error("'Config storage path' must be an absolute path.")
             : ValidationResult.Success();


Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/Settings/RunCommandSettings.cs
```diff
@@ -61,4 +61,11 @@ public sealed class RunCommandSettings : CommandSettings
     [CommandOption("--sqs-eventsource-config <CONFIG>")]
     [Description("The configuration for the SQS event source. The format of the config is a comma delimited key pairs. For example \"QueueUrl=<queue-url>,FunctionName=<function-name>,VisibilityTimeout=100\". Possible keys are: BatchSize, DisableMessageDelete, FunctionName, LambdaRuntimeApi, Profile, QueueUrl, Region, VisibilityTimeout")]
     public string? SQSEventSourceConfig { get; set; }
+
+    /// <summary>
+    /// The path used to save requests to.
+    /// </summary>
+    [CommandOption("--saved-requests-path <SAVED-REQUESTS-PATH>")]
+    [Description("The path used to save requests to.")]
+    public string? SavedRequestsPath { get; set; }
```
GarrettBeatty: do you think we should have a default value of saved requests to the current directory?

GarrettBeatty: i guess the other parameters like port we make them provide, so providing saved request path is fine then

philasmar: This is intentional. We don't want a default value for it.

normj: If I set this to `c:\temp\myrequests` for example I would expect that is the root directory of the save requests but instead I get 2 more levels of sub folders `.aws-lambda-test-tool\SavedRequests`. I think `.aws-lambda-test-tool` only makes sense if we choosing a default directory. Like we were automatically putting the folder on the home directory or when integrate with Aspire we could default to `.aws-lambda-test-tool` in the AppHost. Since the user told us where to put the requests we don't need another layer.

Since we are also storing `settings.json` file as well as request then I can see `SavedRequests` making sense but that means this parameter means more then then just saving requests and we should name it more general. Let the description for the parameter inform users set this for allowing saved requests. Maybe something like `--config-storage-path`.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Components/Pages/SaveRequestDialog.razor.cs
```diff
@@ -0,0 +1,63 @@
...
+    private string? _functionName;
+    private string? _requestBody;
+    private string? _requestName;
+
+    /// <summary>
+    /// Used to set required properties for the save request dialog.
+    /// </summary>
+    /// <param name="functionName">Lambda function name</param>
+    /// <param name="requestBody">The body of the request to be saved.</param>
+    public void ShowDialog(string functionName, string? requestBody)
```
normj, line 44: `_requestName` should be reset to null or empty string otherwise when you save a second unrelated request the dialog defaults to the previous name.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Components/Pages/SaveRequestDialog.razor.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    /// <summary>
+    /// Used to save requests to a specified location.
+    /// </summary>
+    public async Task OnSaveClick()
+    {
+        if (string.IsNullOrEmpty(_functionName))
+            throw new InvalidSaveRequestException($"The Lambda function name is null or empty.");
+
+        if (string.IsNullOrEmpty(_requestName))
```
normj: Add some validation that request name is a valid file name. When I tried to save using a name like `?????????` I got an unhandled error.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Components/Pages/ManageSavedRequestsDialog.razor
```diff
@@ -0,0 +1,56 @@
...
+                                <input class="form-check-input" type="checkbox" @bind="@_showSavedRequests" @bind:after="ReloadSampleRequests" id="_showSavedRequests">
+                                <label class="form-check-label text-nowrap" for="_showSavedRequests">
+                                    Saved requests
+                                </label>
+                            </div>
+                        }
+                        <div class="form-check">
+                            <input class="form-check-input" type="checkbox" @bind="@_showRequestsList" @bind:after="ReloadSampleRequests" id="_showRequestsList">
+                            <label class="form-check-label text-nowrap" for="_showRequestsList">
+                                Requests list
```
normj: nit: Not sure what to call this but the term "Request List" confused me and didn't know what it did till I checked and unchecked it.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/GlobalSettingsService.cs
```diff
@@ -0,0 +1,86 @@
...
+        try
+        {
+            var newSettings = _currentSettings.DeepCopy(); // Create a mutable copy to update
+            updateAction(newSettings); // Apply changes
+
+            if (!AreSettingsEqual(_currentSettings, newSettings))
+            {
+                _currentSettings = newSettings;
+                logger.LogInformation("Global settings updated.");
+                _ = settingsRepository.SaveSettingsAsync(_currentSettings); // Fire-and-forget save
```
normj: You are sort of breaking your lock here by doing a fire and forget. You could exit the lock before `SaveSettingsAsync` completes and then another thread could call the `LoadSettingsAsync` method in this class acquire the lock and attempt to read at the same time the save is happening.

A corner case but not sure if you gain any perceived performance improvement by doing a fire and forget and you are opening it up to other potential issues.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/GlobalSettingsService.cs
```diff
@@ -0,0 +1,86 @@
...
+               s1.ShowRequestsList == s2.ShowRequestsList;
+    }
+
+    /// <summary>
+    /// Dispose of <see cref="GlobalSettingsService"/>.
+    /// </summary>
+    public async ValueTask DisposeAsync()
+    {
+        _settingsLock.Dispose();
+        await Task.CompletedTask;
```
normj: nit: Since this method returns `ValueTask` you should return `ValueTask.CompletedTask`. Returning `Task.CompletedTask` does a runtime cast conversion to `ValueTask` which is not needed.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/IGlobalSettingsService.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// Current value of <see cref="GlobalSettings"/>.
+    /// </summary>
+    GlobalSettings CurrentSettings { get; }
+
+    /// <summary>
+    /// Update the <see cref="GlobalSettings"/>.
+    /// </summary>
+    /// <param name="updateAction"></param>
+    /// <returns></returns>
+    Task UpdateSettings(Action<GlobalSettings> updateAction);
```
normj: nit: `UpdateSettings` -> `UpdateSettingsAsync` for consistency.

normj (changes requested):

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/GlobalSettingsService.cs
philasmar: In my earlier implementation, I was using the browser's local storage where i needed to do a fire and forget. When I settled on saving to disk, i forgot to go back and change it. I have updated the code to await.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Components/Pages/ManageSavedRequestsDialog.razor
philasmar: I went with "Request dropdown". Not sure if it's much better, but naming is hard :/

.autover/changes/7403aeda-5b1c-4ce1-a88b-d15894f9e6d0.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.TestTool",
+      "Type": "Minor",
+      "ChangelogMessages": [
+        "Add the ability to save requests and hide UI elements"
```
GarrettBeatty, line 7: can you update the readme with steps for how to do saved requests

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/Settings/RunCommandSettings.cs
```diff
@@ -61,4 +61,11 @@ public sealed class RunCommandSettings : CommandSettings
     [CommandOption("--sqs-eventsource-config <CONFIG>")]
     [Description("The configuration for the SQS event source. The format of the config is a comma delimited key pairs. For example \"QueueUrl=<queue-url>,FunctionName=<function-name>,VisibilityTimeout=100\". Possible keys are: BatchSize, DisableMessageDelete, FunctionName, LambdaRuntimeApi, Profile, QueueUrl, Region, VisibilityTimeout")]
     public string? SQSEventSourceConfig { get; set; }
+
+    /// <summary>
+    /// The path used to save requests to.
+    /// </summary>
+    [CommandOption("--saved-requests-path <SAVED-REQUESTS-PATH>")]
```
GarrettBeatty: i tried running with 

```
dotnet lambda-test-tool start --lambda-emulator-port 5050 --saved-requests-path .
```

but it didnt seem to save requests in the folder where i am running the tool from (or maybe it gets saved to the test tool nuget package root or something?) Either way can we add validation/make it clear that it needs to be an absolute path

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Components/Pages/Home.razor
```diff
@@ -43,27 +41,39 @@ else
     <div class="row">
         <div class="col-lg-6 gap-2 d-flex flex-column mt-1 mb-3">
             <div class="form-group d-flex">
-                <h4>Queue Event</h4>
-                <button class="btn btn-primary btn-sm ms-auto rounded-pill px-4 fw-bold d-flex align-items-center gap-1" @onclick="OnAddEventClick">
-                    <i class="bi bi-play-fill"></i>
-                    Invoke
-                </button>
+                <h4 class="m-0">Queue Event</h4>
```
GarrettBeatty, line 44: after i save a request, i think it would be useful to have a toast pop up/message saying "saved request successfully" or something to let the user know it worked

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/Settings/RunCommandSettings.cs
philasmar: Fixed

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Components/Pages/SaveRequestDialog.razor.cs
philasmar: Fixed

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Components/Pages/SaveRequestDialog.razor.cs
philasmar: Fixed

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/GlobalSettingsService.cs
philasmar: Fixed

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Services/IGlobalSettingsService.cs
philasmar: Fixed

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Components/Pages/ManageSavedRequestsDialog.razor
GarrettBeatty: i also got confused with the request list - wondering if this option is even needed. i feel like this option this might confuse the user even more. should we just remove it?

.autover/changes/7403aeda-5b1c-4ce1-a88b-d15894f9e6d0.json
philasmar: Done

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/Settings/RunCommandSettings.cs
philasmar: Added validation and improved wording about absolute path

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Components/Pages/Home.razor
philasmar: Added

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Components/Pages/ManageSavedRequestsDialog.razor
philasmar: I prefer to have it. I updated it's name.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/Settings/RunCommandSettings.cs
```diff
@@ -61,4 +62,24 @@ public sealed class RunCommandSettings : CommandSettings
...
+    /// <summary>
+    /// Validate that <see cref="ConfigStoragePath"/> is an absolute path.
+    /// </summary>
+    public override ValidationResult Validate()
+    {
+        if (string.IsNullOrEmpty(ConfigStoragePath))
+            return ValidationResult.Success();
+
+        return !Path.IsPathFullyQualified(ConfigStoragePath)
+            ? ValidationResult.Error("'Config storage path' must be an absolute path.")
```
normj, line 87: Why does the path have to be fully qualified? That seems really unintuitive and limiting.

normj (changes requested):

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Commands/Settings/RunCommandSettings.cs
philasmar: because as Garrett mentioned, using . starts from the nuget package and not the current folder.

normj: I was able to reproduce what @GarrettBeatty saw when I set the location to '.' but when I changed this code to resolve the relative path based on the current working directory it worked as expected.

```csharp
public override ValidationResult Validate()
{
    if (string.IsNullOrEmpty(ConfigStoragePath))
        return ValidationResult.Success();

    if (!Path.IsPathFullyQualified(ConfigStoragePath))
    {
        ConfigStoragePath = Path.Combine(Environment.CurrentDirectory, ConfigStoragePath);
    }

    return !Path.IsPathFullyQualified(ConfigStoragePath) && false
        ? ValidationResult.Error("'Config storage path' must be an absolute path.")
        : ValidationResult.Success();
}
```

philasmar: Why is there  && false ?

normj: Sorry that was a leftover artifact when I wanted to recreate what @GarrettBeatty saw. Ignore it. All I really wanted to show was you can fix the issue by combining with the current directory. I'll leave it up to you if the `Validate` is the right place for this code.

philasmar: I updated it and left it in the `Validate` function since that'll let Spectre handle the validation error.

normj (approved):

PR #2112, opened 2025-07-16
auto update lambda runtime images

GarrettBeatty:
*Issue #, if available:*

This script will run daily to update the docker files. example pr it made https://github.com/aws/aws-lambda-dotnet/pull/2127/files (ignore the extra files that are in this PR. I had to make the branch runtim2 for testing so thats why it picked up extra changes). in reality it wont have the extra files

### DIFF
--- a/.github/workflows/auto-update-Dockerfiles.yml
+++ b/.github/workflows/auto-update-Dockerfiles.yml
@@ -1,0 +1,175 @@
+name: Auto-Update Lambda Dockerfiles Monthly
+
+on:
+  # Run monthly on the 1st day of each month at midnight UTC
+  schedule:
+    - cron: '0 0 1 * *'
+  # Allows to run this workflow manually from the Actions tab for testing
+  workflow_dispatch:
+
+jobs:
+  auto-update:
+    runs-on: ubuntu-latest
+    env:
+      NET_8_AMD64_Dockerfile: "LambdaRuntimeDockerfiles/Images/net8/amd64/Dockerfile"
+      NET_8_ARM64_Dockerfile: "LambdaRuntimeDockerfiles/Images/net8/arm64/Dockerfile"
+      NET_9_AMD64_Dockerfile: "LambdaRuntimeDockerfiles/Images/net9/amd64/Dockerfile"
+      NET_9_ARM64_Dockerfile: "LambdaRuntimeDockerfiles/Images/net9/arm64/Dockerfile"
+      NET_10_AMD64_Dockerfile: "LambdaRuntimeDockerfiles/Images/net10/amd64/Dockerfile"
+      NET_10_ARM64_Dockerfile: "LambdaRuntimeDockerfiles/Images/net10/arm64/Dockerfile"
+
+    steps:
+      # Checks-out the repository under $GITHUB_WORKSPACE
+      - uses: actions/checkout@85e6279cec87321a52edac9c87bce653a07cf6c2 #v4.2.2
+        with:
+          ref: 'dev'
+
+      # Determine the latest ASP.NET Core versions for all .NET versions
+      - name: Determine Latest ASP.NET Core Versions
+        id: get-versions
+        shell: pwsh
+        run: |
+          $versions = ./LambdaRuntimeDockerfiles/get-latest-aspnet-versions.ps1
+          foreach ($line in $versions) {
+            if ($line -match '(.+)=(.+)') {
+              echo "$($matches[1])=$($matches[2])" >> $env:GITHUB_OUTPUT
+              echo "Using $($matches[1])=$($matches[2])"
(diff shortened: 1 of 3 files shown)


.github/workflows/auto-update-Dockerfiles.yml
```diff
@@ -0,0 +1,195 @@
+name: Auto-Update Lambda Dockerfiles Monthly
+
+permissions:
+  contents: write
+  pull-requests: write
+
+on:
+  # Run monthly on the 1st day of each month at midnight UTC
+  schedule:
+    - cron: '0 0 1 * *'
```
philasmar: we should run this daily. We don't want to be a month late for the patches. Our goal is to get the updated OCIs to Lambda as quickly as possible.

LambdaRuntimeDockerfiles/get-latest-aspnet-versions.ps1
```diff
@@ -0,0 +1,147 @@
...
+}
+
+if (-not $allVersionsValid) {
+    exit 1
+}
+
+# Output as GitHub Actions environment variables
+Write-Output "NET_8_NEXT_VERSION=$net8Version"
+Write-Output "NET_9_NEXT_VERSION=$net9Version"
+Write-Output "NET_10_NEXT_VERSION=$net10Version"
```
philasmar: Instead of you outputting the versions in this ps1, let's call it from the github workflow in each .NET version's step. That way, this ps1 stays generic and wouldn't need updating everytime we add or remove a .NET version.

.github/workflows/auto-update-Dockerfiles.yml
```diff
@@ -0,0 +1,195 @@
...
+        id: get-versions
+        shell: pwsh
+        run: |
+          $versions = .\LambdaRuntimeDockerfiles\get-latest-aspnet-versions.ps1
+          foreach ($line in $versions) {
+            if ($line -match '(.+)=(.+)') {
+              echo "$($matches[1])=$($matches[2])" >> $env:GITHUB_OUTPUT
+              echo "Using $($matches[1])=$($matches[2])"
+            }
+          }
```
philasmar: you wouldn't need this step if you update the ps1 based on my recommendation

.github/workflows/auto-update-Dockerfiles.yml
```diff
@@ -0,0 +1,195 @@
...
+            # Write the branch name to GITHUB_OUTPUT for use in the PR step
+            Add-Content -Path $env:GITHUB_OUTPUT -Value "BRANCH=$branch"
+            Add-Content -Path $env:GITHUB_OUTPUT -Value "CHANGES_MADE=true"
+            echo "Changes committed and pushed to branch $branch"
+          } else {
+            echo "No changes detected in Dockerfiles, skipping PR creation"
+          }
+
+      # Create a Pull Request
+      - name: Create Pull Request
```
philasmar, line 150: What would happen if we update this workflow to run daily and we don't approve and merge the created PR that same day? would this step keep creating PRs?

philasmar (changes requested):

LambdaRuntimeDockerfiles/get-latest-aspnet-versions.ps1
GarrettBeatty: fixed

.github/workflows/auto-update-Dockerfiles.yml
GarrettBeatty: fixed to use same branch

.github/workflows/auto-update-Dockerfiles.yml
```diff
@@ -0,0 +1,192 @@
...
+        run: |
+          # Check if there are any changes to commit
+          if (git status --porcelain) {
+            git config --global user.email "github-aws-sdk-dotnet-automation@amazon.com"
+            git config --global user.name "aws-sdk-dotnet-automation"
+            $branch="chore/auto-update-Dockerfiles-daily"
+            
+            # Check if the branch already exists remotely
+            $branchExists = git ls-remote --heads origin $branch
+            if ($branchExists) {
```
philasmar: im not a fan of the branch exists approach. we are not certain that the changes in that branch are what we want them to be. we shouldn't just add more to it. I'd say let's always create a unique branch by appending a timestamp to the name.

philasmar (changes requested):

.github/workflows/auto-update-Dockerfiles.yml
GarrettBeatty: fixed in e679308

philasmar (approved):

AlexDaines (approved):

PR #2141, opened 2025-09-19
Add configuration for staging repo

GarrettBeatty:
Add configuration for staging repo. Please note that this change does not include changes to auto fetch the PAT when it rotates. That will be in a follow up PR. i wasnt able to test end to end fully in my personal account (building the images and publishing to ecr) because my account didnt have permissions. but this change only modifies the pipeline creation so it should work.

In the below picture we can see the staging pipelines get created (the one is in failed state because the webhook limit of 20 was reached but that is fine for testing)

image

image

### DIFF
--- a/LambdaRuntimeDockerfiles/Infrastructure/src/Infrastructure/Configuration.cs
+++ b/LambdaRuntimeDockerfiles/Infrastructure/src/Infrastructure/Configuration.cs
@@ -29,9 +29,14 @@ internal class Configuration
     public string GitHubOwner { get; } = Environment.GetEnvironmentVariable("AWS_LAMBDA_GITHUB_REPO_OWNER");
     public string GitHubRepository { get; } = Environment.GetEnvironmentVariable("AWS_LAMBDA_GITHUB_REPO_NAME");
     public string GitHubBranch { get; } = Environment.GetEnvironmentVariable("AWS_LAMBDA_GITHUB_REPO_BRANCH");
+    
+    public string GitHubOwnerStaging { get; } = Environment.GetEnvironmentVariable("AWS_LAMBDA_GITHUB_REPO_OWNER_STAGING");
+    public string GitHubRepositoryStaging { get; } = Environment.GetEnvironmentVariable("AWS_LAMBDA_GITHUB_REPO_NAME_STAGING");
+    public string GitHubBranchStaging { get; } = Environment.GetEnvironmentVariable("AWS_LAMBDA_GITHUB_REPO_BRANCH_STAGING");
+    
     public Ecrs Ecrs { get; } = new Ecrs();
     public const string ProjectRoot = "LambdaRuntimeDockerfiles/Infrastructure/src/Infrastructure";
-    public static readonly string ProjectName = "aws-lambda-container-images";
+    public string ProjectName { get; } = "aws-lambda-container-images";
 
     public readonly FrameworkConfiguration[] Frameworks = new[]
     {
--- a/LambdaRuntimeDockerfiles/Infrastructure/src/Infrastructure/PipelinesStage.cs
+++ b/LambdaRuntimeDockerfiles/Infrastructure/src/Infrastructure/PipelinesStage.cs
@@ -38,6 +38,11 @@ internal class PipelinesStage : Stage
     {
         for (var i = 0; i < configuration.Frameworks.Length; i++)
         {
+
+            var pipelineName = string.IsNullOrEmpty(pipelinePrefix) 
+            ? $"{configuration.ProjectName}-{configuration.Frameworks[i].Framework}"  // "aws-lambda-container-images-net8"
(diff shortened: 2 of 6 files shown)


LambdaRuntimeDockerfiles/Infrastructure/src/Infrastructure/Configuration.cs
```diff
@@ -29,9 +29,14 @@ internal class Configuration
...
     public string GitHubBranch { get; } = Environment.GetEnvironmentVariable("AWS_LAMBDA_GITHUB_REPO_BRANCH");
+    
+    public string GitHubOwnerStaging { get; } = Environment.GetEnvironmentVariable("AWS_LAMBDA_GITHUB_REPO_OWNER_STAGING");
+    public string GitHubRepositoryStaging { get; } = Environment.GetEnvironmentVariable("AWS_LAMBDA_GITHUB_REPO_NAME_STAGING");
+    public string GitHubBranchStaging { get; } = Environment.GetEnvironmentVariable("AWS_LAMBDA_GITHUB_REPO_BRANCH_STAGING");
+    
     public Ecrs Ecrs { get; } = new Ecrs();
     public const string ProjectRoot = "LambdaRuntimeDockerfiles/Infrastructure/src/Infrastructure";
-    public static readonly string ProjectName = "aws-lambda-container-images";
+    public string ProjectName { get; } = "aws-lambda-container-images";
```
GarrettBeatty, line 39: made non static to be consistent

LambdaRuntimeDockerfiles/Infrastructure/src/Infrastructure/PipelinesStage.cs
```diff
@@ -13,13 +13,42 @@ public PipelinesStage(
...
+        string gitHubBranch)
     {
         for (var i = 0; i < configuration.Frameworks.Length; i++)
         {
+
+            var pipelineName = $"{pipelinePrefix}-{configuration.Frameworks[i].Framework}";
+
             new PipelineStack(this,
-                configuration.Frameworks[i].Framework,
+                $"{pipelinePrefix}-{configuration.Frameworks[i].Framework}",
```
GarrettBeatty, line 45: cloudformation seems to automatically uses the parent pipeline name+ this for the name of the stack and we get something like 

```
aws-lambda-container-images-aws-lambda-container-images-staging-net9
```

but i dont really care about the stack name too much to be honest. This was the only way i found to make the pipeline name + what appears in the ui for the pipeline in the main pipeline look correct

LambdaRuntimeDockerfiles/Infrastructure/src/Infrastructure/SelfMutatingPipelineStack.cs
```diff
@@ -65,9 +65,19 @@ internal SelfMutatingPipelineStack(
                 { "AWS_LAMBDA_DOTNET_FRAMEWORK_CHANNEL",
                     new BuildEnvironmentVariable { Type = BuildEnvironmentVariableType.PLAINTEXT, Value =
                     System.Environment.GetEnvironmentVariable("AWS_LAMBDA_DOTNET_FRAMEWORK_CHANNEL") ?? string.Empty } },
+                { "AWS_LAMBDA_GITHUB_REPO_OWNER_STAGING",
```
GarrettBeatty, line 68: i didnt bother to move all of these to a json file yet because we would still need to send in the token key as env variables anyway. i may come back to this in the future

LambdaRuntimeDockerfiles/Infrastructure/src/Infrastructure/Configuration.cs
```diff
@@ -29,9 +29,14 @@ internal class Configuration
     public string GitHubOwner { get; } = Environment.GetEnvironmentVariable("AWS_LAMBDA_GITHUB_REPO_OWNER");
     public string GitHubRepository { get; } = Environment.GetEnvironmentVariable("AWS_LAMBDA_GITHUB_REPO_NAME");
     public string GitHubBranch { get; } = Environment.GetEnvironmentVariable("AWS_LAMBDA_GITHUB_REPO_BRANCH");
+    
+    public string GitHubOwnerStaging { get; } = Environment.GetEnvironmentVariable("AWS_LAMBDA_GITHUB_REPO_OWNER_STAGING");
```
GarrettBeatty, line 33: i didnt bother to move all of these to a json file yet because we would still need to send in the token key as env variables anyway. i may come back to this in the future

LambdaRuntimeDockerfiles/Infrastructure/src/Infrastructure/PipelinesStage.cs
philasmar: im not a fan of this, but not a big deal

philasmar (approved):

AlexDaines (approved):

PR #2165, opened 2025-10-28
Add NotResource field for custom authorizer IAM policy

RWcoder:
*Issue #, if available:* n/a

*Description of changes:*
Added the option to use NotResource in the policy returned by a custom lambda authorizer for API Gateway. I have a use case where I am caching an API policy for multiple resources as described here: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html#http-api-lambda-authorizer.caching

And I would like to return a policy which uses Deny with NotResource

### DIFF
--- a/Libraries/src/Amazon.Lambda.APIGatewayEvents/Amazon.Lambda.APIGatewayEvents.csproj
+++ b/Libraries/src/Amazon.Lambda.APIGatewayEvents/Amazon.Lambda.APIGatewayEvents.csproj
@@ -6,7 +6,7 @@
     <TargetFrameworks>netstandard2.0;netcoreapp3.1;net8.0</TargetFrameworks>    
     <Description>Amazon Lambda .NET Core support - API Gateway package.</Description>
     <AssemblyTitle>Amazon.Lambda.APIGatewayEvents</AssemblyTitle>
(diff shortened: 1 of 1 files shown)


peterrsongg (approved):
looks good, approving assuming what norm asks for will be added.

Libraries/src/Amazon.Lambda.APIGatewayEvents/Amazon.Lambda.APIGatewayEvents.csproj
```diff
@@ -6,7 +6,7 @@
     <TargetFrameworks>netstandard2.0;netcoreapp3.1;net8.0</TargetFrameworks>    
     <Description>Amazon Lambda .NET Core support - API Gateway package.</Description>
     <AssemblyTitle>Amazon.Lambda.APIGatewayEvents</AssemblyTitle>
-    <Version>2.7.1</Version>
+    <Version>2.7.2</Version>
```
normj: Sorry, should have mentioned this before but can you revert this change. The automation using the change file will take care of version bump the package.

normj (changes requested):

normj (approved):

Issue #2166, opened 2025-10-28
Serialization of CognitoMigrateUserEvent throws


ThomasEg:
### Describe the bug

Having an event handler for user migration in Cognito:
```
[assembly: LambdaSerializer(typeof(Amazon.Lambda.Serialization.SystemTextJson.DefaultLambdaJsonSerializer))]
...
public void CognitoMigrateUserHandler(CognitoMigrateUserEvent cognitoEvent, ILambdaContext context)
{
   ...
}
```

Throws an exception during serialization of CognitoMigrateUserEvent:
```
Amazon.Lambda.Serialization.SystemTextJson.JsonSerializerException: 
Error converting the Lambda event JSON payload to type Amazon.Lambda.CognitoEvents.CognitoMigrateUserEvent: 
The JSON value could not be converted to System.Boolean. 
Path: $.response.forceAliasCreation | LineNumber: 0 | BytePositionInLine: 401.
```

The data sent from cognito was:
```
{
    "version": "1",
    "triggerSource": "UserMigration_Authentication",
    "region": "...",
    "userPoolId": "...",
    "userName": "...",
    "callerContext": {
        "awsSdkVersion": "...",
        "clientId": "..."
    },
    "request": {
        "password": "...",
        "validationData": null,
        "userAttributes": null
    },
    "response": {
        "userAttributes": null,
        "forceAliasCreation": null,
        "enableSMSMFA": null,
        "finalUserStatus": null,
        "messageAction": null,
        "desiredDeliveryMediums": null
    }
}
```

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

I expect it to serialize correctly and then call the method.

### Current Behavior

It throws a deserialization exception.

### Reproduction Steps

Setup a Cognito UserPool to have a `Migrate user trigger` with this target:
```
[assembly: LambdaSerializer(typeof(Amazon.Lambda.Serialization.SystemTextJson.DefaultLambdaJsonSerializer))]
...
public void CognitoMigrateUserHandler(CognitoMigrateUserEvent cognitoEvent, ILambdaContext context)
{
}
```

### Possible Solution

My guess is that the types of the properties in the `response`-object needs to be nullable.

### AWS .NET SDK and/or Package version used

Amazon.Lambda.CognitoEvents: 4.0.0
Amazon.Lambda.Serialization.SystemTextJson:2.4.4

### Targeted .NET Platform

.NET 8

### Operating System and version

[...]

PR #2170, opened 2025-10-29
Make ForceAliasCreation nullable

peterrsongg:
*Issue #, if available:*
Fix Issue #2166 
*Description of changes:*

I validated the change by flipping the `ForceAliasCreation` value in `\Libraries\test\EventsTests.Shared\cognito-migrateuser-event.json` to null and seeing that it was serialized properly (to null).

### DIFF
--- a/.autover/changes/1916e292-d131-4c9a-8036-0d7b68f0ce7a.json
+++ b/.autover/changes/1916e292-d131-4c9a-8036-0d7b68f0ce7a.json
@@ -4,7 +4,7 @@
       "Name": "Amazon.Lambda.CognitoEvents",
       "Type": "Patch",
       "ChangelogMessages": [
-        "Make ForceAliasCreation nullable to allow proper serialization"
+        "[Breaking Change] Fix issue around not handling null value during deserialization for `ForceAliasCreation` by making it nullable."
       ]
     }
   ]


.autover/changes/1916e292-d131-4c9a-8036-0d7b68f0ce7a.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.CognitoEvents",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Make ForceAliasCreation nullable to allow proper serialization"
```
normj: Add `[Breaking Change]` as a prefix. It is a very minor breaking change but we should be as transparent as policy. Maybe rephrase the message as more of a bug fix then just stating a fact. Possible option.
```
[Breaking Change] Fixed issue not handling null value during deserialization for `ForceAliasCreation` by making it nullable.
```

peterrsongg: done

ThomasEg:
I'm the OP of #2166 and i've tried the code in this PR from @peterrsongg in my original codebase and can confirm that it solves the issue.

philasmar (approved):

PR #2217, opened 2025-12-05
Marked ProxyRequestContext.Error property as Obsolete for APIGatewayProxyRequest.

ashishdhingra:
*Issue #, if available:* P340365477

*Description of changes:*
Marked `ProxyRequestContext.Error` property as Obsolete for `APIGatewayProxyRequest`.
___

### DIFF
--- a/Libraries/src/Amazon.Lambda.APIGatewayEvents/APIGatewayProxyRequest.cs
+++ b/Libraries/src/Amazon.Lambda.APIGatewayEvents/APIGatewayProxyRequest.cs
@@ -1,4 +1,4 @@
-﻿namespace Amazon.Lambda.APIGatewayEvents
+namespace Amazon.Lambda.APIGatewayEvents
 {
     using System;
     using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


GarrettBeatty (approved):

Libraries/src/Amazon.Lambda.APIGatewayEvents/APIGatewayProxyRequest.cs
```diff
@@ -234,10 +234,11 @@ public class ProxyRequestContext
             /// Gets and sets the operation name.
             /// </summary>
             public string OperationName { get; set; }
-            
+
             /// <summary>
             /// Gets and sets the error.
             /// </summary>
+            [Obsolete("This property is obsolete since it is not processed by the Velocity Template Language engine and is not available as a field to integration request.")]
```
GarrettBeatty: is there a non obsolete field to recommend instead and if so can you mention that

ashishdhingra: No, there is no non-obsolete field. This property should not be used at all for integration. Please refer ticket (ID in description).

normj: I suspect most users wouldn't know the context of "Velocity Template", I had to look it up myself. I suggest keeping the error message a bit more generic that it just isn't used by the service.

```
This property is obsolete since it is not used by AWS API Gateway and will never have value set by the service.
```

ashishdhingra: Done

normj (approved):

PR #2219, opened 2025-12-08
Prepare Amazon.Lambda.Annotations for upcoming .NET 10 managed runtime

normj:
*Description of changes:*
With the upcoming .NET 10 managed runtime the mapping and allowed runtimes for Amazon.Lambda.Annotations need to be updated to support .NET 10.


Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Generator.cs
```diff
@@ -16,7 +16,6 @@ namespace Amazon.Lambda.Annotations.SourceGenerator
     [Generator]
     public class Generator : ISourceGenerator
     {
-        private const string DEFAULT_LAMBDA_SERIALIZER = "Amazon.Lambda.Serialization.SystemTextJson.DefaultLambdaJsonSerializer";
```
normj, line 19: Removed because it wasn't ever referenced.

PR #2251, opened 2026-01-10
Fix missing entry point exception with Amazon.Lambda.TestTool

normj:
*Issue #, if available:*
https://github.com/aws/integrations-on-dotnet-aspire-for-aws/issues/138

*Description of changes:*
While using a Lambda Project that is both an executable like ASP.NET Core Minimal API but also has some class library based Lambda functions the test tool would end up picking up the version of Amazon.Lambda.RuntimeSupport that is references in the project instead of the version shipped with the test tool. The class library like Lambda functions would fail to start because the referenced version of Amazon.Lambda.RuntimeSupport was compiled as a library without an entry point, meaning not an executable.

The fix is to only build RuntimeSupport one way which is an executable. Setting the `OutputType` to `Exe` for all builds of Amazon.Lambda.RuntimeSupport just adds a bit of extra assembly metadata but it does not prevent it from being packages as a dll in a NuGet package. 

I have built local versions of the NuGet packages with RuntimeSupport and build and deploy and tested with Amazon.Lambda.TestTool and everything work as expected. I confirmed with an ASP.NET Core minimal API Lambda function with extra class library style Lambda functions.

### DIFF
--- a/Libraries/src/Amazon.Lambda.RuntimeSupport/Amazon.Lambda.RuntimeSupport.csproj
+++ b/Libraries/src/Amazon.Lambda.RuntimeSupport/Amazon.Lambda.RuntimeSupport.csproj
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net6.0;net8.0;net9.0;net10.0</TargetFrameworks>
-    <Version>1.14.1112</Version>
+    <Version>1.14.1</Version>
     <Description>Provides a bootstrap and Lambda Runtime API Client to help you to develop custom .NET Core Lambda Runtimes.</Description>
     <AssemblyTitle>Amazon.Lambda.RuntimeSupport</AssemblyTitle>
     <AssemblyName>Amazon.Lambda.RuntimeSupport</AssemblyName>
--- a/Libraries/src/Amazon.Lambda.RuntimeSupport/Amazon.Lambda.RuntimeSupport.csproj
+++ b/Libraries/src/Amazon.Lambda.RuntimeSupport/Amazon.Lambda.RuntimeSupport.csproj
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net6.0;net8.0;net9.0;net10.0</TargetFrameworks>
-    <Version>1.14.1</Version>
+    <Version>1.14.1112</Version>
     <Description>Provides a bootstrap and Lambda Runtime API Client to help you to develop custom .NET Core Lambda Runtimes.</Description>
     <AssemblyTitle>Amazon.Lambda.RuntimeSupport</AssemblyTitle>
     <AssemblyName>Amazon.Lambda.RuntimeSupport</AssemblyName>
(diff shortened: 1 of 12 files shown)


Libraries/src/Amazon.Lambda.RuntimeSupport/Amazon.Lambda.RuntimeSupport.csproj
```diff
@@ -22,11 +22,21 @@
...
+    executable project that also has some class library like style functions. In that scenario Amazon.Lambda.RuntimeSupport
+    would come from project instead of Amazon.Lambda.TestTool when testing. If we didn't compile as a executable
+    the user gets a missing entry point exception.
+    https://github.com/aws/integrations-on-dotnet-aspire-for-aws/issues/138
+    -->
+    <OutputType>Exe</OutputType>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(ExecutableOutputType)'=='true' ">
+  <PropertyGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
```
philasmar: This entire property group is now redundant, no? Since we always set OutputType to Exe on line 35. Even on `netstandard2.0`, OutputType will be Exe.

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Amazon.Lambda.TestTool.csproj
```diff
@@ -15,7 +15,7 @@
     <PackAsTool>true</PackAsTool>
     <PackageId>Amazon.Lambda.TestTool</PackageId>
     <ToolCommandName>dotnet-lambda-test-tool</ToolCommandName>
-    <Version>0.11.1</Version>
+    <Version>0.11.0</Version>
```
philasmar, line 18: Why are you doing a version downgrade?

buildtools/build.proj
```diff
@@ -170,7 +170,6 @@
     <Target Name="build-lambda-test-tool-package" DependsOnTargets="init">
         <Exec Command="dotnet msbuild aws-lambda-test-tool-netcore.sln /t:Rebuild /p:Configuration=$(Configuration) /p:AssemblyOriginatorKeyFile=$(AssemblyOriginatorKeyFile) /p:SignAssembly=$(SignAssembly)" WorkingDirectory="..\Tools\LambdaTestTool"/>	
 		<Exec Command="$(PackCommand) Amazon.Lambda.TestTool.BlazorTester60-pack.csproj" WorkingDirectory="..\Tools\LambdaTestTool\src\Amazon.Lambda.TestTool.BlazorTester"/>		
-		<Exec Command="$(PackCommand) Amazon.Lambda.TestTool.BlazorTester70-pack.csproj" WorkingDirectory="..\Tools\LambdaTestTool\src\Amazon.Lambda.TestTool.BlazorTester"/>
```
philasmar, line 173: Why are we only removing the Net7 version only?

philasmar (changes requested):

Libraries/src/Amazon.Lambda.RuntimeSupport/Amazon.Lambda.RuntimeSupport.csproj
normj: Removed the redundant section

buildtools/build.proj
normj: Because .NET 7 csproj file was actually deleted from the repo. This is a msbuild target used for local builds not what the CI system uses and it had gotten stale. It was currently failing because we previously deleted the .NET 7 version.

PR #2286, opened 2026-02-18
Update AWSLambdaPSCore to target .NET 10 managed runtime

normj:
*Description of changes:*
Following the pattern done when the [.NET 8 managed runtime](https://github.com/aws/aws-lambda-dotnet/pull/1674) was released this updates the AWSLambdaPSCore module to use the .NET 10 runtime. This is a major version update like we did for previous new runtimes. Changes include

* Updated constant values to target .NET 10
* Update NuGet dependencies to latest versions
* Update the version of AWS.Tools modules being referenced in the templates
* Update README file

There is no automatic versioning for this module so I manually updated the version number and added the change log entry.

### Testing
I ran the tests for the module and did E2E tests deploying a PowerShell function to Lambda using the .NET 10 runtime

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Release 2026-02-18
+
+### AWSLambdaPSCore PowerShell Module (5.0.0)
+* Updated to target the .NET 10 Lambda runtime
+
 ## Release 2026-02-10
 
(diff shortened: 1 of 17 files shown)


normj:
The "Release Not Needed" tag was added because the PS module doesn't use the versioning tool in the repo.

PR #2318, opened 2026-04-01
Application Load Balancer (ALB)

GarrettBeatty:
## Application Load Balancer (ALB) Support for Lambda Annotations

https://github.com/aws/aws-lambda-dotnet/discussions/2269

### Overview

This PR adds a new `[ALBApi]` attribute to the Lambda Annotations framework, enabling developers to configure Lambda functions as targets behind an Application Load Balancer using the same declarative, attribute-driven approach already available for API Gateway (`[HttpApi]`, `[RestApi]`) and SQS (`[SQSEvent]`).

### User Experience

Developers can now wire a Lambda function to an ALB listener with a single attribute:

```csharp
[LambdaFunction(ResourceName = "ALBHello", MemorySize = 256, Timeout = 15)]
[ALBApi("@MyListener", "/hello", 1)]
public ApplicationLoadBalancerResponse Hello(ApplicationLoadBalancerRequest request, ILambdaContext context)
{
    return new ApplicationLoadBalancerResponse
    {
        StatusCode = 200,
        Headers = new Dictionary<string, string> { { "Content-Type", "application/json" } },
        Body = "{\"message\": \"Hello from ALB Lambda!\"}"
    };
}
```

The source generator automatically creates the necessary CloudFormation resources — `AWS::Lambda::Permission`, `AWS::ElasticLoadBalancingV2::TargetGroup`, and `AWS::ElasticLoadBalancingV2::ListenerRule` — and keeps them in sync with the code.

### `ALBApi` Attribute Properties

| Property | Required | Description |
|---|---|---|
| `ListenerArn` | ✅ | ALB listener ARN or `@ResourceName` template reference |
| `PathPattern` | ✅ | Path pattern for routing (e.g., `"/api/orders/*"`) |
| `Priority` | ✅ | Listener rule priority (1–50000) |
| `MultiValueHeaders` | | Enable multi-value headers on the target group |
| `HostHeader` | | Host header condition for the listener rule |
| `HttpMethod` | | HTTP method condition (e.g., `"GET"`, `"POST"`) |
| `ResourceName` | | Custom CloudFormation resource name prefix |

### Key Design Decisions

- **Standalone CloudFormation resources** — Unlike API Gateway events which map to SAM event types under the function's `Events` property, ALB integration requires generating separate top-level CloudFormation resources (TargetGroup, ListenerRule, Permission). This is because AWS SAM does not have a built-in ALB event type.
- **Existing listener required** — The attribute references an ALB listener that must already exist (either in the same template or via ARN). It does not create the ALB/VPC infrastructure.
- **`@ResourceName` pattern** — Follows the same convention as other annotations attributes for referencing CloudFormation resources using `Ref`.

### Testing

- 60 ALB-related unit tests passing (attribute validation, model, writer, snapshot)
- Integration test project deploys successfully to AWS with VPC, ALB, listener, and Lambda targets


Libraries/src/Amazon.Lambda.Annotations/README.md
```diff
@@ -852,6 +853,226 @@ The following SQS event source mapping will be generated for the `SQSMessageHand
...
 
+## Application Load Balancer (ALB) Example
+
+This example shows how to use the `ALBApi` attribute to configure a Lambda function as a target behind an [Application Load Balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html). Unlike API Gateway event attributes that map to SAM event types, the ALB integration generates standalone CloudFormation resources — a `TargetGroup`, a `ListenerRule`, and a `Lambda::Permission` — to wire the Lambda function to an existing ALB listener.
+
+The `ALBApi` attribute contains the following properties:
+
+| Property | Type | Required | Default | Description |
+|---|---|---|---|---|
+| `ListenerArn` | `string` | Yes | — | The ARN of the existing ALB listener, or a `@ResourceName` reference to a listener resource defined in the CloudFormation template. |
```
GarrettBeatty, line 864: The reason we specify the listener instead of the actual ALB is because an ALB can have multiple listeners, and the listener is the one that has all of the configuration values

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Validation/LambdaFunctionValidator.cs
```diff
@@ -268,6 +280,48 @@ private static void ValidateSqsEvents(LambdaFunctionModel lambdaFunctionModel, L
...
+                    continue;
+
+                var albApiAttribute = ((AttributeModel<ALBApiAttribute>)att).Data;
+                var validationErrors = albApiAttribute.Validate();
+                validationErrors.ForEach(errorMessage => diagnostics.Add(Diagnostic.Create(DiagnosticDescriptors.InvalidAlbApiAttribute, methodLocation, errorMessage)));
+            }
+
+            // Validate method parameters - When using ALBApiAttribute, the method signature must be
+            // (ApplicationLoadBalancerRequest request) or (ApplicationLoadBalancerRequest request, ILambdaContext context)
+            var parameters = lambdaFunctionModel.LambdaMethod.Parameters;
```
normj, line 305: The ALB attributes should act like the API Gateway attributes supporting all of the `FromX` attributes and by supporting that the customer function can have any number of parameters. Althought a difference I see is ALB does not support the idea of resource path templates like API Gateway does. So you might have to exclude that and if there are any parameters not mapping to something that should cause a diagnostic error.

Libraries/src/Amazon.Lambda.Annotations/ALB/ALBApiAttribute.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// The source generator will create the necessary CloudFormation resources
+    /// (TargetGroup, ListenerRule, Lambda Permission) to wire the Lambda function
+    /// as a target behind the specified ALB listener.
+    /// </summary>
+    /// <remarks>
+    /// The listener ARN (or template reference), path pattern, and priority are required.
+    /// See <a href="https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html">ALB Lambda documentation</a>.
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Method)]
+    public class ALBApiAttribute : Attribute
```
normj, line 22: My understanding besides host-header you should be able to setup a role for any arbitrary header like the following:
```
      {
        "Field": "http-header",
        "HttpHeaderConfig": {
          "HttpHeaderName": "X-Environment",
          "Values": ["dev"]
        }
      }
```

I don't see an property on the attribute for setting that up.

normj (changes requested):

Libraries/src/Amazon.Lambda.Annotations/ALB/ALBApiAttribute.cs
GarrettBeatty: http header https://github.com/aws/aws-lambda-dotnet/pull/2318/changes/d4d45db624a987f5a3421f9489a0660e3a88d2aa

query string and source ip https://github.com/aws/aws-lambda-dotnet/pull/2318/commits/b2970187878b01012fcf30763e16487486cb69f8

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Validation/LambdaFunctionValidator.cs
GarrettBeatty: updated to add support for fromX and other attributes. but one issue is that all of these attributes are in api gateway namespace. 

so in the users code they would need to import that (like this https://github.com/aws/aws-lambda-dotnet/pull/2318/changes/3c2c6966b0dba7a6f78c649fe3f86fe62fe9bbca) which seems weird. 

Should i create duplicate attributes or wondering if we can make breaking change to make it in some common namespace?

GarrettBeatty: i duplicated the attributes in https://github.com/aws/aws-lambda-dotnet/pull/2318/commits/311bcd61e1198dc22bb6bd6e5dee6f9733e224c3 as per slack thread

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/ALBInvokeCode.cs
```diff
@@ -0,0 +1,17 @@
+using Amazon.Lambda.Annotations.SourceGenerator.Models;
```
philasmar, line 4: missing the license header

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/ALBFromQueryAttributeBuilder.cs
```diff
@@ -0,0 +1,25 @@
+using Amazon.Lambda.Annotations.ALB;
```
philasmar, line 4: missing the license header

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/ALBFromHeaderAttributeBuilder.cs
```diff
@@ -0,0 +1,25 @@
+using Amazon.Lambda.Annotations.ALB;
```
philasmar, line 4: missing the license header

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/ALBApiAttributeBuilder.cs
```diff
@@ -0,0 +1,65 @@
+using Amazon.Lambda.Annotations.ALB;
```
philasmar, line 4: missing the license header

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/ALBInvoke.tt
```diff
@@ -0,0 +1,98 @@
+<#@ template language="C#" #>
```
philasmar, line 1: not sure if we add a license header  for the tt

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/ALBSetupParameters.tt
```diff
@@ -0,0 +1,303 @@
+<#@ template language="C#" #>
```
philasmar, line 1: same question about missing license header

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Templates/ALBSetupParametersCode.cs
```diff
@@ -0,0 +1,16 @@
+using Amazon.Lambda.Annotations.SourceGenerator.Models;
```
philasmar, line 4: missing the license header

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Validation/LambdaFunctionValidator.cs
```diff
@@ -1,4 +1,5 @@
-﻿using Amazon.Lambda.Annotations.SourceGenerator.Diagnostics;
+﻿using Amazon.Lambda.Annotations.ALB;
```
philasmar, line 1: missing the license header

Libraries/src/Amazon.Lambda.Annotations/ALB/ALBApiAttribute.cs
```diff
@@ -0,0 +1,185 @@
+using System;
```
philasmar, line 4: missing the license header

Libraries/src/Amazon.Lambda.Annotations/ALB/FromBodyAttribute.cs
```diff
@@ -0,0 +1,15 @@
+using System;
```
philasmar, line 4: missing the license header

Libraries/src/Amazon.Lambda.Annotations/README.md
```diff
@@ -852,6 +853,226 @@ The following SQS event source mapping will be generated for the `SQSMessageHand
...
+For each `ALBApi` attribute, the source generator creates three CloudFormation resources. Here is an example of the generated template for the basic hello endpoint:
+
+```json
+"ALBHello": {
+  "Type": "AWS::Serverless::Function",
+  "Metadata": {
+    "Tool": "Amazon.Lambda.Annotations"
+  },
+  "Properties": {
+    "Runtime": "dotnet6",
```
philasmar: nit: make this dotnet8

Libraries/src/Amazon.Lambda.Annotations/README.md
```diff
@@ -852,6 +853,226 @@ The following SQS event source mapping will be generated for the `SQSMessageHand
...
+    "CodeUri": ".",
+    "MemorySize": 256,
```
philasmar: nit: make this 512

normj (approved):
Approve assuming @philasmar comments are resolved.

GarrettBeatty:
fixed headers and readme @philasmar

philasmar (approved):

PR #2322, opened 2026-04-03
Add [SNSEvent] annotation attribute and source generator support

GarrettBeatty:
## Summary

Adds `[SNSEvent]` annotation attribute support to the Lambda Annotations framework, enabling developers to declaratively configure SNS topic-triggered Lambda functions directly in C# code. The source generator automatically produces the corresponding SAM/CloudFormation template configuration at build time.

## User Experience

With this change, developers can write SNS-triggered Lambda functions like this:

```csharp
[LambdaFunction]
[SNSEvent("@MyTopic")]
public async Task ProcessMessages(SNSEvent evnt)
{
    foreach (var record in evnt.Records)
    {
        // Handle SNS messages
    }
}
```

The source generator will automatically generate the SAM template entry:

```yaml
ProcessMessages:
  Type: AWS::Serverless::Function
  Properties:
    Events:
      MyTopic:
        Type: SNS
        Properties:
          Topic: !Ref MyTopic
```

### Attribute Properties

| Property | Required | Description | Default |
|----------|----------|-------------|---------|
| `Topic` | Yes | SNS topic reference (`@TopicName` for CloudFormation Ref, or topic ARN) | - |
| `ResourceName` | No | CloudFormation event resource name | Derived from topic name |
| `FilterPolicy` | No | JSON filter policy for the subscription | Not set |
| `Enabled` | No | Whether the event source is enabled | `true` |

The implementation is pretty similar to sqs event

Related: DOTNET-8573


Libraries/src/Amazon.Lambda.Annotations/SQS/SQSEventAttribute.cs
```diff
@@ -60,7 +60,7 @@ public string ResourceName
         /// </summary>
         public bool Enabled
         {
-            get => enabled.GetValueOrDefault();
+            get => enabled.GetValueOrDefault(true);
```
GarrettBeatty, line 63: i think this was a bug. default should be true as per the docs. right now it returns false by default

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/SNSEventAttributeBuilder.cs
```diff
@@ -0,0 +1,40 @@
+using Amazon.Lambda.Annotations.SNS;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.Annotations/SNS/SNSEventAttribute.cs
```diff
@@ -0,0 +1,113 @@
+using System;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.Annotations/SQS/SQSEventAttribute.cs
philasmar: Call that out in the changelog

normj (approved):

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/SNSEventAttributeBuilder.cs
GarrettBeatty: added everywhere

philasmar (approved):

PR #2324, opened 2026-04-03
Add [FunctionUrl] annotation attribute with CORS support and source generator

GarrettBeatty:
### What is this?

Adds first-class support for [Lambda Function URLs](https://docs.aws.amazon.com/lambda/latest/dg/lambda-urls.html) to the Lambda Annotations framework. Function URLs give your Lambda function a dedicated HTTPS endpoint without needing API Gateway — ideal for webhooks, simple APIs, and microservices.

### Usage

```csharp
[LambdaFunction]
[FunctionUrl(AuthType = FunctionUrlAuthType.NONE)]
public IHttpResult GetItems([FromQuery] string category, ILambdaContext context)
{
    return HttpResults.Ok(new { items = new[] { "item1", "item2" }, category });
}
```

This generates a handler that accepts HTTP API v2 payloads and emits a `FunctionUrlConfig` in your SAM template — no API Gateway resource needed.

### CORS support

```csharp
[FunctionUrl(AuthType = FunctionUrlAuthType.NONE,
             AllowOrigins = new[] { "https://example.com" },
             AllowMethods = new[] { "GET", "POST" },
             AllowHeaders = new[] { "Content-Type" },
             AllowCredentials = true,
             MaxAge = 3600)]
```

CORS configuration is optional — the `Cors` block is only emitted when at least one CORS property is set.

### Authentication

Two auth modes via `FunctionUrlAuthType`:
- **`NONE`** — Public endpoint, anyone with the URL can invoke the function
- **`AWS_IAM`** — Only authenticated IAM users/roles can invoke the function

### Key points

- Function URLs only support the HTTP API v2 payload format (`APIGatewayHttpApiV2ProxyRequest/Response`), so the source generator always uses v2 types — there is no version selection like `[HttpApi]` has
- All existing parameter binding (`[FromQuery]`, `[FromHeader]`, `[FromBody]`, `[FromRoute]`) and `IHttpResult` returns work out of the box
- Generates `FunctionUrlConfig` as a property on the function resource (not as an event source), matching the SAM specification
- Automatically cleans up `FunctionUrlConfig` from the template when the attribute is removed
- Handles switching between `[FunctionUrl]` and `[HttpApi]`/`[RestApi]` seamlessly
- Detects `[FunctionUrl]` without `[LambdaFunction]` and reports a diagnostic error

Related: DOTNET-8569


Libraries/src/Amazon.Lambda.Annotations/APIGateway/FunctionUrlAuthType.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        /// <summary>
+        /// No authentication. Anyone with the Function URL can invoke the function.
+        /// </summary>
+        NONE,
+
+        /// <summary>
+        /// IAM authentication. Only authenticated IAM users and roles can invoke the function.
+        /// </summary>
+        AWS_IAM
```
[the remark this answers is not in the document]
GarrettBeatty: i think it should be all caps

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Writers/CloudFormationWriter.cs
```diff
@@ -290,6 +301,47 @@ private string ProcessHttpApiAttribute(ILambdaFunctionSerializable lambdaFunctio
...
+
+                if (functionUrlAttribute.ExposeHeaders != null)
+                    _templateWriter.SetToken($"{corsPath}.ExposeHeaders", new List<string>(functionUrlAttribute.ExposeHeaders), TokenType.List);
+
+                if (functionUrlAttribute.AllowCredentials)
+                    _templateWriter.SetToken($"{corsPath}.AllowCredentials", true);
+
+                if (functionUrlAttribute.MaxAge > 0)
+                    _templateWriter.SetToken($"{corsPath}.MaxAge", functionUrlAttribute.MaxAge);
+            }
```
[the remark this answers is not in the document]
GarrettBeatty: fixed

Libraries/test/Amazon.Lambda.Annotations.SourceGenerators.Tests/WriterTests/FunctionUrlTests.cs
```diff
@@ -0,0 +1,215 @@
...
+            var report = GetAnnotationReport(new List<ILambdaFunctionSerializable> { lambdaFunctionModel });
+            ITemplateWriter templateWriter = templateFormat == CloudFormationTemplateFormat.Json ? new JsonWriter() : new YamlWriter();
+
+            cloudFormationWriter.ApplyReport(report);
+
+            templateWriter.Parse(mockFileManager.ReadAllText(ServerlessTemplateFilePath));
+            Assert.True(templateWriter.Exists("Resources.TestMethod.Properties.FunctionUrlConfig.AuthType"));
+            Assert.False(templateWriter.Exists("Resources.TestMethod.Properties.FunctionUrlConfig.Cors"));
+        }
+
```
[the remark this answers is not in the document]
GarrettBeatty: fixed

Libraries/src/Amazon.Lambda.Annotations/APIGateway/FunctionUrlAuthType.cs
```diff
@@ -0,0 +1,18 @@
+namespace Amazon.Lambda.Annotations.APIGateway
```
GarrettBeatty, line 4: technically function url is completely different than api gateway but uses the same api gateway v2 response type, so i kept it in the same namespace for now

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Models/Attributes/FunctionUrlAttributeBuilder.cs
```diff
@@ -0,0 +1,45 @@
+using System.Linq;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.Annotations/APIGateway/FunctionUrlAttribute.cs
```diff
@@ -0,0 +1,48 @@
+using System;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.Annotations/APIGateway/FunctionUrlAuthType.cs
philasmar, line 4: add license header

philasmar (approved):

philasmar (approved):

Libraries/src/Amazon.Lambda.Annotations/APIGateway/FunctionUrlAttribute.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        /// <summary>
+        /// The allowed origins for CORS requests. Example: new[] { "https://example.com" }
+        /// </summary>
+        public string[] AllowOrigins { get; set; }
+
+        /// <summary>
+        /// The allowed HTTP methods for CORS requests. Example: new[] { "GET", "POST" }
+        /// </summary>
+        public string[] AllowMethods { get; set; }
```
normj: I think we should be using an enum like we do in the API Gateway events. For the API Gateway we created our own enum to support "Any". Not sure if that is a valid situation here or is the absence of this property essentially "Any".

normj (changes requested):

Libraries/src/Amazon.Lambda.Annotations/APIGateway/FunctionUrlAttribute.cs
GarrettBeatty: https://github.com/aws/aws-lambda-dotnet/pull/2324/commits/69c37e369ffdcc32f50c840b23e19903944ec80b

GarrettBeatty: >". Not sure if that is a valid situation here or is the absence of this property essentially "Any".

i updated to use enums. but i dont think he absence of the property should be any (`*`). Since CORS is supposed to be more restrictive, i think not setting anything for it as the default makes the most sense. Making it Wildcard may not be intuitive for the user

normj (approved):

PR #2340, opened 2026-04-22
Update .NET Targets and Fix warnings

normj:
*Description of changes:*
The main driver was I wanted to drop out of support targets in `Amazon.Lambda.RuntimeSupport` which allows me to clean up the code and remove all for `#if` code branches. I decided since I was cleaning the targets I would also fix all of the compiler warnings. This spiraled into fixing up warnings in all of the other packages and remove out of support targets like .NET Core 3.1 and .NET 6 in them.

The PR contains a lot of files but most files just include deleted code for the removed targets.

Changes
* Defined msbuild parameter `DefaultPackageTargets` in the `common.props` used to configure the default .NET targets to build for. The value is `net8.0;net10.0`.
* Update all the event packages that had a mismatch of targets be used to `DefaultPackageTargets`
* Removed the `#if` that were no longer needed because the .NET targets were raised
* `Amazon.Lambda.Annotations` and `Amazon.Lambda.RuntimeSupport` do not use `DefaultPackageTargets` because adding new targets requires more work then just changing the value for this projects. I don't want anybody to think by changing `DefaultPackageTargets` that will automatically fix up those libraries
* In `Amazon.Lambda.RuntimeSupport` I was able to delete a lot of old code since we lowest supported target is now .NET 8.
* Addressed all compiler warnings and enabled treat warnings as errors in the `common.props`.
* The change in build targets caused a lot of instability with the unit tests particular with the GitHub action and the Annotation source generator. So had to do a fair amount of hardening up those tests.
* All the packages are getting a major version bump because of the build target changes.


buildtools/common.props
```diff
@@ -3,6 +3,9 @@
   <PropertyGroup>
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)/public.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
+    <NoWarn>$(NoWarn);CA1822</NoWarn>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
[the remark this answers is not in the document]
normj: Fixed

buildtools/common.props
```diff
@@ -1,28 +1,27 @@
...
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)/public.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
-
     <Authors>Amazon Web Services</Authors>
 	
+    <DefaultPackageTargets>net8.0;net10.0</DefaultPackageTargets>
+
     <OutputType>Library</OutputType>
+    <NoWarn>$(NoWarn);CA1822</NoWarn>	
     <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
GarrettBeatty: >Addressed all compiler warnings and enabled treat warnings as errors in the common.props.

but here i see its still false. should be true

normj: Ha, I deleted the wrong line. Fixed it now.

GarrettBeatty (approved):
just curious if the warnings come from because the upgrade or if they can be done separately

Libraries/src/Amazon.Lambda.Annotations/Amazon.Lambda.Annotations.csproj
```diff
@@ -2,7 +2,8 @@
 
   <PropertyGroup>
     <RootNamespace>Amazon.Lambda.Annotations</RootNamespace>
-    <TargetFrameworks>netstandard2.0;net6.0;net8.0</TargetFrameworks>
+    <!-- The .NET Standard 2.0 target is required for the source generator -->
+    <TargetFrameworks>netstandard2.0;net8.0;net10.0</TargetFrameworks>
```
philasmar, line 11: I don't believe the main app needs to have netstandard2.0. Only the SourceGenerator project. Correct me if I'm wrong.

Libraries/src/Amazon.Lambda.PowerShellHost/Amazon.Lambda.PowerShellHost.csproj
```diff
@@ -12,11 +12,13 @@
     <PackageTags>AWS;Amazon;Lambda;PowerShell</PackageTags>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
-    <PackageReference Include="Microsoft.PowerShell.SDK" Version="7.2.18" />
-  </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
-    <PackageReference Include="Microsoft.PowerShell.SDK" Version="7.4.1" />
+    <PackageReference Include="Microsoft.PowerShell.SDK" Version="7.4.14" />
+    <PackageReference Include="System.Security.Cryptography.Xml" Version="8.0.3" />
```
philasmar, line 17: Do you want to call this out in the changelog?

normj: Done

Libraries/src/Amazon.Lambda.Annotations/Amazon.Lambda.Annotations.csproj
normj: Updated the comment explaining how there is some code in the project used only at runtime that requiring newer .NET targets.

philasmar (approved):

Issue #2216, opened 2025-12-02
Support Lambda Durable Executions


jeastham1993:
### Describe the feature

Lambda now supports [durable executions](https://aws.amazon.com/blogs/aws/build-multi-step-applications-and-ai-workflows-with-aws-lambda-durable-functions/). Which allows stateful, durable workflows to be defined using application code. It has launched with Python/NodeJS support, but it would be great to add support for .NET as well.

### Use Case

Step Functions is a great product, but it's a proprietary language. Being able to define durable workflows in pure .NET code would remove the dependency on ASL, and bring .NET in line with other runtimes from a feature parity perspective.

### Proposed Solution

Happy to support implementation if details on the internals can be shared.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

N/A

### Targeted .NET Platform

N/A

### Operating System and version

N/A

jeastham1993:
@normj totally understand the prioritisation, figured you'd have other stuff going on. Would love to see this in .NET though 👍 if I can help in any way ping me.

madhub:
It appears that support for “Lambda durable functions” was added to the AWS SDK in version [4.0.144.0](https://github.com/aws/aws-sdk-net/blob/b983b3d304c530d1ea6e88a21eaec0d4bb8d1ad2/changelogs/SDK.CHANGELOG.2025.md#401440-2025-12-02-1643-utc). However, there is currently no documentation explaining how to use this feature.


<img width="1136" height="889" alt="Image" src="https://github.com/user-attachments/assets/ac65124a-e2ee-41e0-b799-86ce58e61946" />

<img width="1125" height="625" alt="Image" src="https://github.com/user-attachments/assets/56ee7c39-8e09-468b-b5ac-0af6a375bb05" />

GarrettBeatty:
just an update on this issue, i have started working on it!

irunjic:
@GarrettBeatty do you know if Lambda Durable Executions will also be supported when using Amazon.Lambda.AspNetCoreServer.Hosting?

GarrettBeatty:
@irunjic Currently there isn't a plan to support `Amazon.Lambda.AspNetCoreServer.Hosting`. Can you expand on your use case more?

irunjic:
[...]

PR #2360, opened 2026-05-12
Add durable execution Step + Wait end-to-end

GarrettBeatty:

## What

This is the first real end-to-end slice of the `Amazon.Lambda.DurableExecution` SDK. After this PR, you can write a workflow that calls `StepAsync` and `WaitAsync`, deploy it as a Lambda, and have it run against the actual durable-execution service with replay-aware checkpointing.

Public API introduced:

| Type | Purpose |
|---|---|
| `DurableFunction.WrapAsync` | Entry point; handles the durable-execution envelope (input hydration, output construction, status mapping). Four overloads: typed-input/typed-output and typed-input/void, each with and without an explicit `IAmazonLambda` client. |
| `IDurableContext` | User-facing context: `StepAsync` (typed-result and void overloads), `WaitAsync`, `LambdaContext`, `Logger`, `ExecutionContext`. |
| `StepConfig` | Per-step configuration. Empty in this PR — `RetryStrategy` and `Semantics` get wired in #2363. |
| `DurableExecutionException` | Base exception type for durable-execution failures. |

Step input/output and workflow envelope serialization both go through the `ILambdaSerializer` registered on `ILambdaContext.Serializer` (added in #2378)

## Why

Durable execution lets a Lambda function suspend and resume across invocations by checkpointing each side-effect to the service. This PR lays down the minimum needed to build everything that comes after — retries, callbacks, parallelism, the Annotations integration, the test runner package — all of it builds on the `Step + Wait` primitives and the replay machinery here.

I kept the scope narrow on purpose. Anything that does not block real-Lambda execution is pushed to follow-up PRs (see *Out of scope*) so this stays reviewable.

## How

The runtime runs a `Task.WhenAny` race in `DurableExecutionHandler` between the user's workflow task and a suspension signal. Every `StepAsync` / `WaitAsync` call goes to a per-operation class (`StepOperation` / `WaitOperation`) that checks `ExecutionState` (built from operations the service replayed) before deciding what to do:

- **Replay hit** — the operation has a `SUCCEEDED` / `FAILED` record from a prior invocation. Return the cached result (or rethrow) without re-running user code.
- **Cache miss** — run user code, enqueue a `SUCCEED`/`FAIL` checkpoint, return.
- **Wait pending** — the wait was scheduled previously but has not expired. Call `TerminationManager.SuspendAndAwait` to win the `WhenAny` race, returning `Pending` so the service re-invokes us when the timer fires.

Replay determinism comes from `OperationIdGenerator`, which produces stable IDs from the workflow's call sequence so the same step always lands on the same record across invocations.

**Serialization** is unified on `ILambdaContext.Serializer`. Both `WrapAsync` (workflow input/output envelope) and `StepOperation` (step result checkpointing) read the registered `ILambdaSerializer` off the per-invocation context and use it for all JSON conversion. AOT-safety is determined entirely by which serializer the caller registers with `LambdaBootstrapBuilder.Create` — register `SourceGeneratorLambdaJsonSerializer<TContext>` for AOT/trimmed deployments, or `DefaultLambdaJsonSerializer` for reflection-based scenarios. `StepConfig` is intentionally empty in this PR — it's the configuration carrier for `RetryStrategy` (#2363) and future per-step knobs.

**Checkpoint flushing** goes through `CheckpointBatcher`, a `Channel`-based queue with a single background worker. Each `EnqueueAsync` returns a `Task` that completes when the worker has flushed the containing batch to the service. `WrapAsyncCore` sets up the batcher, threads it into `DurableContext`, and `await`s `DrainAsync()` before returning to Lambda. Defaults match the Java SDK (`MaxBatchOperations = 200`, `MaxBatchBytes = 750 KB`, `FlushInterval = 0` — flush as soon as the queue drains). There's a `TODO` for the async-flush overload that Map/Parallel/Child Context will eventually need.

Key files:

- `DurableFunction.cs` — envelope + replay-state hydration + batcher lifecycle; pulls `ILambdaSerializer` from `ILambdaContext.Serializer`
- `DurableContext.cs` — facade; constructs the right per-operation class
- `Internal/DurableOperation.cs` — abstract base with `StartAsync` / `ReplayAsync` template methods
- `Internal/StepOperation.cs` / `Internal/WaitOperation.cs` — per-op replay logic; `StepOperation` serializes step results via the registered `ILambdaSerializer`
- `Internal/CheckpointBatcher.cs` / `Internal/CheckpointBatcherConfig.cs` — checkpoint queue + worker
- `Internal/ExecutionState.cs` — operation lookup + replay-mode flag
- `Internal/OperationIdGenerator.cs` — deterministic IDs
- `Internal/TerminationManager.cs` — suspension trigger
- `Internal/DurableExecutionHandler.cs` — the `WhenAny` race
- `Services/LambdaDurableServiceClient.cs` — service client wrapper

## Testing

Unit tests in `Amazon.Lambda.DurableExecution.Tests` cover: enums, exceptions, models, `OperationIdGenerator`, `TerminationManager`, `ExecutionState`, the handler race, both `Step` and `Wait` paths through `DurableContext`, `DurableFunction.WrapAsync`, and `CheckpointBatcher` (enqueue/flush, batching within window, overflow splitting, error propagation, drain, dispose, token updates, concurrency).

End-to-end integration tests in `Amazon.Lambda.DurableExecution.IntegrationTests` build each test workflow into a Docker container, deploy it as a real Lambda on `provided.al2023`, and run against the durable-execution service:

- `StepWaitStep` — basic Step → Wait → Step sequence
- `MultipleSteps` — several sequential steps
- `WaitOnly` — wait without any steps
- `LongerWait` — wait that spans multiple invocations
- `ReplayDeterminism` — verifies stable operation IDs across replays
- `StepFails` — verifies a failed step surfaces correctly

The AOT smoke test publishes the sample workflow with the source-generator serializer and produces zero `IL2026` / `IL3050` warnings.

- `IRetryStrategy`, `ExponentialRetryStrategy`, retry decision factories, `StepConfig.RetryStrategy` / `StepConfig.Semantics` *(#2363)*
- `StepException` and the per-step failure exception type
- `DurableLogger` replay-suppression (currently returns `NullLogger`)
- `Callbacks`, `InvokeAsync`, `ParallelAsync`, `MapAsync`, `RunInChildContextAsync`, `WaitForConditionAsync` — the interface intentionally does not declare these yet
- Annotations source-generator integration (including a `[DurableExecution]` marker attribute)


Libraries/src/Amazon.Lambda.DurableExecution/DurableFunction.cs
```diff
@@ -0,0 +1,262 @@
...
+    [RequiresUnreferencedCode("Uses reflection-based JSON for TInput/TOutput. Use the JsonSerializerContext overload for AOT.")]
+    [RequiresDynamicCode("Uses reflection-based JSON for TInput/TOutput. Use the JsonSerializerContext overload for AOT.")]
+    public static Task<DurableExecutionInvocationOutput> WrapAsync<TInput, TOutput>(
+        Func<TInput, IDurableContext, Task<TOutput>> workflow,
+        DurableExecutionInvocationInput invocationInput,
+        ILambdaContext lambdaContext)
+    {
+        _cachedLambdaClient ??= new AmazonLambdaClient();
+        return WrapAsyncCore(workflow, invocationInput, lambdaContext, _cachedLambdaClient, jsonContext: null);
+    }
```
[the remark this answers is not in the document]
GarrettBeatty: this is not possible i dont think

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/Amazon.Lambda.DurableExecution.IntegrationTests.csproj
```diff
@@ -0,0 +1,38 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
GarrettBeatty: reminder to self to update this and other csprojs

Libraries/src/Amazon.Lambda.DurableExecution/IDurableContext.cs
```diff
@@ -0,0 +1,110 @@
...
+/// The primary interface for durable execution operations.
+/// Passed to user workflow functions to access checkpointed steps and waits.
+/// Additional operations (callbacks, parallel, map, etc.) are added in
+/// follow-up PRs.
+/// </summary>
+public interface IDurableContext
+{
+    /// <summary>
+    /// Logger for the durable execution context.
+    /// TODO: Implement replay-safe de-duplication (currently returns NullLogger).
```
GarrettBeatty: TODO in later PR

Docs/durable-execution-design.md
```diff
@@ -1112,10 +1275,11 @@ public class StepConfig
     /// </summary>
     public StepSemantics Semantics { get; set; } = StepSemantics.AtLeastOncePerRetry;
 
-    /// <summary>
-    /// Custom serializer for the step result. Default is System.Text.Json.
-    /// </summary>
-    public ICheckpointSerializer? Serializer { get; set; }
```
GarrettBeatty, line 1118: this is a deviation from the previous design. originally i had the serializer as part of the step config but after playing around with the code, i realized this needs to be a parameter in the stepasync/waitasync functions so that native aot can have its own function calls

Docs/durable-execution-design.md
```diff
@@ -158,7 +158,7 @@ public class Function
     {
         // Step 1: Validate the order (checkpointed automatically)
         var validation = await context.StepAsync(
-            async () => await ValidateOrder(input.OrderId),
+            async (step) => await ValidateOrder(input.OrderId),
```
GarrettBeatty, line 161: in my design doc i previously had an api which the users function does not receive the step context, in case the user didnt need it) however, after looking at javas code, they deprecated this function and only kept the api which gives the user the step context.

Libraries/src/Amazon.Lambda.DurableExecution/Internal/TerminationManager.cs
```diff
@@ -0,0 +1,77 @@
...
+    public string? Message { get; init; }
+    public Exception? Exception { get; init; }
+}
+
+/// <summary>
+/// Manages the suspension signal for durable execution.
+/// Uses a TaskCompletionSource that resolves when the function should suspend.
+/// Only the first Terminate() call wins; subsequent calls are ignored.
+/// </summary>
+internal sealed class TerminationManager
```
GarrettBeatty, line 29: this is used in the instance where "we need to suspend the lambda function".  theres actually two ways to implement this functionality.

 Python and java  raises a special SuspendExecution exception from inside wait() that bubbles up the user's stack to a top-level except in the wrapper, which converts it to PENDING.
 
 .NET (and JS) don't throw. WaitAsync flips a one-shot signal on the TerminationManager and hands user code a Task that never completes, so the user's await parks forever. The wrapper is meanwhile Task.WhenAny-ing the user's workflow against that signal — the signal wins the race, the wrapper returns PENDING, and the abandoned user task gets GC'd.

GarrettBeatty: python and java are able to do this because they both throw some exception i.e BaseException and java.lang.Error which users do not usually catch (they usually just catch Exception type). in .net i think every exception comes from Exception and most users will catch a generic Exception so we cant do that

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/TestFunctions/LongerWaitFunction/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM public.ecr.aws/lambda/provided:al2023
```
GarrettBeatty, line 1: using docker files is a workaround until step function team allow lists .net managed runtime to call durable functions

Libraries/src/Amazon.Lambda.DurableExecution/DurableContext.cs
```diff
@@ -0,0 +1,147 @@
...
+        TimeSpan duration,
+        string? name = null,
+        CancellationToken cancellationToken = default)
+    {
+        // Service timer granularity is 1 second; sub-second waits would round to 0.
+        // WaitOptions.WaitSeconds is integer in [1, 31_622_400] (1 second to ~1 year).
+        if (duration < TimeSpan.FromSeconds(1))
+            throw new ArgumentOutOfRangeException(nameof(duration), duration, "Wait duration must be at least 1 second.");
+
+        if (duration > TimeSpan.FromSeconds(31_622_400))
```
GarrettBeatty, line 94: should we be validating this on our end?

Libraries/src/Amazon.Lambda.DurableExecution/Internal/CheckpointBatcher.cs
```diff
@@ -0,0 +1,213 @@
...
+/// <remarks>
+/// TODO: when Map / Parallel / ChildContext / WaitForCondition land — or when
+/// AtLeastOncePerRetry step START gets a non-blocking variant — they will need
+/// a fire-and-forget overload like
+/// <c>Task EnqueueAsync(SdkOperationUpdate update, bool sync)</c> where
+/// <c>sync=false</c> returns as soon as the item is queued. Java's
+/// <c>sendOperationUpdate</c> vs <c>sendOperationUpdateAsync</c> is the model.
+/// Today every call site is sync, so the API stays minimal.
+/// </remarks>
+internal sealed class CheckpointBatcher : IAsyncDisposable
```
GarrettBeatty, line 22: right now every step async "batch" is one item. but eventually when we implement Map/Parallel operations it will do concurrent operations and then things will be batched

Libraries/src/Amazon.Lambda.DurableExecution/DurableExecutionHandler.cs
```diff
@@ -0,0 +1,119 @@
...
+    internal static async Task<HandlerResult<TResult>> RunAsync<TResult>(
+        ExecutionState executionState,
+        TerminationManager terminationManager,
+        Func<Task<TResult>> userHandler)
+    {
+        // Run user code on a threadpool thread so it executes independently of
+        // the termination signal. When TerminationManager fires (e.g., WaitAsync),
+        // we need the WhenAny race below to resolve immediately without waiting
+        // for the user task to reach an await point.
+        var userTask = Task.Run(userHandler);
```
GarrettBeatty, line 73: the reason for this is imagine the user had

```
 async Task<TestResult> Workflow(TestEvent input, IDurableContext ctx)
  {
      // Imagine the user does CPU work or sync I/O before any await:
      Thread.Sleep(2000);                                // or a long compute loop
      await ctx.WaitAsync(TimeSpan.FromSeconds(5));      // first real await
      ...
  }
```


If we called userHandler() directly instead of Task.Run(userHandler):

var userTask = userHandler();                          // ← starts running RIGHT HERE, synchronously
var winner = await Task.WhenAny(userTask, terminationManager.TerminationTask);

The userHandler() invocation runs synchronously up to the first real await. If the user sleeps, blocks on sync I/O, or does any non-yielding work first, we don't even reach the await Task.WhenAny(...) line yet. The wrapper is stuck inside the user's call.

Libraries/src/Amazon.Lambda.DurableExecution/Internal/CheckpointBatcher.cs
GarrettBeatty: the fire and forget and START step will be implemented in https://github.com/aws/aws-lambda-dotnet/pull/2363. 

its technically not needed in this PR because durable functions really only care about seeing SUCCEEDED and FAILED steps. But once we add retries, it needs to know how many times (i.e the number of START steps) so its required then

Docs/durable-execution-design.md
```diff
@@ -906,21 +944,95 @@ Static helper for the non-Annotations handler path. Wraps a workflow function, h
...
+    [RequiresUnreferencedCode("Uses reflection-based JSON. Use the JsonSerializerContext overload for AOT.")]
+    [RequiresDynamicCode("Uses reflection-based JSON. Use the JsonSerializerContext overload for AOT.")]
     public static Task<DurableExecutionInvocationOutput> WrapAsync<TInput, TOutput>(
         Func<TInput, IDurableContext, Task<TOutput>> workflow,
         DurableExecutionInvocationInput invocationInput,
         ILambdaContext lambdaContext);
 
     /// <summary>
-    /// Wrap a workflow that takes typed input and returns no value.
+    /// Wrap a workflow (typed input + output) with explicit Lambda client.
```
GarrettBeatty, line 972: updated doc to be more clear about native aot apis

Libraries/src/Amazon.Lambda.DurableExecution/Internal/CheckpointBatcher.cs
```diff
@@ -0,0 +1,216 @@
...
+        _channel.Writer.TryComplete();
+        _shutdownCts.Cancel();
+        try { await _worker.ConfigureAwait(false); }
+        catch { /* swallow on dispose */ }
+        _shutdownCts.Dispose();
+    }
+
+    private async Task RunWorkerAsync(CancellationToken shutdownToken)
+    {
+        // TODO: also enforce _config.MaxBatchBytes here. Today we only cap by
```
GarrettBeatty, line 114: will be done in a follow up PR

Docs/durable-execution-design.md
```diff
@@ -244,9 +245,46 @@ public Task<DurableExecutionInvocationOutput> FunctionHandler(
...
+```csharp
+[JsonSerializable(typeof(OrderEvent))]
+[JsonSerializable(typeof(OrderResult))]
+internal partial class MyJsonContext : JsonSerializerContext { }
+
+public class Function
+{
+    public Task<DurableExecutionInvocationOutput> FunctionHandler(
+        DurableExecutionInvocationInput invocationInput, ILambdaContext context)
+        => DurableFunction.WrapAsync<OrderEvent, OrderResult>(
```
GarrettBeatty: needed to add another overload here for native AOT

Libraries/src/Amazon.Lambda.DurableExecution/Services/LambdaDurableServiceClient.cs
```diff
@@ -0,0 +1,108 @@
...
+            return checkpointToken;
+
+        var request = new CheckpointDurableExecutionRequest
+        {
+            DurableExecutionArn = durableExecutionArn,
+            CheckpointToken = checkpointToken ?? "",
+            Updates = pendingOperations is List<SdkOperationUpdate> list ? list : pendingOperations.ToList()
+        };
+
+        var response = await _lambdaClient.CheckpointDurableExecutionAsync(request, cancellationToken);
```
normj: Since the SDK call can fail for a variety of reasons like permissions. I wonder if we should wrap the SDK call in a try/catch and then wrap the SDK's exception in a `DurableExecutionException` and then you can give an error message with the context of what the SDK was being used for with the SDK exception message. I'm wanting users to not see an SDK stack trace and exception without any context of what it was being used for.

Libraries/src/Amazon.Lambda.DurableExecution/Services/LambdaDurableServiceClient.cs
```diff
@@ -0,0 +1,108 @@
...
+        CancellationToken cancellationToken = default)
+    {
+        var request = new GetDurableExecutionStateRequest
+        {
+            DurableExecutionArn = durableExecutionArn,
+            CheckpointToken = checkpointToken ?? "",
+            Marker = marker
+        };
+
+        var response = await _lambdaClient.GetDurableExecutionStateAsync(request, cancellationToken);
```
normj: Since the SDK call can fail for a variety of reasons like permissions. I wonder if we should wrap the SDK call in a try/catch and then wrap the SDK's exception in a `DurableExecutionException` and then you can give an error message with the context of what the SDK was being used for with the SDK exception message. I'm wanting users to not see an SDK stack trace and exception without any context of what it was being used for.

Libraries/src/Amazon.Lambda.DurableExecution/Internal/ExecutionState.cs
```diff
@@ -0,0 +1,134 @@
...
+
+/// <summary>
+/// In-memory store of the operations replayed from <see cref="InitialExecutionState"/>
+/// plus replay-mode tracking. Outbound checkpoints are owned by
+/// <see cref="CheckpointBatcher"/>; this type is the inbound side only.
+/// </summary>
+/// <remarks>
+/// Replay tracking mirrors the Python / Java / JavaScript reference SDKs:
+/// <list type="bullet">
+///   <item>At construction the workflow is "replaying" iff any user-replayable
```
normj: typo `iff`

Libraries/src/Amazon.Lambda.DurableExecution/Internal/OperationIdGenerator.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// </summary>
+    public string? ParentId { get; }
+
+    /// <summary>
+    /// Generates the next operation ID. The counter is pre-incremented so the
+    /// first ID is <c>hash("1")</c>, matching the reference SDKs.
+    /// </summary>
+    public string NextId()
+    {
+        var counter = ++_counter;
```
normj: Wasn't sure if threading for this is a concern when you get to the parallel steps. If so then maybe this line should be:
```
var counter = System.Threading.Interlocked.Increment(ref _counter);
```

Libraries/src/Amazon.Lambda.DurableExecution/Internal/OperationIdGenerator.cs
```diff
@@ -0,0 +1,101 @@
...
+        SHA256.HashData(bytes, hash);
+#else
+        using var sha = SHA256.Create();
+        var computed = sha.ComputeHash(bytes);
+        computed.CopyTo(hash);
+#endif
+        return ToHex(hash);
+    }
+
+    private static string ToHex(ReadOnlySpan<byte> bytes)
```
normj: Since you have the SDK you could potentially use the SDK's ToHex method `Amazon.Util.AWSSDKUtils.ToHex()`.

Libraries/src/Amazon.Lambda.DurableExecution/Internal/StepOperation.cs
```diff
@@ -0,0 +1,160 @@
...
+        try
+        {
+            var stepContext = new StepContext(OperationId, attemptNumber: 1, _logger);
+            var result = await _func(stepContext);
+
+            await EnqueueAsync(new SdkOperationUpdate
+            {
+                Id = OperationId,
+                Type = OperationTypes.Step,
+                Action = "SUCCEED",
```
normj: Any reason you can't use the enum `OperationAction.SUCCEED` instead of the string literal?

Libraries/src/Amazon.Lambda.DurableExecution/Internal/StepOperation.cs
```diff
@@ -0,0 +1,160 @@
...
+        catch (Exception ex)
+        {
+            // No retry logic in this commit: any thrown exception becomes a
+            // FAIL checkpoint and is re-thrown as a StepException. On replay,
+            // the FAILED branch above will re-throw without re-executing.
+            await EnqueueAsync(new SdkOperationUpdate
+            {
+                Id = OperationId,
+                Type = OperationTypes.Step,
+                Action = "FAIL",
```
normj: Any reason you can't use the enum `OperationAction.FAIL` instead of the string literal?

Libraries/src/Amazon.Lambda.DurableExecution/Internal/WaitOperation.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    protected override async Task<object?> StartAsync(CancellationToken cancellationToken)
+    {
+        // Sync-flush WAIT START before suspending — the service can't schedule
+        // a timer for a checkpoint it hasn't received.
+        await EnqueueAsync(new SdkOperationUpdate
+        {
+            Id = OperationId,
+            Type = OperationTypes.Wait,
+            Action = "START",
```
normj: Use enum

Libraries/src/Amazon.Lambda.DurableExecution/DurableExecutionInvocationInput.cs
```diff
@@ -0,0 +1,53 @@
+using System.Text.Json.Serialization;
```
normj, line 1: I would rather not do the pattern of putting classes in folders that don't match the namespace. Eventually somebody will add a file in the folder and the IDE will default to matching the folder.

For these files I would recommend just moving them to the parent folder instead of requiring everybody add a `using Amazon.Lambda.DurableExecution.Model` statement.

Libraries/src/Amazon.Lambda.DurableExecution/StepConfig.cs
```diff
@@ -0,0 +1,13 @@
+namespace Amazon.Lambda.DurableExecution;
```
normj, line 1: I would rather not do the pattern of putting classes in folders that don't match the namespace. Eventually somebody will add a file in the folder and the IDE will default to matching the folder.

For this file I'm indifferent if you want to move this to the parent folder or update the using statement. Especially since the file isn't really defined yet so not sure how common you expect users to use this.

Libraries/src/Amazon.Lambda.DurableExecution/DurableExecutionException.cs
```diff
@@ -0,0 +1,49 @@
+namespace Amazon.Lambda.DurableExecution;
```
normj, line 1: I would rather not do the pattern of putting classes in folders that don't match the namespace. Eventually somebody will add a file in the folder and the IDE will default to matching the folder.

For these files I would recommend just moving them to the parent folder.

normj (changes requested):
I think we need to revisit the JSON serializing and having a separate ICheckpointSerializer instead of reusing the ILambdaSerializer but I'm good with that being a separate PR.

Libraries/src/Amazon.Lambda.DurableExecution/Internal/StepOperation.cs
GarrettBeatty: fixed

Libraries/src/Amazon.Lambda.DurableExecution/Internal/WaitOperation.cs
GarrettBeatty: fixed

Libraries/src/Amazon.Lambda.DurableExecution/Internal/OperationIdGenerator.cs
GarrettBeatty: fixed

Libraries/src/Amazon.Lambda.DurableExecution/Internal/OperationIdGenerator.cs
GarrettBeatty: fixed

Libraries/src/Amazon.Lambda.DurableExecution/Internal/StepOperation.cs
GarrettBeatty: fixed

Libraries/src/Amazon.Lambda.DurableExecution/Internal/ExecutionState.cs
GarrettBeatty: should say if and only if

Libraries/src/Amazon.Lambda.DurableExecution/Services/LambdaDurableServiceClient.cs
GarrettBeatty: updated

Libraries/src/Amazon.Lambda.DurableExecution/Services/LambdaDurableServiceClient.cs
GarrettBeatty: updated

philasmar:
Well-architected first slice of the durable execution SDK. The separation of concerns (DurableFunction entry point → DurableExecutionHandler race → per-operation classes →
  CheckpointBatcher) is clean, replay semantics align with the Java/Python/JS reference SDKs, and test coverage is extensive. A few items worth addressing:

  ---

  ### Important

  **1. `TrackReplay` is O(n) per operation call, making replay O(n²)**

  `ExecutionState.TrackReplay` iterates all operations on every call to check if all terminal non-EXECUTION ops have been visited. For a workflow with hundreds of steps, this
  becomes noticeable.

  Consider precomputing the count at `LoadFromCheckpoint` time and decrementing on each visit:

  ```csharp
  private int _remainingReplayOps;

  // in LoadFromCheckpoint:
  _remainingReplayOps = _operations.Values
      .Count(op => op.Type != OperationTypes.Execution && IsTerminalStatus(op.Status));

  // in TrackReplay:
  if (_visitedOperations.Add(operationId)
      && _operations.TryGetValue(operationId, out var op)
      && op.Type != OperationTypes.Execution
      && IsTerminalStatus(op.Status))
  {
      if (--_remainingReplayOps <= 0)
          _isReplaying = false;
  }
 ```

  2. ExtractUserPayload silently returns default! on missing EXECUTION operation

  If the service envelope arrives without an EXECUTION-type operation (malformed input, wire issue), the workflow receives null/default input with no signal. This would surface
  as a confusing NullReferenceException deep in user code rather than a clear "durable execution envelope is malformed" error. Consider throwing DurableExecutionException when no
   EXECUTION op is found and TInput is a non-nullable reference type.

  3. No CancellationToken forwarding in WrapAsyncCore

  The state-pagination loop (while (!string.IsNullOrEmpty(nextMarker))) and DrainAsync don't observe any cancellation token. If the Lambda runtime signals shutdown mid-hydration,
   these operations could delay termination beyond the runtime's patience. Consider threading lambdaContext's remaining time through as a CancellationToken.

  4. UpperSnakeCaseEnumConverter<T> is public

  This converter lives in the Amazon.Lambda.DurableExecution namespace (not Internal) and is public sealed. It seems like an implementation detail that shouldn't be on the public
   API surface — locking in its behavior as a public contract limits future flexibility. Consider making it internal or moving it to the Internal namespace.

  ---
  Nits
  
  - DurableFunctionDeployment.ShortId() uses [..Math.Min(8, 32)] — Math.Min always evaluates to 8; could just be [..8].
  - CheckpointBatcherConfig.MaxBatchBytes is declared but unenforced (documented via TODO — confirming this is intentional for this PR scope).
  - The DurableExecutionHandler.RunAsync XML doc body references RetryScheduled / CallbackPending / InvokePending patterns that aren't wired yet. Minor doc-vs-shipped-code
  consistency point.

  ---
  Questions

  - IDurableContext.Logger always returns NullLogger.Instance — are there plans to wire DurableLogger before GA, or is this acceptable for the initial release?

GarrettBeatty:
>IDurableContext.Logger always returns NullLogger.Instance — are there plans to wire DurableLogger before GA, or is this acceptable for the initial release?

this will be in follow up PR. will address your other comments though

GarrettBeatty:
>    noo CancellationToken forwarding in WrapAsyncCore The state-pagination loop (while (!string.IsNullOrEmpty(nextMarker))) and DrainAsync don't observe any cancellation token. If the Lambda runtime signals shutdown mid-hydration,
these operations could delay termination beyond the runtime's patience. Consider threading lambdaContext's remaining time through as a CancellationToken.

i dont think this matters too much, im going to keep this as-is for now

GarrettBeatty:
@philasmar addressed your comments

Libraries/src/Amazon.Lambda.DurableExecution/DurableExecutionInvocationInput.cs
GarrettBeatty: fixed

Docs/durable-execution-design.md
```diff
@@ -244,9 +245,46 @@ public Task<DurableExecutionInvocationOutput> FunctionHandler(
...
 {
-    await context.StepAsync(async () => await SendNotification(input.UserId), name: "notify");
+    await context.StepAsync(async (step) => await SendNotification(input.UserId), name: "notify");
     await context.WaitAsync(TimeSpan.FromHours(1), name: "cooldown");
-    await context.StepAsync(async () => await Cleanup(input.UserId), name: "cleanup");
+    await context.StepAsync(async (step) => await Cleanup(input.UserId), name: "cleanup");
+}
+```
+
+For **NativeAOT** deployments, pass a `JsonSerializerContext` so the SDK can serialize/deserialize your input and output types without reflection:
```
normj: Can you update the design doc now that for Native AOT we are relying on the user to properly configure ILambdaSerializer?

Libraries/src/Amazon.Lambda.DurableExecution/IDurableContext.cs
```diff
@@ -0,0 +1,97 @@
...
+
+    /// <summary>
+    /// The underlying Lambda context.
+    /// </summary>
+    ILambdaContext LambdaContext { get; }
+
+    /// <summary>
+    /// Execute a step with automatic checkpointing. The step result is serialized
+    /// to a checkpoint using the <see cref="ILambdaSerializer"/> registered on
+    /// <see cref="ILambdaContext.Serializer"/> (typically configured via
```
normj: Don't say "(typically configured via <c>LambdaBootstrapBuilder.Create(handler, serializer)" because most users use the class library programming model which relies on the assembly attribute.

Libraries/src/Amazon.Lambda.DurableExecution/DurableContext.cs
```diff
@@ -0,0 +1,128 @@
...
+    private Task<T> RunStep<T>(
+        Func<IStepContext, Task<T>> func,
+        string? name,
+        StepConfig? config,
+        CancellationToken cancellationToken)
+    {
+        var serializer = LambdaContext.Serializer
+            ?? throw new InvalidOperationException(
+                "No ILambdaSerializer is registered on ILambdaContext.Serializer. " +
+                "Register a serializer via LambdaBootstrapBuilder.Create(handler, serializer) " +
```
normj: You need to rework the wording here because must user use the class library programming model which likes on the assembly attribute.

Libraries/src/Amazon.Lambda.DurableExecution/DurableFunction.cs
```diff
@@ -0,0 +1,244 @@
...
+    private static async Task<DurableExecutionInvocationOutput> WrapAsyncCore<TInput, TOutput>(
+        Func<TInput, IDurableContext, Task<TOutput>> workflow,
+        DurableExecutionInvocationInput invocationInput,
+        ILambdaContext lambdaContext,
+        IAmazonLambda lambdaClient)
+    {
+        var serializer = lambdaContext.Serializer
+            ?? throw new InvalidOperationException(
+                "No ILambdaSerializer is registered on ILambdaContext.Serializer. " +
+                "Register a serializer via LambdaBootstrapBuilder.Create(handler, serializer) " +
```
normj: You need to rework the wording here because must user use the class library programming model which likes on the assembly attribute.

Libraries/src/Amazon.Lambda.DurableExecution/DurableExecutionInvocationInput.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// <summary>
+    /// Previously checkpointed operation state for replay. Internal — consumed
+    /// only by <c>DurableFunction.WrapAsync</c> for replay correlation; user code
+    /// should never read or modify this. Marked <see cref="JsonIncludeAttribute"/>
+    /// so System.Text.Json populates it during deserialization despite being internal
+    /// (framework needs it, but it's not part of the public API contract).
+    /// </summary>
+    [JsonPropertyName("InitialExecutionState")]
+    [JsonInclude]
+    internal InitialExecutionState? InitialExecutionState { get; set; }
```
normj: I don't think internal properties will get set for serialization using a JsonSerializerContext when that JsonSerializerContext is defined in a different assembly then the type with internal properties. You should double check my thinking but we might need to handle this different to support source generator based serialization.

philasmar (commented):
Approved, assuming you address Norm's comments

philasmar (approved):

Libraries/src/Amazon.Lambda.DurableExecution/DurableExecutionInvocationInput.cs
GarrettBeatty: i just checked this and you are right

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/TestFunctions/MultipleStepsFunction/Function.cs
```diff
@@ -0,0 +1,46 @@
...
+
+public class Function
+{
+    private static readonly DurableEntryPoint<TestEvent, TestResult> _entry = new(Workflow);
+
+    public static async Task Main()
+    {
+        await LambdaBootstrapBuilder
+            .Create(_entry.InvokeAsync, new DefaultLambdaJsonSerializer())
+            .Build()
```
GarrettBeatty: another idea phil had was to have some `.WithDurableExecution` function instead of the user having to know to pass in this DurableEntryPoint wrapper class and function

Libraries/src/Amazon.Lambda.DurableExecution/IDurableContext.cs
GarrettBeatty: https://github.com/aws/aws-lambda-dotnet/pull/2360/changes/81380c969bd40a336a833fd708be9f5d686582de

Libraries/src/Amazon.Lambda.DurableExecution/DurableContext.cs
GarrettBeatty: https://github.com/aws/aws-lambda-dotnet/pull/2360/changes/81380c969bd40a336a833fd708be9f5d686582de

Libraries/src/Amazon.Lambda.DurableExecution/DurableFunction.cs
GarrettBeatty: https://github.com/aws/aws-lambda-dotnet/pull/2360/changes/81380c969bd40a336a833fd708be9f5d686582de

Libraries/src/Amazon.Lambda.DurableExecution/DurableExecutionInvocationInput.cs
GarrettBeatty: https://github.com/aws/aws-lambda-dotnet/pull/2360/commits/72f8824281f581dd7f518a4d254d1249194b9e73 made it public

Docs/durable-execution-design.md
GarrettBeatty: 850c901f953e7d98f3335482e9f8933d5c76f4f4

normj (approved):

PR #2150, opened 2025-10-16
Reset Stream Position back to 0

KurogamiLight3303:
Features from ASP Net Core like Output cache aren't working probably because of this

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1702, opened 2024-03-19
IOutputCache does not work with Lambda runtime


tvytrykush:
### Describe the bug

When using IOutputCache feature of asp.net core the cached value is not populated from the response body and is empty

### Expected Behavior

Response's cache value is properly cached from response body alongside the key(that is cached properly)

### Current Behavior

Because empty value is being cached with valid key - on subsequent requests using same path the response body is going to be empty.

### Reproduction Steps

1. Create new project using "Lambda ASP.NET Core Web API" template
2. Get redis instance running
3. Add Output cache to service collection in Startup:
```C#
services
            .AddOutputCache(options =>
                options.AddBasePolicy(builder =>
                        builder.AddPolicy<OutputCacheRedisPolicy>().Cache(),
                    true))
            .AddStackExchangeRedisOutputCache(opts =>
            {
                opts.Configuration = "localhost:<redis-port>";
                opts.InstanceName = "lambda";
            });
```
4. Add `app.UseOutputCache();` before app.UseEndpoints in Configure
5. Add IOutputCachePolicy implementation class into project:
```C#
public class OutputCacheRedisPolicy : IOutputCachePolicy
{
    private const string TagPrefix = "response-cache-";
    private static readonly TimeSpan CacheExpirationPeriod = TimeSpan.FromMinutes(15);

    ValueTask IOutputCachePolicy.CacheRequestAsync(OutputCacheContext context, CancellationToken cancellationToken)
    {
        var attemptOutputCaching = AttemptOutputCaching(context);
        
        context.EnableOutputCaching = true;
        context.AllowCacheLookup = attemptOutputCaching;
        context.AllowCacheStorage = attemptOutputCaching;
        context.AllowLocking = true;
        context.ResponseExpirationTimeSpan = CacheExpirationPeriod;
        context.CacheVaryByRules.QueryKeys = "*";

        context.Tags.Add(TagPrefix + context.HttpContext.Request.RouteValues["controller"]);
        
        return ValueTask.CompletedTask;
    }
    
    private static bool AttemptOutputCaching(OutputCacheContext context)
    {
        var request = context.HttpContext.Request;

        return HttpMethods.IsGet(request.Method) || HttpMethods.IsHead(request.Method);
    }
    
    public ValueTask ServeFromCacheAsync(OutputCacheContext context, CancellationToken cancellation) 
        => ValueTask.CompletedTask;

[...]

PR #2364, opened 2026-05-12
Fix IOutputCache empty-body bug (#1702): bump feature revision in InvokeFeatures.Set

GarrettBeatty:
## Summary

- Fixes [#1702](https://github.com/aws/aws-lambda-dotnet/issues/1702): `IOutputCache` (and any other middleware that wraps the response body) caches/sees zero bytes when an ASP.NET Core app is hosted in Lambda.
- One-line behavioral change in `Internal/InvokeFeatures.cs`: `Set<TFeature>` now delegates to the indexer so `_containerRevision` is bumped on every mutation.

## The bug

When middleware does `HttpContext.Response.Body = wrapperStream`, ASP.NET Core's `DefaultHttpResponse.Body.set` replaces the active `IHttpResponseBodyFeature` via `Features.Set<IHttpResponseBodyFeature>(new StreamResponseBodyFeature(value, oldFeature))` ([source](https://github.com/dotnet/aspnetcore/blob/main/src/Http/Http/src/Internal/DefaultHttpResponse.cs)).

ASP.NET Core caches feature lookups via `FeatureReferences<T>.Fetch`, which only re-reads the dictionary when `Collection.Revision` changes ([source](https://github.com/dotnet/aspnetcore/blob/main/src/Extensions/Features/src/FeatureReferences.cs)).

Our `InvokeFeatures.Set<TFeature>` was bypassing the indexer:

```csharp
public void Set<TFeature>(TFeature instance)
{
    if (instance == null)
        return;

    _features[typeof(TFeature)] = instance;   // <-- never bumps _containerRevision
}
```

So after `Response.Body = OutputCacheStream`:
1. The dictionary entry for `IHttpResponseBodyFeature` is updated to the new `StreamResponseBodyFeature` wrapping `OutputCacheStream`.
2. `_containerRevision` is **not** bumped.
3. ASP.NET Core's cached `IHttpResponseBodyFeature` reference (and the `PipeWriter` it produced) still point at the original feature, which writes directly to the underlying `MemoryStream`.
4. `WriteAsJsonAsync` (and friends) → `response.BodyWriter` → cached writer → bytes flow into the `MemoryStream` and never visit `OutputCacheStream`.
5. `OutputCacheMiddleware.FinalizeCacheBodyAsync` reads `context.OutputCacheStream.GetCachedResponseBody()` → empty → caches an empty entry.
6. Every subsequent request is served from the cache as an empty body.

## The fix

```csharp
public void Set<TFeature>(TFeature instance)
{
    this[typeof(TFeature)] = instance;   // indexer already does _containerRevision++
}
```

The indexer setter (just above in the same file) was already correct — it increments `_containerRevision` on every change and supports null-removal per the `IFeatureCollection` contract. Delegating to it makes both code paths behave identically. No other change is needed; the bug is invisible until middleware swaps a feature mid-request.

This change also helps any other middleware that wraps the body via `Response.Body =` / `Response.BodyWriter =` (e.g. `ResponseCompression`).

## How we tested

End-to-end on real AWS Lambda + API Gateway HTTP API v2:

1. **Reproduced the bug on `dev` (unfixed):** deployed a minimal app — `services.AddOutputCache()` + `app.UseOutputCache()` + `MapGet(\"/api/values\", () => new[]{\"value1\",\"value2\"})`. First request returned `Content-Length: 19`, body `[\"value1\",\"value2\"]`. Every subsequent request returned `Content-Length: 0`, empty body, with `age` header confirming it was served from the cache.

2. **Verified PR [#2150](https://github.com/aws/aws-lambda-dotnet/pull/2150) does not fix it.** That PR adds `responseFeatures.Body.Position = 0L` after marshalling. Deployed it; bug still repros. A diag endpoint confirmed the underlying `MemoryStream` had `len=19, pos=0` after the request — the position reset worked, but it was on the wrong stream. The OutputCache buffer was always empty because writes never reached `OutputCacheStream`.

3. **Verified the real fix works.** Built a NuGet from this branch, redeployed the same sample. Result:

```
Request 1 (cold):  200, Content-Length: 19, body: [\"value1\",\"value2\"]
Request 2:         200, Content-Length: 19, body: [\"value1\",\"value2\"], age: 0
Request 3:         200, Content-Length: 19, body: [\"value1\",\"value2\"], age: 0
Request 4:         200, Content-Length: 19, body: [\"value1\",\"value2\"], age: 1
```

All four requests return the full body. Requests 2-4 carry the `age` header, proving they're served from the OutputCache (and that the cache contains real bytes, not zero-length).

## Test plan

- [x] Reproduce the bug end-to-end against current `dev`
- [x] Verify PR #2150's proposed fix does not solve the bug
- [x] Verify this PR's fix solves the bug end-to-end against the same repro
- [x] CI: existing `Amazon.Lambda.AspNetCoreServer.Test` suite passes

## Notes

- PR #2150 should be closed in favor of this; its `Position = 0L` line addresses a symptom on the wrong stream and has no effect on the OutputCache scenario.
- Single-line behavioral change; same fix applies to all three hosting modes (`APIGatewayProxyFunction`, `APIGatewayHttpApiV2ProxyFunction`, `ApplicationLoadBalancerFunction`) because they all share `InvokeFeatures`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.autover/changes/a0faddb0-c508-415c-b436-d99d68eafc56.json
+++ b/.autover/changes/a0faddb0-c508-415c-b436-d99d68eafc56.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.AspNetCoreServer",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Fix InvokeFeatures.Set<TFeature> to bump the feature collection revision so middleware that wraps the response body (e.g. OutputCache, ResponseCompression) is properly visible to ASP.NET Core's FeatureReferences cache. Resolves https://github.com/aws/aws-lambda-dotnet/issues/1702 where IOutputCache stored empty response bodies."
+      ]
+    }
+  ]
+}
--- a/Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/InvokeFeatures.cs
+++ b/Libraries/src/Amazon.Lambda.AspNetCoreServer/Internal/InvokeFeatures.cs
@@ -117,10 +117,11 @@ namespace Amazon.Lambda.AspNetCoreServer.Internal
 
         public void Set<TFeature>(TFeature instance)
         {
-            if (instance == null)
-                return;
-
-            _features[typeof(TFeature)] = instance;
+            // Delegate to the indexer so _containerRevision is bumped, otherwise
+            // ASP.NET Core's FeatureReferences cache will return stale feature
+            // references after middleware (e.g. OutputCache, ResponseCompression)
+            // wraps the response body via HttpContext.Response.Body = wrapper.
+            this[typeof(TFeature)] = instance;
         }
 
         IEnumerator IEnumerable.GetEnumerator()
--- a/Libraries/test/Amazon.Lambda.AspNetCoreServer.Test/UtilitiesTest.cs
+++ b/Libraries/test/Amazon.Lambda.AspNetCoreServer.Test/UtilitiesTest.cs
@@ -29,5 +29,57 @@ namespace Amazon.Lambda.AspNetCoreServer.Test
             var feature = new InvokeFeatures() as IHttpResponseFeature;
             Assert.Equal(200, feature.StatusCode);
         }
+
+        // Regression test for https://github.com/aws/aws-lambda-dotnet/issues/1702.
+        // ASP.NET Core's FeatureReferences cache uses Revision to detect when a
+        // feature has been swapped (e.g. OutputCache/ResponseCompression replacing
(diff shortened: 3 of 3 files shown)


GarrettBeatty:
@normj tbh i dont fully understand this fix or this area of the code, but from testing at least it seems to fix the issue. I am not really sure on the full implications of this change

GarrettBeatty:
I can add a unit test if the change looks good

normj (approved):
Interesting catch that we were missing the revision getting incremented.

philasmar (approved):

PR #2368, opened 2026-05-13
ci: add contents: read to change-file-in-pr workflow

arpitjain099:
Single-workflow PR. `change-file-in-pr.yml` checks every PR for a release-note change file under `.autover/changes/` using `tj-actions/changed-files`. Pure read; no comments, pushes, or release writes. `contents: read` is the minimum.


GarrettBeatty:
we havent seen any issues with the current permissions so not sure if this is needed?

PR #2370, opened 2026-05-14
Add RunInChildContextAsync

GarrettBeatty:

## What

Adds child-context support to `Amazon.Lambda.DurableExecution`. A user function can now run inside a logical sub-workflow with its own deterministic operation-ID space; the result is checkpointed as a `CONTEXT` operation so subsequent invocations replay the cached value without re-executing the func.

Public API:

| Type | Purpose |
|---|---|
| `IDurableContext.RunInChildContextAsync<T>(...)` | Run a user func inside a child context, return `T`. |
| `IDurableContext.RunInChildContextAsync(...)` | Void overload. |
| `ChildContextConfig` | Optional config: `SubType` (observability label) + `ErrorMapping` (exception remapping). |
| `ChildContextException` | Surfaced on child-context failure; carries `SubType`, `ErrorType`, `ErrorData`, `OriginalStackTrace`. |

## How

`Internal/ChildContextOperation<T>` mirrors the Step/Wait pattern from #2360:

- **Fresh execution.** Synchronously flushes a `CONTEXT START` checkpoint, then invokes the user func with a child `IDurableContext`. The child has its own operation-ID counter, so steps/waits inside it get deterministic IDs scoped to the child. On success the SDK writes `SUCCEED` with the serialized result; on throw it writes `FAIL` with the error details and throws `ChildContextException`.
- **Replay.** `SUCCEEDED` returns the cached value (deserialized via the registered `ILambdaSerializer`). `FAILED` reconstructs and throws `ChildContextException` (after `ErrorMapping`, if provided). `STARTED` / `PENDING` re-runs the func and lets the child's own operations replay from their own checkpoints — a child whose inner step suspended will re-suspend the same way.
- **Result serialization** uses `ILambdaContext.Serializer` (consistent with `StepOperation`);
- **Error mapping** is applied both on first failure and on replay of a `FAILED` checkpoint, so callers see the same exception type either way.
- `LambdaDurableServiceClient.MapFromSdkOperation` is extended to copy `ContextDetails` (`Result` + `Error`) for the new `CONTEXT` operation kind.

This is a building block for upcoming `WaitForCallbackAsync`, which (per the Java/JS reference SDKs) wraps `CreateCallbackAsync` + a submitter step inside a child context for observability and clean error mapping.

## Testing

14 new unit tests in `ChildContextOperationTests.cs`:

- Fresh execution + checkpoint emission (`CONTEXT START` -> user func -> `SUCCEED`).
- Deterministic child operation IDs (child's inner steps get IDs scoped to the child).
- Replay `SUCCEEDED` returns cached value without re-running.
- Replay `FAILED` throws `ChildContextException`, with and without `ErrorMapping`.
- Suspended child: `Wait` inside child propagates termination correctly.
- Replay `STARTED` with completed inner step.
- Void overload.
- Type-mismatch non-determinism detection.
- `SubType` propagation to the wire.

161/161 tests pass on net8.0 and net10.0. Production build clean: 0 warnings, 0 errors.

- `WaitForCallbackAsync` / `CreateCallbackAsync` / `InvokeAsync`
- `MapAsync` / `ParallelAsync` / `WaitForConditionAsync`

### DIFF
--- a/Libraries/src/Amazon.Lambda.DurableExecution/ChildContextConfig.cs
+++ b/Libraries/src/Amazon.Lambda.DurableExecution/ChildContextConfig.cs
@@ -1,0 +1,32 @@
+namespace Amazon.Lambda.DurableExecution;
+
+/// <summary>
+/// Configuration for a child context.
+/// </summary>
+/// <remarks>
+/// A child context is a logical sub-workflow with its own deterministic
+/// operation-ID space, persisted as a <c>CONTEXT</c> operation. Use
+/// <see cref="IDurableContext.RunInChildContextAsync{T}(System.Func{IDurableContext, System.Threading.Tasks.Task{T}}, string?, ChildContextConfig?, System.Threading.CancellationToken)"/>
+/// (and overloads) to run code inside one.
+/// </remarks>
+public sealed class ChildContextConfig
+{
+    /// <summary>
+    /// Operation sub-type label for observability (e.g. <c>"WaitForCallback"</c>).
+    /// Surfaces on the wire <c>OperationUpdate.SubType</c> field.
+    /// </summary>
+    public string? SubType { get; set; }
+
+    /// <summary>
+    /// Optional function to transform exceptions thrown by the child context's
+    /// user function before they surface to the caller. Useful for wrapping
+    /// low-level errors into domain-specific exceptions.
+    /// </summary>
+    /// <remarks>
+    /// Applied when the user function throws (the mapped exception propagates
+    /// to the caller of <c>RunInChildContextAsync</c>) and on replay of a
+    /// <c>FAILED</c> child context (the constructed
+    /// <see cref="ChildContextException"/> is mapped before being thrown).
+    /// </remarks>
+    public Func<Exception, Exception>? ErrorMapping { get; set; }
+}
--- a/Libraries/src/Amazon.Lambda.DurableExecution/Services/LambdaDurableServiceClient.cs
+++ b/Libraries/src/Amazon.Lambda.DurableExecution/Services/LambdaDurableServiceClient.cs
@@ -7,6 +7,7 @@ using Operation = Amazon.Lambda.DurableExecution.Operation;
 using StepDetails = Amazon.Lambda.DurableExecution.StepDetails;
 using WaitDetails = Amazon.Lambda.DurableExecution.WaitDetails;
 using ExecutionDetails = Amazon.Lambda.DurableExecution.ExecutionDetails;
+using ContextDetails = Amazon.Lambda.DurableExecution.ContextDetails;
 
 namespace Amazon.Lambda.DurableExecution.Services;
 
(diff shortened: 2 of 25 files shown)


Libraries/src/Amazon.Lambda.DurableExecution/Services/LambdaDurableServiceClient.cs
```diff
@@ -116,7 +117,9 @@ private static Operation MapFromSdkOperation(SdkOperation sdkOp)
                 Error = sdkOp.StepDetails.Error != null ? new ErrorObject
                 {
                     ErrorType = sdkOp.StepDetails.Error.ErrorType,
-                    ErrorMessage = sdkOp.StepDetails.Error.ErrorMessage
+                    ErrorMessage = sdkOp.StepDetails.Error.ErrorMessage,
```
GarrettBeatty, line 120: not related to child context, but missed copying the error details before here

Libraries/src/Amazon.Lambda.DurableExecution/ChildContextConfig.cs
```diff
@@ -0,0 +1,32 @@
+namespace Amazon.Lambda.DurableExecution;
+
+/// <summary>
+/// Configuration for a child context.
+/// </summary>
+/// <remarks>
+/// A child context is a logical sub-workflow with its own deterministic
```
GarrettBeatty, line 7:
```csharp
ctx.StepAsync(..., "validate");                 // root  #1
ctx.RunInChildContextAsync(child => {           // root  #2  (the CONTEXT op)
    child.StepAsync(..., "charge");             // child #1
    child.StepAsync(..., "ship");               // child #2
}, name: "phase");
ctx.StepAsync(..., "notify");                   // root  #3

```
<p>Resulting IDs:</p>

Op | Id | ParentId
-- | -- | --
validate | h("1") | null
phase | h("2") | null
charge | h("h(\"2\")-1") | h("2")
ship | h("h(\"2\")-2") | h("2")
notify | h("3") | null


</body>
</html>

philasmar (changes requested):
## Important Issues - Should Fix

  1. **Framework exceptions (NonDeterministicExecutionException) from inner operations get wrapped and checkpointed as CONTEXT FAIL**
     - The catch-all catch (Exception ex) in ExecuteFunc will catch NonDeterministicExecutionException thrown by inner operations (e.g., a step inside the child has a type/name
  mismatch on replay). This exception indicates the entire execution is corrupt and should abort, but currently it gets wrapped in ChildContextException and checkpointed as a
  CONTEXT FAIL. On subsequent invocations, the service would replay the FAIL status instead of re-detecting the non-determinism. Consider re-throwing
  NonDeterministicExecutionException (and possibly StepInterruptedException) before the generic catch.
     - Suggestion: Add a catch clause before the generic one: catch (NonDeterministicExecutionException) { throw; }

  2. **Fresh-execution ChildContextException has fewer fields populated than replay ChildContextException**
     - File: `Libraries/src/Amazon.Lambda.DurableExecution/Internal/ChildContextOperation.cs`
     - On first failure, the thrown ChildContextException only sets SubType and ErrorType. But on replay of a FAILED checkpoint (BuildChildContextException), ErrorData and
  OriginalStackTrace are also populated from the stored error. This means ErrorMapping sees different exception shapes depending on fresh vs replay. Since you already capture the
   stack trace in ToSdkError(ex), you could populate OriginalStackTrace on the fresh-path exception too.
     - Suggestion: Add OriginalStackTrace = ex.StackTrace?.Split(...) to the fresh-path ChildContextException construction

  ## Minor Issues
  
  1. Test comment references NullCheckpointSerializer which does not exist in `Libraries/test/Amazon.Lambda.DurableExecution.Tests/ChildContextOperationTests.cs`
  2. Dockerfiles run as root (Semgrep flagged) in `Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/TestFunctions/ChildContextFunction/Dockerfile`

GarrettBeatty:
e34c6c846b0c47cf4ba4574be66506bb5359fb3e addressed comments @philasmar

philasmar (approved):

normj (approved):

PR #2371, opened 2026-05-14
Add replay-aware logger to Amazon.Lambda.DurableExecution

GarrettBeatty:

--- --- ---

## What

Implements `context.Logger`, the replay-aware `ILogger` described in `Docs/durable-execution-design.md` and shipped by the Python / Java / JavaScript reference SDKs.

Public API surface introduced:

| Type | Purpose |
|---|---|
| `IDurableContext.Logger` | Replay-safe `ILogger` (was `NullLogger.Instance`). |
| `IDurableContext.ConfigureLogger(LoggerConfig)` | Swap the inner logger and/or disable replay-aware filtering. |
| `LoggerConfig` | `CustomLogger` + `ModeAware` configuration record. |

## Why

Without replay-aware logging, every `Console.WriteLine` (or any non-suppressing logger) repeats on every replay invocation. A 30-step workflow re-invoked 30 times produces 30 copies of every log line — noisy at best, misleading at worst. The reference SDKs all solve this by reading replay state on each log call and suppressing emission while the workflow is re-deriving prior operations from checkpoint state. This PR ports that behavior to .NET on top of the per-operation replay tracker introduced in #2360.

## How

**ReplayAwareLogger.** An `ILogger` decorator that consults `ExecutionState.IsReplaying` on every call. Short-circuits both `Log<TState>` and `IsEnabled` during replay so `LoggerExtensions.LogXxx` doesn't even format the message string. `BeginScope` always passes through so the scope stack stays balanced — suppression only applies at log emission.

**Default inner logger.** `LambdaCoreLogger` — a minimal in-package adapter that delegates to `Amazon.Lambda.Core.LambdaLogger.Log`, so logs flow into the standard Lambda runtime pipeline (JSON when `AWS_LAMBDA_LOG_FORMAT=JSON`, level-filtered by `AWS_LAMBDA_LOG_LEVEL`). Two structured-logging behaviors:

- When `state` is the `FormattedLogValues` produced by `LoggerExtensions.LogXxx`, the original template and named-argument values are forwarded so the runtime's JSON formatter surfaces `{OrderId}`-style placeholders as top-level structured attributes.
- `BeginScope` maintains an `AsyncLocal` chain of scope state. KVP-shaped scope state is appended to the outgoing template as named placeholders (inner→outer order, inner wins on key collision; explicit message args win over scope keys), so `durableExecutionArn` / `operationId` / etc. show up as top-level JSON fields without callers having to swap in a third-party logger.

Mirrors the structured-logging pattern in `Amazon.Lambda.Logging.AspNetCore.LambdaILogger`. Avoids forcing a dependency on `Amazon.Lambda.Logging.AspNetCore`. Users who want Serilog/Powertools/etc. swap their own logger via `ConfigureLogger`.

**Metadata scopes.** `DurableFunction.WrapAsyncCore` opens a `BeginScope` around the workflow body carrying `durableExecutionArn` + `awsRequestId`. `StepOperation` opens a per-step scope (`operationId`, `operationName`, `attempt`) around the user-func invocation only. Combined with the scope-aware default logger above, every log line emitted by user code is automatically tagged with execution / step metadata.

Key files:
- `LoggerConfig.cs` — public configuration type
- `Internal/ReplayAwareLogger.cs` — the replay-aware decorator
- `Internal/LambdaCoreLogger.cs` — default inner logger; preserves structured args + flattens scope chain
- `DurableContext.cs` — replaces `NullLogger` default; implements `ConfigureLogger`
- `DurableFunction.cs` — execution-level scope
- `Internal/StepOperation.cs` — step-level scope around user func

## Testing

**Unit tests** (21 new in `Amazon.Lambda.DurableExecution.Tests`):

- `ReplayAwareLoggerTests` (7) — replay suppression, execution passthrough, `ModeAware=false`, `IsEnabled` short-circuit during replay, `IsEnabled` delegation during execution, `BeginScope` always-delegates, mid-workflow REPLAY → NEW transition (mirrors Python's `test_logger_replay_then_new_logging`).
- `DurableContextTests` (3 new + 1 updated) — `ConfigureLogger_NullArg_Throws`, `ConfigureLogger_WithCustomLogger_ReachesUserLogger`, `ConfigureLogger_ModeAwareFalse_LogsDuringReplay`. The pre-existing `Logger_Defaults_ToNullLogger` is updated to `Logger_Default_IsReplayAwareLogger` to assert the new default.
- `LambdaCoreLoggerTests` (11) — installs capture delegates into `LambdaLogger._loggingWithLevelAction` (the same hook RuntimeSupport uses) and asserts: named placeholders + arg values are forwarded intact, the exception variant works, plain messages pass through as literals, non-`FormattedLogValues` state falls back to `formatter(state, exception)`, `IsEnabled(None)` returns false, KVP scopes are appended, nested scopes flatten inner→outer with inner winning on collision, explicit message args win over scope keys, scope is popped on dispose, `AsyncLocal` isolates concurrent tasks, and non-KVP scopes are ignored.

**Integration test** (`ReplayAwareLoggerTest` in `Amazon.Lambda.DurableExecution.IntegrationTests`):

End-to-end proof on real AWS infra. Deploys a `step → wait(3s) → step` workflow that pairs each `context.Logger.LogInformation` line with a `Console.WriteLine` ''control'' line; the test function runs with `AWS_LAMBDA_LOG_FORMAT=JSON`. After the durable execution completes (across two invocations driven by the wait), queries CloudWatch Logs and asserts:

- Each replay-aware line appears **exactly once** across both invocations.
- Each control line appears **once per invocation** that reached it (proving the function genuinely replayed).
- Parsed JSON log records carry the expected scope-derived top-level fields: `durableExecutionArn` + `awsRequestId` on workflow-level lines; additionally `operationId` + `operationName` + `attempt` on lines emitted inside a step delegate.

This pins both the replay-suppression contract and the structured-scope contract end-to-end against the actual durable-execution service.

- `MapAsync` / `ParallelAsync` / `RunInChildContextAsync` / `WaitForConditionAsync`
- `CallbackAsync`, `InvokeAsync`

### DIFF
--- a/Libraries/src/Amazon.Lambda.DurableExecution/DurableContext.cs
+++ b/Libraries/src/Amazon.Lambda.DurableExecution/DurableContext.cs
@@ -1,7 +1,6 @@
 using Amazon.Lambda.Core;
 using Amazon.Lambda.DurableExecution.Internal;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Logging.Abstractions;
 
 namespace Amazon.Lambda.DurableExecution;
 
@@ -17,6 +16,7 @@ internal sealed class DurableContext : IDurableContext
     private readonly OperationIdGenerator _idGenerator;
     private readonly string _durableExecutionArn;
     private readonly CheckpointBatcher? _batcher;
+    private ILogger _logger;
 
     public DurableContext(
         ExecutionState state,
@@ -32,12 +32,24 @@ internal sealed class DurableContext : IDurableContext
         _durableExecutionArn = durableExecutionArn;
         _batcher = batcher;
         LambdaContext = lambdaContext;
+        _logger = new ReplayAwareLogger(new LambdaCoreLogger(), state, modeAware: true);
     }
 
-    // Replay-safe logger ships in a follow-up PR; see IDurableContext.Logger doc.
-    public ILogger Logger => NullLogger.Instance;
+    public ILogger Logger => _logger;
     public IExecutionContext ExecutionContext => new DurableExecutionContext(_durableExecutionArn);
     public ILambdaContext LambdaContext { get; }
+
+    public void ConfigureLogger(LoggerConfig config)
+    {
+        if (config == null) throw new ArgumentNullException(nameof(config));
+
+        // If the user supplies a CustomLogger, wrap it. Otherwise re-wrap the
+        // existing inner logger (unwrapping if it was already a ReplayAwareLogger)
+        // so toggling ModeAware works without losing the previous custom logger.
+        var inner = config.CustomLogger
+            ?? (_logger is ReplayAwareLogger existing ? existing.Inner : _logger);
+        _logger = new ReplayAwareLogger(inner, _state, config.ModeAware);
+    }
 
     public Task<T> StepAsync<T>(
         Func<IStepContext, Task<T>> func,
--- a/Libraries/src/Amazon.Lambda.DurableExecution/DurableFunction.cs
+++ b/Libraries/src/Amazon.Lambda.DurableExecution/DurableFunction.cs
@@ -7,6 +7,7 @@ using Amazon.Lambda.DurableExecution.Internal;
 using Amazon.Lambda.DurableExecution.Services;
 using Amazon.Lambda.Model;
 using Amazon.Runtime;
+using Microsoft.Extensions.Logging;
 
 namespace Amazon.Lambda.DurableExecution;
 
@@ -111,9 +112,20 @@ public static class DurableFunction
         HandlerResult<TOutput> result;
         try
         {
-            result = await DurableExecutionHandler.RunAsync<TOutput>(
-                state, terminationManager,
-                async () => await workflow(userPayload, context));
+            // Push execution-level metadata into a logging scope so structured
+            // providers (the runtime's JSON formatter, Serilog, Powertools,
+            // etc.) tag every log line emitted by user code with the
+            // execution ARN and request id.
+            using (context.Logger.BeginScope(new Dictionary<string, object>
+            {
+                ["durableExecutionArn"] = invocationInput.DurableExecutionArn,
+                ["awsRequestId"] = lambdaContext.AwsRequestId ?? string.Empty,
+            }))
+            {
+                result = await DurableExecutionHandler.RunAsync<TOutput>(
+                    state, terminationManager,
+                    async () => await workflow(userPayload, context));
+            }
 
             await batcher.DrainAsync();
         }
--- a/Libraries/src/Amazon.Lambda.DurableExecution/IDurableContext.cs
+++ b/Libraries/src/Amazon.Lambda.DurableExecution/IDurableContext.cs
@@ -12,12 +12,20 @@ namespace Amazon.Lambda.DurableExecution;
 public interface IDurableContext
 {
     /// <summary>
-    /// A logger scoped to the durable execution. Currently returns
-    /// <see cref="Microsoft.Extensions.Logging.Abstractions.NullLogger.Instance"/>;
-    /// the replay-safe <c>DurableLogger</c> (suppresses messages during replay)
-    /// ships in a follow-up PR.
+    /// Replay-safe logger. Messages emitted while the workflow is re-deriving
+    /// prior operations from checkpointed state are suppressed by default, so
+    /// a 30-step workflow re-invoked 30 times still emits each line once.
+    /// Use this instead of <c>Console.WriteLine</c> or other ambient loggers,
+    /// which will repeat on every replay. Replace the underlying logger or
+    /// disable replay-aware filtering via <see cref="ConfigureLogger"/>.
     /// </summary>
     ILogger Logger { get; }
+
+    /// <summary>
+    /// Swap the underlying logger or toggle replay-aware filtering. Idempotent —
+    /// later calls overwrite earlier configuration.
+    /// </summary>
+    void ConfigureLogger(LoggerConfig config);
 
     /// <summary>
     /// Metadata about the current durable execution.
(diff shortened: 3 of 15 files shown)


philasmar:
### Important

  **1. `IsEnabled` always returns true regardless of runtime log level** (`LambdaCoreLogger.cs:49`)

  ```csharp
  public bool IsEnabled(LogLevel logLevel) => logLevel != LogLevel.None;
```
  The Lambda runtime filters based on AWS_LAMBDA_LOG_LEVEL, but IsEnabled doesn't reflect that. This means LoggerExtensions.LogXxx will still format the message string
  (allocating the FormattedLogValues, running string interpolation) even when the runtime will discard it. For high-frequency logging in step bodies, this could cause unnecessary
   GC pressure.

  Mitigation options: parse AWS_LAMBDA_LOG_LEVEL / DOTNET_LAMBDA_LOG_LEVEL once at construction and cache it. However, this matches the pattern in
  Amazon.Lambda.Logging.AspNetCore.LambdaILogger (which also defers filtering to the runtime), so it's defensible as-is for V1.

  2. Template mutation via scope appending (LambdaCoreLogger.cs:100-135)

  The scope flattening approach appends {key} placeholders to the message template string. This works brilliantly with the Lambda runtime's JSON formatter, but it's worth noting:

  - If a user swaps in a custom logger via ConfigureLogger that doesn't understand the scope-augmented template, the scope keys still land in the template forwarded to
  LambdaCoreLogger -- but that path is taken only when LambdaCoreLogger is the inner. If they swap the inner, their logger gets the original template via ReplayAwareLogger
  delegation. This is fine by construction.
  - The StringBuilder allocation per scoped log call is acceptable given Lambda's per-invocation lifecycle, but worth noting for future perf work.

  3. Thread-safety of _logger field in DurableContext (DurableContext.cs:19, 44)
  
```csharp
  private ILogger _logger;
  // ...
  public ILogger Logger => _logger;
  public void ConfigureLogger(LoggerConfig config) { ... _logger = new ReplayAwareLogger(...); }
```

  The _logger field is read via Logger and written via ConfigureLogger without synchronization. In practice this is safe because durable workflows are single-threaded (the async
  state machine runs on one thread at a time), but if ConfigureLogger is called from one task while another reads Logger, there's a theoretical torn-read concern. A volatile
  annotation would make the intent explicit, though it's not strictly necessary given the execution model.

---
  Questions
  
  1. Should ConfigureLogger be callable after the workflow has started executing steps? Currently nothing prevents it, but calling it mid-workflow could change suppression
  behavior for already-in-progress replay transitions. Is this intentional flexibility or should there be a "frozen after first step" guard?
  2. The LambdaCoreLogger tests use reflection to swap LambdaLogger._loggingWithLevelAction. If that field name changes in a future Amazon.Lambda.Core release, these tests
  silently break (null reference). Is there a way to make this coupling more explicit (e.g., an internal-visible-to or a compile-time check)?

GarrettBeatty:
> ### Important
> 
> **1. `IsEnabled` always returns true regardless of runtime log level** (`LambdaCoreLogger.cs:49`)
> 
> ```cs
> public bool IsEnabled(LogLevel logLevel) => logLevel != LogLevel.None;
> ```
> 
> The Lambda runtime filters based on AWS_LAMBDA_LOG_LEVEL, but IsEnabled doesn't reflect that. This means LoggerExtensions.LogXxx will still format the message string (allocating the FormattedLogValues, running string interpolation) even when the runtime will discard it. For high-frequency logging in step bodies, this could cause unnecessary GC pressure.
> 
> Mitigation options: parse AWS_LAMBDA_LOG_LEVEL / DOTNET_LAMBDA_LOG_LEVEL once at construction and cache it. However, this matches the pattern in Amazon.Lambda.Logging.AspNetCore.LambdaILogger (which also defers filtering to the runtime), so it's defensible as-is for V1.
> 
>     2. Template mutation via scope appending (LambdaCoreLogger.cs:100-135)
> 
> 
> The scope flattening approach appends {key} placeholders to the message template string. This works brilliantly with the Lambda runtime's JSON formatter, but it's worth noting:
> 
>     * If a user swaps in a custom logger via ConfigureLogger that doesn't understand the scope-augmented template, the scope keys still land in the template forwarded to
>       LambdaCoreLogger -- but that path is taken only when LambdaCoreLogger is the inner. If they swap the inner, their logger gets the original template via ReplayAwareLogger
>       delegation. This is fine by construction.
> 
>     * The StringBuilder allocation per scoped log call is acceptable given Lambda's per-invocation lifecycle, but worth noting for future perf work.
> 
> 
>     3. Thread-safety of _logger field in DurableContext (DurableContext.cs:19, 44)
> 
> 
> ```cs
>   private ILogger _logger;
>   // ...
>   public ILogger Logger => _logger;
>   public void ConfigureLogger(LoggerConfig config) { ... _logger = new ReplayAwareLogger(...); }
> ```
> 
> The _logger field is read via Logger and written via ConfigureLogger without synchronization. In practice this is safe because durable workflows are single-threaded (the async state machine runs on one thread at a time), but if ConfigureLogger is called from one task while another reads Logger, there's a theoretical torn-read concern. A volatile annotation would make the intent explicit, though it's not strictly necessary given the execution model.
> 
> Questions
> 
>     1. Should ConfigureLogger be callable after the workflow has started executing steps? Currently nothing prevents it, but calling it mid-workflow could change suppression
>        behavior for already-in-progress replay transitions. Is this intentional flexibility or should there be a "frozen after first step" guard?
> 
>     2. The LambdaCoreLogger tests use reflection to swap LambdaLogger._loggingWithLevelAction. If that field name changes in a future Amazon.Lambda.Core release, these tests
>        silently break (null reference). Is there a way to make this coupling more explicit (e.g., an internal-visible-to or a compile-time check)?

for the logging comment. the runtime will still determine whether to log it or not internally. 

>Template mutation via scope appending (LambdaCoreLogger.cs:100-135) 
i dont think there is any concern here?

>Thread-safety of _logger field in DurableContext (DurableContext.cs:19, 44)
same here i dont really think this is possible?

>Should ConfigureLogger be callable after the workflow has started executing steps? Currently nothing prevents it, but calling it mid-workflow could change suppression
behavior for already-in-progress replay transitions. Is this intentional flexibility or should there be a "frozen after first step" guard?

i also dont think this is a big deal?

>The LambdaCoreLogger tests use reflection to swap LambdaLogger._loggingWithLevelAction. If that field name changes in a future Amazon.Lambda.Core release, these tests
silently break (null reference). Is there a way to make this coupling more explicit (e.g., an internal-visible-to or a compile-time check)?

i dont think this is worth changing

normj (approved):

philasmar (approved):

PR #2373, opened 2026-05-14
Add CreateCallbackAsync and WaitForCallbackAsync (DOTNET-8660)

GarrettBeatty:

Fixes DOTNET-8660.

## What

Adds callback support to `Amazon.Lambda.DurableExecution`. A workflow can now hand a service-allocated `CallbackId` to an external system (a queue consumer, a human-approval UI, a long-running job runner) and suspend until that system reports back via the durable execution service. Two entry points: a low-level handle (`CreateCallbackAsync`) and the common "submit + wait" composition (`WaitForCallbackAsync`).

Public API:

| Type | Purpose |
|---|---|
| `IDurableContext.CreateCallbackAsync<T>(...)` | Allocate a callback; returns an `ICallback<T>` handle. Errors are deferred to `GetResultAsync` so user code between create and await runs deterministically across replays. |
| `IDurableContext.WaitForCallbackAsync<T>(...)` | Composite: `CreateCallback` + submitter step + `GetResultAsync` inside a child context. Common path for "submit job, wait for completion". |
| `ICallback<T>` | Handle exposing `CallbackId` (give to the external system) and `GetResultAsync` (suspends until completion). |
| `IWaitForCallbackContext` | Logger-only context passed to the submitter delegate. Distinct from `IStepContext` so the submitter API can evolve independently. |
| `CallbackConfig` | `Timeout` + `HeartbeatTimeout`. Sub-second positive values are rejected (service timer granularity is 1s); `TimeSpan.Zero` disables. |
| `WaitForCallbackConfig : CallbackConfig` | Adds `RetryStrategy` for the submitter step. |
| `CallbackException` (base) + `CallbackFailedException`, `CallbackTimeoutException`, `CallbackSubmitterException` | Subclass tree so `catch` clauses can pattern-match the failure mode. Carries `CallbackId`, `ErrorType`, `ErrorData`, `OriginalStackTrace`. |

Both APIs read the `ILambdaSerializer` from `ILambdaContext.Serializer` (typically registered via `LambdaBootstrapBuilder.Create(handler, serializer)`) and throw `InvalidOperationException` if no serializer is registered. AOT and reflection-based scenarios share a single overload — the AOT story is determined entirely by the registered serializer (e.g., `SourceGeneratorLambdaJsonSerializer<TContext>` for AOT).

## How

`Internal/CallbackOperation<T>` mirrors the Step/Wait pattern from #2360 and the child-context pattern from #2370:

- **Fresh execution.** Synchronously flushes a `CALLBACK START` checkpoint. The service stamps a freshly-allocated `CallbackId` onto the response; `LambdaDurableServiceClient` gains an `onNewOperations` hook so that ID flows back into `ExecutionState` during the START flush, where the operation can read it. The handle is returned immediately — `CreateCallbackAsync` always succeeds.
- **`GetResultAsync` suspends.** On the invocation that first reaches the await, the workflow hits `Termination.SuspendAndAwait` and Lambda exits. When the external system delivers a result, the service re-invokes; replay observes the terminal checkpoint and returns (or throws) immediately.
- **Replay.** `SUCCEEDED` returns the cached value (deserialized via the registered `ILambdaSerializer`). `FAILED` throws `CallbackFailedException`. `TIMED_OUT` throws `CallbackTimeoutException`. `STARTED` / `PENDING` re-suspend (external system hasn't responded yet). Any other status throws `NonDeterministicExecutionException`.
- **Deferred error propagation.** Terminal status observed during `Start`/`Replay` is stashed on `_terminalReplay` and only resolved inside `GetResultAsync`. This keeps `CreateCallbackAsync` deterministically successful, so user code between create and await sees the same control flow on fresh execution and replay.
- **`WaitForCallbackAsync`** composes `RunInChildContextAsync` (from #2370) + `CreateCallbackAsync` + a submitter `StepAsync` + `GetResultAsync`. The child-context wrapper gives a clean observability boundary (`SubType = WaitForCallback`) and a single error-mapping site: submitter step failures surface as `CallbackSubmitterException`; callback failures/timeouts preserve their subclass through child-context replay (a `CallbackTimeoutException` thrown inside the child remains a `CallbackTimeoutException` after the parent CONTEXT-FAILED replay).

## Testing

42 new unit tests across `CallbackOperationTests`, `WaitForCallbackTests`, `DurableFunctionTests`, and `ExceptionsTests`:

- Fresh execution + sync-flush of `CALLBACK START` (CallbackId stamped onto state).
- Suspend on first `GetResultAsync`; replay returns cached value without re-running.
- Terminal-state replay: `SUCCEEDED` deserializes, `FAILED` throws `CallbackFailedException`, `TIMED_OUT` throws `CallbackTimeoutException`.
- `STARTED`/`PENDING` replay re-suspends; unknown status throws `NonDeterministicExecutionException`.
- `CreateCallbackAsync` is always successful even for terminal-state replays (deferred error propagation).
- `CallbackConfig.Timeout` / `HeartbeatTimeout` validation: rejects negative and sub-second positive values, accepts `TimeSpan.Zero`.
- `WaitForCallbackAsync`: submitter receives the CallbackId; submitter failure (after retries exhausted) surfaces as `CallbackSubmitterException`; callback failure/timeout subclass survives parent CONTEXT-FAILED replay; happy-path returns the deserialized result.
- Exception type hierarchy + property serialization round-trip.

5 new integration tests (require AWS credentials to run):
`CreateCallbackHappyPath`, `CallbackTimeout`, `CallbackFailed`, `WaitForCallbackHappyPath`, `WaitForCallbackSubmitterFails`. Each ships a deployable test function + Dockerfile under `IntegrationTests/TestFunctions/`.

203/203 unit tests pass on net8.0 and net10.0 (161 base + 42 new). Production build clean: 0 warnings, `TreatWarningsAsErrors` enforced.

- `InvokeAsync` / `MapAsync` / `ParallelAsync` / `WaitForConditionAsync`


.autover/changes/118fe72a-a0e5-4119-ae9a-7d6e41ef1b71.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
```
GarrettBeatty: i think this api  and the existing step + wait can be the first preview release

philasmar (commented):
Important Issues

  1. Thread-safety of ExecutionState.AddOperations called from batcher callback
    - File: Libraries/src/Amazon.Lambda.DurableExecution/DurableFunction.cs (line ~99)
    - state.AddOperations is wired as the onNewOperations callback from CheckpointBatcher. The batcher worker runs on a background thread while the user's workflow code reads
  from ExecutionState on the workflow thread. If AddOperations or GetOperation aren't synchronized, this is a data race. Worth verifying that ExecutionState is either thread-safe
   or that the batcher guarantees the callback runs before the EnqueueAsync task completes (which appears to be the case given the sync-flush semantics, but should be
  documented).
  2. CallbackOperation implements ICallback<T> directly — lifetime coupling
    - File: Libraries/src/Amazon.Lambda.DurableExecution/Internal/CallbackOperation.cs
    - The operation is the callback handle returned to user code. This means the user holds a reference to the full DurableOperation<T> machinery (State, Termination, Batcher
  references) for the lifetime of the handle. This is fine functionally, but be aware that if the user stores ICallback<T> long-term (e.g., in a field), those references remain
  rooted. The same pattern exists in the Java SDK so this is probably intentional, but worth a comment noting the design choice.

Nits

  1. Dockerfiles run as root (Semgrep flagged on 7+ Dockerfiles) — same pattern as #2370, test-only containers under Lambda isolation. Low priority.
  2. CallbackConfig rejects negative TimeSpan but doesn't reject TimeSpan.MaxValue — unlikely in practice, but Math.Ceiling(TimeSpan.MaxValue.TotalSeconds) would overflow int.
  Consider capping at a reasonable max (e.g., 7 days = 604800s).

normj (approved):
Assuming you address Phil's comments

GarrettBeatty:
@philasmar addressed your comments in 724d4dee83c9af2e8f4c3da4a1e84ab48d97cfcd

philasmar (approved):

PR #2375, opened 2026-05-14
Add ParallelAsync for concurrent branch execution (DOTNET-8662)

GarrettBeatty:

## What

Adds parallel branch execution to `Amazon.Lambda.DurableExecution`. `ParallelAsync` runs N branches concurrently with configurable concurrency limits and completion policies, returning an `IBatchResult<T>` with per-branch status and error information. The shared `IBatchResult<T>` family is reused by `MapAsync` in Wave 2.

Public API:

| Type | Purpose |
|---|---|
| `IDurableContext.ParallelAsync<T>(Func[], ...)` | Run unnamed branches concurrently. |
| `IDurableContext.ParallelAsync<T>(DurableBranch<T>[], ...)` | Same, but each branch carries an explicit name for traces / test inspection. |
| `DurableBranch<T>(Name, Func)` | Named-branch record. |
| `ParallelConfig` | `MaxConcurrency`, `CompletionConfig`, `NestingType`. |
| `CompletionConfig` | When the batch is considered complete. Factories: `AllSuccessful()`, `FirstSuccessful()`, `AllCompleted()`. Validated `MinSuccessful` / `ToleratedFailureCount` / `ToleratedFailurePercentage` (0.0–1.0). |
| `IBatchResult<T>` | Per-branch view: `All` / `Succeeded` / `Failed` / `Started`, `GetResults`, `GetErrors`, `ThrowIfError`, `HasFailure`, `CompletionReason`, count properties. |
| `IBatchItem<T>` | Single-branch record (`Index`, `Name`, `Status`, `Result`, `Error`). |
| `BatchItemStatus` | `Succeeded` / `Failed` / `Started`. |
| `CompletionReason` | `AllCompleted` / `MinSuccessfulReached` / `FailureToleranceExceeded`. |
| `NestingType` | `Nested` (default); `Flat` reserved for a follow-up PR (throws `NotSupportedException` today). |
| `ParallelException` | Thrown when `CompletionConfig` signals `FailureToleranceExceeded`; carries the `IBatchResult<T>`. |

Per-branch checkpoint payloads are serialized via the `ILambdaSerializer` registered on `ILambdaContext.Serializer` — same pattern as `StepAsync` / `RunInChildContextAsync` from #2370. There are no separate reflection / AOT-safe overload pairs: the AOT story is determined entirely by which serializer the user registers with the runtime (e.g., `SourceGeneratorLambdaJsonSerializer<TContext>`).

## Testing

31 new unit tests in `ParallelOperationTests.cs` and supporting fixtures:

- `CompletionConfig` matrix: `AllSuccessful`, `AllCompleted`, `FirstSuccessful`, `MinSuccessful`, `ToleratedFailureCount`, `ToleratedFailurePercentage` — both pass and fail thresholds.
- Concurrency: `MaxConcurrency` enforced via semaphore; unbounded when null; cancel-mid-dispatch leaves no orphan branches.
- Concurrent `ExecutionState` access regression test (parallel writers do not corrupt the visited-set).
- Replay determinism: mixed-status replay (`SUCCEEDED` + `FAILED` + `STARTED`), `FirstSuccessful` with all-fail, named vs. unnamed branches.
- `IBatchResult<T>` accessors and `GetResults` / `GetErrors` / `ThrowIfError` semantics.
- `NestingType.Flat` throws `NotSupportedException` (placeholder for follow-up).

6 new integration tests build successfully (require AWS credentials to run): happy path, max-concurrency, first-successful, partial-failure, failure-tolerance, and replay-determinism.

### DIFF
--- a/Libraries/src/Amazon.Lambda.DurableExecution/BatchItemStatus.cs
+++ b/Libraries/src/Amazon.Lambda.DurableExecution/BatchItemStatus.cs
@@ -1,0 +1,34 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace Amazon.Lambda.DurableExecution;
+
+/// <summary>
+/// Status of an individual item in a <see cref="IBatchResult{T}"/>.
+/// </summary>
+/// <remarks>
+/// Mirrors the wire-state of the per-branch checkpoint at the moment the batch
+/// resolved. Items that finished produce <see cref="Succeeded"/> or
+/// <see cref="Failed"/>; items that were not dispatched because a
+/// <see cref="CompletionConfig"/> short-circuit fired are reported as
+/// <see cref="Started"/>.
+/// </remarks>
+public enum BatchItemStatus
+{
+    /// <summary>
+    /// The branch ran to completion and produced a result.
+    /// </summary>
+    Succeeded,
+
+    /// <summary>
+    /// The branch ran to completion and threw.
+    /// </summary>
+    Failed,
+
+    /// <summary>
+    /// The branch was not dispatched before the batch's <see cref="CompletionConfig"/>
+    /// resolved (e.g., <see cref="CompletionConfig.FirstSuccessful"/> short-circuited
+    /// before this branch was started), or no per-branch checkpoint exists on replay.
+    /// </summary>
+    Started
+}
--- a/Libraries/src/Amazon.Lambda.DurableExecution/CompletionConfig.cs
+++ b/Libraries/src/Amazon.Lambda.DurableExecution/CompletionConfig.cs
@@ -1,0 +1,114 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace Amazon.Lambda.DurableExecution;
+
+/// <summary>
+/// Defines completion criteria for parallel/map operations.
+/// </summary>
+/// <remarks>
+/// Construct via the static factories (<see cref="AllSuccessful"/>,
+/// <see cref="AllCompleted"/>, <see cref="FirstSuccessful"/>) or set the
+/// individual properties directly. Multiple criteria combine: the operation
+/// resolves as soon as any criterion is met (success short-circuit) or violated
+/// (failure short-circuit).
+/// </remarks>
+public sealed class CompletionConfig
+{
+    private int? _minSuccessful;
+    private int? _toleratedFailureCount;
+    private double? _toleratedFailurePercentage;
+
+    /// <summary>
+    /// Minimum number of <see cref="BatchItemStatus.Succeeded"/> items required
+    /// before the operation resolves successfully. <c>null</c> = no minimum.
+    /// </summary>
+    /// <exception cref="System.ArgumentOutOfRangeException">
+    /// Thrown by the setter if the value is less than <c>1</c>. A minimum of
+    /// zero (or negative) would resolve the operation immediately without
+    /// dispatching any branch.
+    /// </exception>
+    public int? MinSuccessful
+    {
+        get => _minSuccessful;
+        set
+        {
+            if (value is { } v && v < 1)
+            {
+                throw new ArgumentOutOfRangeException(nameof(value), v,
+                    "MinSuccessful must be at least 1.");
+            }
+            _minSuccessful = value;
+        }
+    }
+
+    /// <summary>
+    /// Maximum tolerated <see cref="BatchItemStatus.Failed"/> count. When the
+    /// failure count <i>strictly exceeds</i> this value, the operation resolves
+    /// with <see cref="CompletionReason.FailureToleranceExceeded"/>.
+    /// <c>null</c> = no count-based failure threshold.
+    /// </summary>
+    /// <exception cref="System.ArgumentOutOfRangeException">
+    /// Thrown by the setter if the value is negative. A negative tolerance
+    /// would fail the operation immediately without dispatching any branch.
+    /// </exception>
+    public int? ToleratedFailureCount
+    {
+        get => _toleratedFailureCount;
+        set
+        {
+            if (value is { } v && v < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(value), v,
+                    "ToleratedFailureCount must be zero or greater.");
+            }
+            _toleratedFailureCount = value;
+        }
+    }
+
+    /// <summary>
+    /// Maximum tolerated failure ratio, expressed as a value in the range
+    /// <c>0.0</c> to <c>1.0</c> (inclusive). For example, <c>0.25</c> means
+    /// "tolerate up to 25% failures; fail when the failure ratio strictly
+    /// exceeds 25%". <c>null</c> = no ratio-based failure threshold.
+    /// </summary>
+    /// <exception cref="System.ArgumentOutOfRangeException">
+    /// Thrown by the setter if the value is outside <c>[0.0, 1.0]</c>.
+    /// </exception>
+    public double? ToleratedFailurePercentage
+    {
+        get => _toleratedFailurePercentage;
+        set
+        {
+            if (value is { } v && (v < 0.0 || v > 1.0))
+            {
+                throw new ArgumentOutOfRangeException(nameof(value), v,
+                    "ToleratedFailurePercentage must be a ratio in [0.0, 1.0].");
+            }
+            _toleratedFailurePercentage = value;
+        }
+    }
+
+    /// <summary>
+    /// All items must succeed. Equivalent to
+    /// <see cref="ToleratedFailureCount"/> = 0. The default for
+    /// <see cref="ParallelConfig.CompletionConfig"/>.
+    /// </summary>
+    public static CompletionConfig AllSuccessful() => new() { ToleratedFailureCount = 0 };
+
+    /// <summary>
+    /// Run every branch regardless of failures; surface failures per-item via
+    /// <see cref="IBatchResult{T}.Failed"/>. Resolution does not auto-throw —
+    /// the caller can inspect the result and call
+    /// <see cref="IBatchResult{T}.ThrowIfError"/> if they want strict-success
+    /// behavior.
+    /// </summary>
+    public static CompletionConfig AllCompleted() => new();
+
+    /// <summary>
+    /// Resolve once at least one branch has succeeded. Branches that were not
+    /// dispatched before the completion criteria was met are reported as
+    /// <see cref="BatchItemStatus.Started"/>.
+    /// </summary>
+    public static CompletionConfig FirstSuccessful() => new() { MinSuccessful = 1 };
+}
--- a/Libraries/src/Amazon.Lambda.DurableExecution/CompletionReason.cs
+++ b/Libraries/src/Amazon.Lambda.DurableExecution/CompletionReason.cs
@@ -1,0 +1,32 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace Amazon.Lambda.DurableExecution;
+
+/// <summary>
+/// Why a batch operation (<see cref="IDurableContext.ParallelAsync{T}(IReadOnlyList{System.Func{IDurableContext, System.Threading.CancellationToken, System.Threading.Tasks.Task{T}}}, string?, ParallelConfig?, System.Threading.CancellationToken)"/>
+/// or future Map) resolved.
+/// </summary>
+public enum CompletionReason
+{
+    /// <summary>
+    /// Every branch finished — no <see cref="CompletionConfig"/> short-circuit
+    /// was triggered. Branches may be a mix of <see cref="BatchItemStatus.Succeeded"/>
+    /// and <see cref="BatchItemStatus.Failed"/>.
+    /// </summary>
+    AllCompleted,
+
+    /// <summary>
+    /// <see cref="CompletionConfig.MinSuccessful"/> branches succeeded; remaining
+    /// branches were left in <see cref="BatchItemStatus.Started"/>.
+    /// </summary>
+    MinSuccessfulReached,
+
+    /// <summary>
+    /// <see cref="CompletionConfig.ToleratedFailureCount"/> or
+    /// <see cref="CompletionConfig.ToleratedFailurePercentage"/> was exceeded.
+    /// The batch is considered failed and surfaces a
+    /// <see cref="ParallelException"/> when awaited.
+    /// </summary>
+    FailureToleranceExceeded
+}
--- a/Libraries/src/Amazon.Lambda.DurableExecution/DurableBranch.cs
+++ b/Libraries/src/Amazon.Lambda.DurableExecution/DurableBranch.cs
@@ -1,0 +1,18 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace Amazon.Lambda.DurableExecution;
+
+/// <summary>
+/// A named branch for
+/// <see cref="IDurableContext.ParallelAsync{T}(IReadOnlyList{DurableBranch{T}}, string?, ParallelConfig?, System.Threading.CancellationToken)"/>.
+/// Names appear in execution traces and on the wire <c>OperationUpdate.Name</c>
+/// field, and surface on <see cref="IBatchItem{T}.Name"/>.
+/// </summary>
+/// <typeparam name="T">The branch's result type.</typeparam>
+/// <param name="Name">Human-readable branch name. Required.</param>
+/// <param name="Func">The user function executed inside the branch's child
+/// context. It receives the branch's <see cref="IDurableContext"/> and a
+/// <see cref="System.Threading.CancellationToken"/> linking the caller-supplied
+/// token with the SDK's workflow-shutdown signal.</param>
+public sealed record DurableBranch<T>(string Name, Func<IDurableContext, System.Threading.CancellationToken, Task<T>> Func);
--- a/Libraries/src/Amazon.Lambda.DurableExecution/IBatchItem.cs
+++ b/Libraries/src/Amazon.Lambda.DurableExecution/IBatchItem.cs
@@ -1,0 +1,41 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace Amazon.Lambda.DurableExecution;
+
+/// <summary>
+/// One item inside an <see cref="IBatchResult{T}"/> — the outcome of a single
+/// branch (parallel) or item (map).
+/// </summary>
+/// <typeparam name="T">The branch/item result type.</typeparam>
+public interface IBatchItem<T>
+{
+    /// <summary>
+    /// Zero-based position in the original branches/items list. Stable across
+    /// replays.
+    /// </summary>
+    int Index { get; }
+
+    /// <summary>
+    /// Optional human-readable name for this branch/item.
+    /// Surfaces on the wire <c>OperationUpdate.Name</c> field for observability.
+    /// </summary>
+    string? Name { get; }
+
+    /// <summary>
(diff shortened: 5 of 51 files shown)


Libraries/src/Amazon.Lambda.DurableExecution/NestingType.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </summary>
+    Nested,
+
+    /// <summary>
+    /// Branches use virtual contexts sharing the parent. Reduces checkpoint
+    /// cost at the expense of less granular execution traces.
+    /// </summary>
+    /// <remarks>
+    /// Not yet implemented in the .NET SDK; passing this value throws
+    /// <see cref="System.NotSupportedException"/>.
```
GarrettBeatty, line 37: will do in another pr

philasmar:
Important
  
  1. ExecutionState single-lock may bottleneck for high branch counts

  The switch from ConcurrentDictionary to a single lock is well-reasoned (compound operations in LoadFromCheckpoint need atomicity). Each critical section is O(1). For the branch
   counts in this PR's tests (3-10), contention is negligible. However, the lock is on the read path (GetOperation, IsReplaying) that every branch hits during replay. If MapAsync
   (mentioned as a follow-up) dispatches hundreds of items, this could become measurable. No action needed now — just flag for the MapAsync PR to benchmark.

  2. Dispatched branches are never cancelled on short-circuit

  The design explicitly states: once dispatched, branches are never cancelled — even after a MinSuccessful or FailureToleranceExceeded short-circuit fires. This guarantees replay
   determinism (all dispatched branches end in a terminal state), but it means with MaxConcurrency=null and FirstSuccessful, all branches execute to completion even though only
  the first one's result matters. The parallel.md doc doesn't mention this tradeoff. Consider adding a note like: "Dispatched branches always run to completion for replay
  determinism — use MaxConcurrency to limit wasted compute under FirstSuccessful."

  3. ParallelException.Result is type-erased (IBatchResult not IBatchResult<T>)

  public IBatchResult? Result { get; init; }

  Since ParallelException isn't generic, the result is boxed behind the non-generic IBatchResult marker. Callers who catch ParallelException need to cast:
  (IBatchResult<MyType>)ex.Result. This is fine, but worth noting in the exception's XML doc that the cast is expected (it's mentioned on ParallelException but a snippet would
  help users).

GarrettBeatty:
> Important
> 
>     1. ExecutionState single-lock may bottleneck for high branch counts
> 
> 
> The switch from ConcurrentDictionary to a single lock is well-reasoned (compound operations in LoadFromCheckpoint need atomicity). Each critical section is O(1). For the branch counts in this PR's tests (3-10), contention is negligible. However, the lock is on the read path (GetOperation, IsReplaying) that every branch hits during replay. If MapAsync (mentioned as a follow-up) dispatches hundreds of items, this could become measurable. No action needed now — just flag for the MapAsync PR to benchmark.
> 
>     2. Dispatched branches are never cancelled on short-circuit
> 
> 
> The design explicitly states: once dispatched, branches are never cancelled — even after a MinSuccessful or FailureToleranceExceeded short-circuit fires. This guarantees replay determinism (all dispatched branches end in a terminal state), but it means with MaxConcurrency=null and FirstSuccessful, all branches execute to completion even though only the first one's result matters. The parallel.md doc doesn't mention this tradeoff. Consider adding a note like: "Dispatched branches always run to completion for replay determinism — use MaxConcurrency to limit wasted compute under FirstSuccessful."
> 
>     3. ParallelException.Result is type-erased (IBatchResult not IBatchResult)
> 
> 
> public IBatchResult? Result { get; init; }
> 
> Since ParallelException isn't generic, the result is boxed behind the non-generic IBatchResult marker. Callers who catch ParallelException need to cast: (IBatchResult)ex.Result. This is fine, but worth noting in the exception's XML doc that the cast is expected (it's mentioned on ParallelException but a snippet would help users).

fixed number 2 only, the rest are already documented https://github.com/aws/aws-lambda-dotnet/pull/2375/commits/3179531a509634a56488a2c2b637f58f14ceda8e

Libraries/src/Amazon.Lambda.DurableExecution/Internal/BatchItem.cs
```diff
@@ -0,0 +1,15 @@
+namespace Amazon.Lambda.DurableExecution.Internal;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.DurableExecution/Internal/BatchResult.cs
```diff
@@ -0,0 +1,80 @@
+namespace Amazon.Lambda.DurableExecution.Internal;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.DurableExecution/Internal/ParallelJsonContext.cs
```diff
@@ -0,0 +1,15 @@
+using System.Text.Json.Serialization;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.DurableExecution/Internal/ParallelOperation.cs
```diff
@@ -0,0 +1,654 @@
+using System.IO;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.DurableExecution/Internal/ParallelSummary.cs
```diff
@@ -0,0 +1,31 @@
+using System.Text.Json.Serialization;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.DurableExecution/BatchItemStatus.cs
```diff
@@ -0,0 +1,31 @@
+namespace Amazon.Lambda.DurableExecution;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.DurableExecution/CompletionConfig.cs
```diff
@@ -0,0 +1,111 @@
+namespace Amazon.Lambda.DurableExecution;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.DurableExecution/CompletionReason.cs
```diff
@@ -0,0 +1,29 @@
+namespace Amazon.Lambda.DurableExecution;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.DurableExecution/DurableBranch.cs
```diff
@@ -0,0 +1,13 @@
+namespace Amazon.Lambda.DurableExecution;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.DurableExecution/IBatchItem.cs
```diff
@@ -0,0 +1,38 @@
+namespace Amazon.Lambda.DurableExecution;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.DurableExecution/IBatchResult.cs
```diff
@@ -0,0 +1,90 @@
+namespace Amazon.Lambda.DurableExecution;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.DurableExecution/NestingType.cs
```diff
@@ -0,0 +1,37 @@
+namespace Amazon.Lambda.DurableExecution;
```
philasmar, line 4: add license header

Libraries/src/Amazon.Lambda.DurableExecution/ParallelConfig.cs
```diff
@@ -0,0 +1,57 @@
+namespace Amazon.Lambda.DurableExecution;
```
philasmar, line 4: add license header

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/TestFunctions/ParallelFailureToleranceFunction/Function.cs
```diff
@@ -0,0 +1,60 @@
+using Amazon.Lambda.Core;
```
philasmar, line 4: add license header

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/TestFunctions/ParallelFirstSuccessfulFunction/Function.cs
```diff
@@ -0,0 +1,79 @@
+using Amazon.Lambda.Core;
```
philasmar, line 4: add license header

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/TestFunctions/ParallelHappyPathFunction/Function.cs
```diff
@@ -0,0 +1,40 @@
+using Amazon.Lambda.Core;
```
philasmar, line 4: add license header

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/TestFunctions/ParallelMaxConcurrencyFunction/Function.cs
```diff
@@ -0,0 +1,67 @@
+using Amazon.Lambda.Core;
```
philasmar, line 4: add license header

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/TestFunctions/ParallelPartialFailureFunction/Function.cs
```diff
@@ -0,0 +1,61 @@
+using Amazon.Lambda.Core;
```
philasmar, line 4: add license header

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/TestFunctions/ParallelReplayDeterminismFunction/Function.cs
```diff
@@ -0,0 +1,57 @@
+using Amazon.Lambda.Core;
```
philasmar, line 4: add license header

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/ParallelFailureToleranceTest.cs
```diff
@@ -0,0 +1,70 @@
+using System.Linq;
```
philasmar, line 4: add license header

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/ParallelFirstSuccessfulTest.cs
```diff
@@ -0,0 +1,81 @@
+using System.Linq;
```
philasmar, line 4: add license header

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/ParallelHappyPathTest.cs
```diff
@@ -0,0 +1,72 @@
+using System.Linq;
```
philasmar, line 4: add license header

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/ParallelMaxConcurrencyTest.cs
```diff
@@ -0,0 +1,76 @@
+using System.Linq;
```
philasmar, line 4: add license header

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/ParallelPartialFailureTest.cs
```diff
@@ -0,0 +1,74 @@
+using System.Linq;
```
philasmar, line 4: add license header

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/ParallelReplayDeterminismTest.cs
```diff
@@ -0,0 +1,122 @@
+using System.Linq;
```
philasmar, line 4: add license header

Libraries/test/Amazon.Lambda.DurableExecution.Tests/ParallelOperationTests.cs
```diff
@@ -0,0 +1,1163 @@
+using Amazon.Lambda.DurableExecution;
```
philasmar, line 4: add license header

philasmar (approved):

Libraries/src/Amazon.Lambda.DurableExecution/ParallelConfig.cs
```diff
@@ -0,0 +1,60 @@
...
+    public CompletionConfig CompletionConfig { get; set; } = CompletionConfig.AllSuccessful();
+
+    /// <summary>
+    /// How branches are represented in the checkpoint graph. Defaults to
+    /// <see cref="NestingType.Nested"/>.
+    /// </summary>
+    /// <remarks>
+    /// <see cref="NestingType.Flat"/> is not yet supported in the .NET SDK and
+    /// will throw <see cref="System.NotSupportedException"/> when the parallel
+    /// operation is invoked.
```
GarrettBeatty, line 57: will be implemented in https://github.com/aws/aws-lambda-dotnet/pull/2409

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/TestFunctions/ParallelFailureToleranceFunction/ParallelFailureToleranceFunction.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
normj: All test projects should target .NET 10

Libraries/src/Amazon.Lambda.DurableExecution/Internal/ParallelOperation.cs
```diff
@@ -0,0 +1,657 @@
...
+/// </list>
+/// Per-branch errors do NOT abort the parallel directly — the orchestrator
+/// catches each branch's <see cref="ChildContextException"/>, records it as a
+/// failed <see cref="IBatchItem{T}"/>, and consults the
+/// <see cref="CompletionConfig"/> after every completion. Only when the
+/// completion config marks the run as
+/// <see cref="CompletionReason.FailureToleranceExceeded"/> does the parallel
+/// throw.
+/// </remarks>
+internal sealed class ParallelOperation<T> : DurableOperation<IBatchResult<T>>
```
GarrettBeatty, line 43: this whole class is going to get refactored later on in https://github.com/aws/aws-lambda-dotnet/pull/2408 to support mapasync as well

Libraries/src/Amazon.Lambda.DurableExecution/Internal/ParallelOperation.cs
```diff
@@ -0,0 +1,657 @@
...
+                // a short-circuit fires — orphan branches that continue
+                // writing checkpoints would diverge between the original run
+                // and replay. Letting them finish guarantees determinism: all
+                // dispatched branches end up Succeeded or Failed. Only
+                // un-dispatched branches surface as Started.
+                if (inFlight.Count > 0)
+                {
+                    try
+                    {
+                        await Task.WhenAll(inFlight).ConfigureAwait(false);
```
normj: As we discussed on the side. We need a mechanism to communicate to the inflight branches that they should cleanly shut down. For example provide them their own cancellation token that we cancel from here.

normj (changes requested):

Libraries/src/Amazon.Lambda.DurableExecution/Internal/ParallelOperation.cs
GarrettBeatty: @normj i think in your example on the call, you were thinking something like 

```

  new ParallelConfig
  {
      MaxConcurrency = 10,
      CompletionConfig = new CompletionConfig { MinSuccessful = 2 }
  }
```

in this case the short circuit would not fire and we would await all 10 tasks.

but in the case like 

```

  new ParallelConfig
  {
      MaxConcurrency = 3,
      CompletionConfig = new CompletionConfig { MinSuccessful = 2 }
  }
```

### When does `MinSuccessful` actually save work?

  Lower the concurrency window. Same 10 branches, `MaxConcurrency = 3`, `MinSuccessful = 2`:

  | Time | In flight       | Done   | `succeeded` | Action                                                                                              |
  | ---- | --------------- | ------ | ----------- | --------------------------------------------------------------------------------------------------- |
  | t0   | branches 0,1,2  | –      | 0           | semaphore full; dispatch loop blocked on `WaitAsync`                                                |
  | t1   | branches 1,2,3  | 0      | 1           | branch 0 finishes → releases slot → loop wakes, dispatches branch 3                                 |
  | t2   | branches 2,3    | 0,1    | 2           | branch 1 finishes → `ShouldStopDispatching` is true → release slot, `break` out of dispatch loop    |
  | t3   | –               | 0,1,2,3| 2+          | `Task.WhenAll` settles the two in-flight branches (2 and 3)                                         |
  | end  | –               | 0,1,2,3| –           | branches 4–9 were never dispatched; they surface as `Started`                                       |

  Result: 4 `Succeeded`, 6 `Started`, `CompletionReason = MinSuccessfulReached` — six branches' worth of compute saved.

GarrettBeatty: and as far as cancellation token goes - i think you were saying we can still cancel the inflight ones. but my only thing i would need to think about is if cancellation -> FAILED status. i think in other places (i would need to double check), we decided cancellation is different than having it failed.

i think theres some edge cases where if we just cancel the task, and some checkpoints are written already (and we just cancel it without it letting write the final checkpoints) then the replay will be corrupted

normj: But given the default of `MaxConcurrency` is the number of branches I suspect we will get more task running at a time then `MinSuccessful`.

You also have the problem which is probably more common that if you have an exception in one of the branches and the configuration is to break if any exceptions have occurred we still wait for all inflight branches even though they won't matter because we are going to return a failure back.

philasmar (approved):

Libraries/src/Amazon.Lambda.DurableExecution/Internal/ParallelOperation.cs
GarrettBeatty: I added in a `shortCircuitCts`  which does this. 

`shortCircuitCts`  is  per-parallel CancellationTokenSource that gets `Cancel()ed` the moment a CompletionConfig short-circuit is decided (at a dispatch-loop break or in onComplete), and its token flows only into each running branch's user Func

GarrettBeatty: https://github.com/aws/aws-lambda-dotnet/pull/2375/commits/fe7c417f9e428fe12908dc29e378d0c321ebb994

GarrettBeatty: i also wired in the workflowcancellation where applicable too

normj (approved):

PR #2376, opened 2026-05-14
Add WaitForConditionAsync polling primitive (DOTNET-8665)

GarrettBeatty:

## What

Adds service-mediated polling to `Amazon.Lambda.DurableExecution`. `WaitForConditionAsync` repeatedly evaluates a check function with a configurable wait strategy between attempts; each iteration runs in its own Lambda invocation (suspended via `STEP` + `RETRY` checkpoints carrying `NextAttemptDelaySeconds`), so polling does not consume compute time while waiting between checks.

Public API:

| Type | Purpose |
|---|---|
| `IDurableContext.WaitForConditionAsync<TState>(...)` | Poll until the user-supplied predicate signals done or `maxAttempts` is exhausted. Per-iteration state is serialized via the `ILambdaSerializer` registered on `ILambdaContext.Serializer` (same pattern as `StepAsync` / `RunInChildContextAsync`). |
| `IConditionCheckContext` | Passed to the check func: `Logger`, `AttemptNumber`. |
| `WaitForConditionConfig<TState>` | Required `InitialState` + `WaitStrategy`. |
| `IWaitStrategy<TState>` | `Decide(state, attempt)` returning `WaitDecision`. |
| `WaitDecision` | Readonly record struct: `ShouldContinue`, `Delay`, factories `Stop()` / `ContinueAfter(TimeSpan)`. |
| `WaitStrategy` factories | `Exponential`, `Linear`, `Fixed`, `FromDelegate` — each accepting an optional `Func<TState, bool> isDone` predicate. |
| `WaitForConditionException` | Thrown when `maxAttempts` is exhausted; carries `AttemptsExhausted` and `LastState` (preserved across both live execution and replay). |

## How

`Internal/WaitForConditionOperation<TState>` runs as a `STEP` with `SubType = "WaitForCondition"`:

- **Each iteration** calls the user check function with the current state. The strategy's `Decide(state, attempt)` returns either `Stop()` (success — return state) or `ContinueAfter(delay)` (suspend until next attempt). Continuing emits `Action=RETRY` with the new state in the payload and `NextAttemptDelaySeconds` set so the durable execution service schedules the next invocation without consuming compute time.
- **Strategies** signal max-attempts exhausted by throwing `WaitForConditionException` directly from `Decide()`; the operation enriches with `LastState` before checkpointing FAIL. `LastState` survives FAIL replay: serialized into the FAIL payload at write time and deserialized in `BuildFailureException` with a warning-logged fallback for legacy / corrupt data.
- **Serialization** is delegated to the registered `ILambdaSerializer` via stream-based `Serialize<T>` / `Deserialize<T>` calls — no AOT trim attributes on the public API. Mirrors `StepOperation` / `ChildContextOperation`.
- **`ExponentialBackoff` helper** is extracted for sharing between this operation and `ExponentialRetryStrategy`. Math is byte-for-byte identical to the existing implementation.

**Defaults:** 60 attempts, 5s initial delay, 300s max delay, 1.5x rate, Full jitter — distinct from `RetryStrategy.Default` and matching Python/JS/Java reference SDKs.

> **Cross-SDK note:** Python returns success on max-attempts exhausted; .NET / Java / JS throw. Workflows ported from Python should review for new failure modes. Documented in the design doc.

### DIFF
--- a/.autover/changes/ef22a418-1be3-4359-a442-f086667635ec.json
+++ b/.autover/changes/ef22a418-1be3-4359-a442-f086667635ec.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.DurableExecution",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Add WaitForConditionAsync polling primitive: IDurableContext.WaitForConditionAsync<TState>, IConditionCheckContext, WaitForConditionConfig<TState>, IWaitStrategy<TState>, WaitDecision, WaitStrategy factory (Exponential/Linear/Fixed/FromDelegate), and WaitForConditionException with AttemptsExhausted and LastState"
+      ]
+    }
+  ]
+}
--- a/Docs/durable-execution-design.md
+++ b/Docs/durable-execution-design.md
@@ -424,6 +424,29 @@ await context.WaitAsync(TimeSpan.FromDays(7), name: "weekly_reminder");
 
 > **Validation:** The duration must be at least 1 second. Values less than 1 second throw `ArgumentOutOfRangeException`. Sub-second precision is truncated to whole seconds (the underlying service operates at second granularity).
 
+#### Wait For Condition
+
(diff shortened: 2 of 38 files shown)


philasmar (approved):

normj (approved):
I notice the public methods on `IDurableContext` documentation is missing the `param` elements for each of the method arguments. We should get those added at some point before we do a release.

PR #2387, opened 2026-05-21
Fix thread pool starvation under multi-concurrency

zachvan2:
## Summary

- **Set polling task count to MC level** instead of `Math.Max(2, processorCount)` — ensures enough `/next` calls are pending to avoid `Runtime.Unavailable` race conditions
- **Pre-size the thread pool** (`SetMinThreads`) to `MC + ProcessorCount` so polling loop continuations aren't starved by handler work

## Problem

With the current defaults (2 polling tasks on 1 vCPU), the runtime doesn't call RAPID's `/next` fast enough under load. When handlers do blocking work, the .NET thread pool (which defaults `MinThreads` to `ProcessorCount`) can't resume the polling loop's `await` continuation, causing RAPID to timeout with `Runtime.Unavailable`.

## Benchmark Results

Tested with RIE across configurations from MC=8/1vCPU to MC=128/4vCPU with both `Thread.Sleep` and CPU-bound workloads:
- **Without fix**: Success rate drops as soon as load exceeds polling task count
- **With fix**: 100% success up to MC, graceful degradation beyond
- Improvement ranges from **+26%** to **+93%** success rate

## Test plan

- [x] Existing `LambdaBootstrapMultiConcurrencyTests` pass
- [x] Updated `UtilsTest.DetermineProcessingTaskCount` cases reflect new defaults
- [x] New `UtilsTest.GetMaxConcurrency` tests cover parsing edge cases
- [x] RIE integration benchmarks at MC=8,16,32,64,128 with 1,2,4 vCPUs


trivenay (approved):

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/LambdaBootstrap.cs
```diff
@@ -629,6 +601,48 @@ private void AdjustMemorySettings()
...
+        /// When running in multi-concurrency mode, pre-size the .NET ThreadPool to ensure there are enough
+        /// threads available for both handler execution and polling task continuations. Without this,
+        /// blocking handlers (Thread.Sleep, .Result, .Wait()) can exhaust the ThreadPool, preventing
+        /// polling tasks from cycling back to /next and causing Runtime.Unavailable errors from RAPID.
+        /// </summary>
+        private void AdjustThreadPoolSettings()
+        {
+            try
+            {
+                var maxConcurrency = Utils.GetMaxConcurrency(_environmentVariables);
```
normj, line 647: I'm thinking we should check if `Constants.ENVIRONMENT_VARIABLE_AWS_LAMBDA_DOTNET_PROCESSING_TASKS` is set or create a separate environment variable to check. If that environment is set then use that value over `AWS_LAMBDA_MAX_CONCURRENCY`. That way if a customer finds in their work load that the number of threads we are creating hurts their performance they still have a knob they can use to adjust. For most flexibility in this advanced use case I prefer a separate environment, maybe `AWS_LAMBDA_DOTNET_MIN_THREADS`

normj (changes requested):

Libraries/src/Amazon.Lambda.RuntimeSupport/Bootstrap/LambdaBootstrap.cs
zachvan2: Thanks for thinking about this in some detail, but I'm worried this might be unneeded complexity.  If customers want fewer threads, they can set lower MC, and if they want fewer processing tasks, they can set that.  Is there a use case you can think of where threads could hurt performance?

normj: It would depend on the type of work load load happening but what I was thinking is customers might find that creating the MC number of threads might use up more resources then they want and for they don't need that many threads to handle the pollers.

Basically my concern comes from it is hard to predict all of the types of workloads and if we get a customer that is having issues with their workload it would be better to have these possible knobs for them to tweak instead of us having to do a deployment change to help them out.

normj:
Merged with https://github.com/aws/aws-lambda-dotnet/pull/2395

PR #2403, opened 2026-06-02
Reapply Lambda Architecture when deploy test app

normj:
*Description of changes:*
Now that the ARM tests are back in the pipeline the integ tests for the Annotations TestServerlessApp project fail because the containers are being built using ARM architecture but the functions are being deployed defaulting to X64. This commit https://github.com/aws/aws-lambda-dotnet/commit/fcb5208ec20b44522afdab3c1b10d9ce1ccebdd7#diff-9990804bd7941f64d259a5da783877ca3153cbb32aac36e0fa7b54dbc787c3ed merged a few months ago accidently removed the CloudFormation template parameter for the deployment to set the architecture. 

The PR adds back the CloudFormation template parameter that the `DeploymentScript.ps1` script used to deploy was already setting.

### DIFF
--- a/Libraries/test/TestServerlessApp/serverless.template
+++ b/Libraries/test/TestServerlessApp/serverless.template
@@ -2,6 +2,30 @@
   "AWSTemplateFormatVersion": "2010-09-09",
   "Transform": "AWS::Serverless-2016-10-31",
   "Description": "This template is partially managed by Amazon.Lambda.Annotations (v2.0.1.0).",
+  "Parameters": {
+    "ArchitectureTypeParameter": {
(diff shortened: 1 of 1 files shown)


peterrsongg (approved):

AlexDaines (approved):

GarrettBeatty (approved):
I think in other places I actually made it so the integration test (some cs file or script) dynamically adds stuff to the template. The reason being is because for testing developers usually deleted the whole file and recreated it which would lose the value. Might be good to do similar (I think you can search for sqs queue or similar in the integration test)

PR #2408, opened 2026-06-05
Add MapAsync for concurrent collection processing

GarrettBeatty:

## What

Adds `IDurableContext.MapAsync` to `Amazon.Lambda.DurableExecution`. `MapAsync` processes a collection in parallel with one child context per item, mirroring the Python/JS/Java SDKs where `Map` is a sibling of `Parallel` sharing one concurrency engine. It reuses the `IBatchResult<T>` family and concurrency/completion machinery introduced by `ParallelAsync` in #2375.

Public API:

| Type | Purpose |
|---|---|
| `IDurableContext.MapAsync<TItem, TResult>(IReadOnlyList<TItem> items, Func<IDurableContext, TItem, int, IReadOnlyList<TItem>, CancellationToken, Task<TResult>> func, string? name, MapConfig? config, CancellationToken)` | Process each item concurrently in its own child context, returning `Task<IBatchResult<TResult>>`. The per-item func receives the durable context, the item, its zero-based index, the full source list (matching the Python/JS SDKs), and a `CancellationToken` linking the caller's token with the SDK's workflow-shutdown signal (also tripped cooperatively when a sibling satisfies the `CompletionConfig` and the map short-circuits). |
| `MapConfig` | `MaxConcurrency` (`int?`, null = unlimited, must be ≥1), `CompletionConfig`, `NestingType`, `ItemNamer`. Defaults `CompletionConfig` to `AllCompleted()` (permissive), matching Python/Java `Map` — intentionally differs from `ParallelConfig`'s `AllSuccessful()`. `NestingType` defaults to `Nested`; `NestingType.Flat` is not yet supported and throws `NotSupportedException` when invoked. |
| `MapConfig.ItemNamer` | `Func<object, int, string>?` — supplies a per-item branch name (item + zero-based index) surfaced in traces and on `IBatchItem<T>.Name`. When null, branches are named by index (`"0"`, `"1"`, ...). (No `ItemBatcher` — not implemented in any reference SDK.) |
| `MapException` | Thrown when `CompletionConfig` signals `FailureToleranceExceeded`, so callers can distinguish `Map` failures from `Parallel` failures. Exposes a type-erased `IBatchResult? Result` (cast to `IBatchResult<T>` when the item type is known) and the `CompletionReason`. Base type for map failures; catching it stays forward-compatible. |

Behavior: by default (`AllCompleted`) every item runs regardless of per-item failures, which are captured on the corresponding `IBatchItem<T>` and surfaced via `IBatchResult<T>.Failed` rather than thrown. The map throws `MapException` only when the `CompletionConfig` criteria are violated. For fail-fast semantics, set `CompletionConfig` to `AllSuccessful()` or call `IBatchResult<T>.ThrowIfError()` on the result.

Implementation notes:

- Extracts a `ConcurrentOperation<T>` base holding all orchestration, completion, checkpoint, and replay logic. `ParallelOperation` and `MapOperation` are thin subclasses supplying only the per-unit `(name, func)`, sub-type labels, and failure-exception factory. Adds `Map` / `MapItem` to `OperationSubTypes`.
- Generalizes `ParallelSummary` / `ParallelJsonContext` into shared `BatchSummary` / `BatchJsonContext`.

Per-item checkpoint payloads are serialized via the `ILambdaSerializer` registered on `ILambdaContext.Serializer` — the same pattern as `ParallelAsync` / `StepAsync` / `RunInChildContextAsync`. The AOT story is determined entirely by which serializer the user registers with the runtime (e.g., `SourceGeneratorLambdaJsonSerializer<TContext>`). `MapConfig` does not expose a serializer slot.

## Testing

24 new unit tests in `MapOperationTests.cs`, mirroring the `Parallel` set:

- `CompletionConfig` matrix: `AllSuccessful`, `AllCompleted`, `FirstSuccessful`, `MinSuccessful`, `ToleratedFailureCount`, `ToleratedFailurePercentage` — both pass and fail thresholds.
- Concurrency: `MaxConcurrency` enforced; unbounded when null.
- Replay determinism: mixed-status replay, named vs. unnamed items via `ItemNamer`.
- `IBatchResult<T>` accessors and `GetResults` / `GetErrors` / `ThrowIfError` semantics.

6 new integration functions/tests mirroring the `Parallel` set (require AWS credentials to run): `MapHappyPath`, `MapPartialFailure`, `MapFailureTolerance`, `MapFirstSuccessful`, `MapMaxConcurrency`, `MapReplayDeterminism`.


Libraries/src/Amazon.Lambda.DurableExecution/Internal/BatchSummary.cs
```diff
@@ -0,0 +1,33 @@
...
+/// children's own checkpoints.
+/// </summary>
+internal sealed class BatchSummary
+{
+    [JsonPropertyName("CompletionReason")]
+    public string? CompletionReason { get; set; }
+
+    [JsonPropertyName("Units")]
+    public IList<BatchUnitSummary> Units { get; set; } = new List<BatchUnitSummary>();
+}
```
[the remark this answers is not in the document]
GarrettBeatty: this is fine - since we are in preview

Libraries/src/Amazon.Lambda.DurableExecution/Internal/ParallelOperation.cs
```diff
@@ -1,53 +1,22 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
-using System.IO;
-using System.Text;
 using Amazon.Lambda;
 using Amazon.Lambda.Core;
-using SdkErrorObject = Amazon.Lambda.Model.ErrorObject;
-using SdkOperationUpdate = Amazon.Lambda.Model.OperationUpdate;
 
 namespace Amazon.Lambda.DurableExecution.Internal;
```
GarrettBeatty, line 7: all the common logic (mostly everything) shared between map and parallel was moved to ConcurrentOperation

Libraries/src/Amazon.Lambda.DurableExecution/Internal/ConcurrentOperation.cs
```diff
@@ -0,0 +1,694 @@
...
+        var message =
+            $"{OperationNoun} operation failed: failure tolerance exceeded " +
+            $"({result.FailureCount} of {result.TotalCount} {UnitNounPlural} failed).";
+        return CreateException(message, result);
+    }
+
+    private async Task CheckpointParentResultAsync(
+        BatchResult<T> result,
+        CompletionReason completionReason,
+        DurableExecutionException? failureException,
```
normj: The `failureException` parameter is not used in this private method. Should it be removed or is it a miss?

normj (commented):
Approved with one nit

philasmar (approved):

Libraries/src/Amazon.Lambda.DurableExecution/Internal/ConcurrentOperation.cs
GarrettBeatty: it should be removed - forgot to remove it

normj (approved):

PR #2410, opened 2026-06-08
Integrate [DurableExecution] with the Annotations source generator

GarrettBeatty:
Add a [DurableExecution] attribute (in Amazon.Lambda.Annotations) and source generator support so a method annotated with [LambdaFunction] + [DurableExecution] generates:
- a typed-envelope handler wrapper delegating to DurableFunction.WrapAsync<TInput[,TOutput]>
- a DurableConfig block on the function resource in the generated CFN/SAM template
- lambda:CheckpointDurableExecution / lambda:GetDurableExecutionState IAM permissions (per-function inline policy; skipped when an explicit Role is set)

*Issue #, if available:*
https://github.com/aws/aws-lambda-dotnet/issues/2418 
*Description of changes:*


Libraries/test/TestCustomAuthorizerApp/TestCustomAuthorizerApp.csproj
```diff
@@ -7,6 +7,9 @@
     <AWSProjectType>Lambda</AWSProjectType>
     <!-- This property makes the build directory similar to a publish directory and helps the AWS .NET Lambda Mock Test Tool find project dependencies. -->
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+    <!-- src\Function is a separate, self-contained project. Exclude its subtree so this project's
+         default globs don't pick up its sources or generated obj\AssemblyInfo (which causes CS0579). -->
+    <DefaultItemExcludes>$(DefaultItemExcludes);src\**</DefaultItemExcludes>
```
GarrettBeatty, line 12: was running into build issue so fixed this too

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/TestFunctions/AnnotationsClassLibraryFunction/Function.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace AnnotationsClassLibraryFunction;
+
+public class Function
+{
+    // The generator turns this into a wrapper class Function_Workflow_Generated whose Workflow method
+    // delegates to DurableFunction.WrapAsync<TestEvent, TestResult>(Workflow, __request__, __context__) —
+    // the exact entry point the hand-written and executable test functions call, so the
+    // checkpoint/replay/history behavior is identical.
+    [LambdaFunction]
+    [DurableExecution]
```
GarrettBeatty, line 25: i manually deployed this too and also played around with updating the attributes and checking that they reflected in the template too

Libraries/src/Amazon.Lambda.Annotations.SourceGenerator/Validation/LambdaFunctionValidator.cs
```diff
@@ -72,10 +72,84 @@ internal static bool ValidateFunction(GeneratorExecutionContext context, IMethod
...
+            // Durable functions support BOTH programming models on the managed runtime:
+            //  - executable: the generated wrapper delegates to DurableFunction.WrapAsync, which reads the
+            //    serializer off the ILambdaContext populated by the bootstrap loop the executable hosts.
+            //  - class library: the managed runtime hosts its own bootstrap, resolves [assembly: LambdaSerializer],
+            //    and populates ILambdaContext.Serializer the same way, so WrapAsync finds the serializer there too.
+            // Either way the wrapper is identical; only the deployed Handler string differs (assembly name vs.
+            // Assembly::Type::Method), which LambdaFunctionModel.Handler already derives from IsExecutable. So no
+            // OutputKind gate is needed here.
+
+            // Image packaging strips Handler/Runtime from the function resource, which the durable
```
normj: I don't understand this comment and why image functions don't work. Especially since that is how we tell customers to test the preview. Maybe a quick sync up between us to help me understand what is causing the issue.

GarrettBeatty: Might be AI slop. I also asked Claude about this, I think he mentioned the other annotations have the same limitation. I'll double check

GarrettBeatty: fixed in https://github.com/aws/aws-lambda-dotnet/pull/2410/commits/3bee3f57e01dcee9b356a5bb59720ca511bbf164. claude incorrectly assumed some things

philasmar (commented):
1. Resource: `"*"` on the IAM policy — The inline checkpoint policy grants `lambda:CheckpointDurableExecution` and `lambda:GetDurableExecutionState` on `"*"`. The comment says this is a known preview limitation pending a scopable ARN format. This is fine for preview, but should be tracked as a follow-up to scope down once the service supports it. Consider adding a TODO comment or linking an issue so this doesn't get forgotten post-GA.
  2. Diagnostic ID gap — The `AnalyzerReleases.Unshipped.md` lists `AWSLambda0142-0144`, but the design doc mentions `AWSLambda0141` was intended for a Zip-packaging validation. Either allocate 0141 for the Zip check (the validation is mentioned in the PR description) or document why it was skipped to avoid confusion if someone later tries to use that ID.
  3. `IsDurableCheckpointStatement` uses string serialization — The helper `IsDurableCheckpointStatement` serializes the policy object to JSON and checks for a substring match on the Sid. This works but is fragile if the Sid string ever appears in a comment or different context. A more robust approach would be to try casting the object to `Dictionary<string, object>` and checking the structure directly. Low risk in practice but worth noting.
  4. Design doc checked in — `docs/design/annotations-integration-plan.md` is a superseded planning artifact with outdated information (e.g., references to
  `AWSLambda0140` that was never allocated, `"executable-only"` gating that was removed). Consider removing it or moving it to a wiki/issue to avoid confusion for future contributors reading the source tree.

normj (approved):

GarrettBeatty:
>Diagnostic ID gap — The AnalyzerReleases.Unshipped.md lists AWSLambda0142-0144, but the design doc mentions AWSLambda0141 was intended for a Zip-packaging validation. Either allocate 0141 for the Zip check (the validation is mentioned in the PR description) or document why it was skipped to avoid confusion if someone later tries to use that ID.
IsDurableCheckpointStatement uses string serialization — The helper IsDurableCheckpointStatement serializes the policy object to JSON and checks for a substring match on the Sid. This works but is fragile if the Sid string ever appears in a comment or different context. A more robust approach would be to try casting the object to Dictionary<string, object> and checking the structure directly. Low risk in practice but worth noting.

i dont think these are actual issues

GarrettBeatty:
>Resource: "*" on the IAM policy — The inline checkpoint policy grants lambda:CheckpointDurableExecution and lambda:GetDurableExecutionState on "*". The comment says this is a known preview limitation pending a scopable ARN format. This is fine for preview, but should be tracked as a follow-up to scope down once the service supports it. Consider adding a TODO comment or linking an issue so this doesn't get forgotten post-GA.

fixed in latest commit

philasmar (approved):

PR #2414, opened 2026-06-10
cancellation token

GarrettBeatty:
*Issue #, if available:*

*Description of changes:*

Threads a `CancellationToken` into every user `Func` accepted by `IDurableContext` so step bodies can observe both upstream caller cancel intent and SDK-driven workflow teardown.

## What changed

The four `IDurableContext` methods that take a user `Func` now pass a `CancellationToken` into that `Func`:

- `StepAsync` (both overloads): `Func<IStepContext, CancellationToken, Task<T>>`
- `RunInChildContextAsync` (both overloads): `Func<IDurableContext, CancellationToken, Task<T>>`
- `WaitForCallbackAsync` submitter: `Func<string, IWaitForCallbackContext, CancellationToken, Task>`
- `WaitForConditionAsync` check: `Func<TState, IConditionCheckContext, CancellationToken, Task<TState>>`

`WaitAsync`, `CreateCallbackAsync`, and `InvokeAsync` are unchanged — they take no user `Func`.

The token is a linked source combining:
1. The `CancellationToken` parameter passed to the `IDurableContext` method (the caller's intent).
2. A new SDK-owned workflow-shutdown signal (`WorkflowCancellation`) that fires when `TerminationManager` resolves — i.e. when the workflow is being torn down for any reason (wait scheduled, callback pending, retry scheduled, invoke pending, checkpoint failed).

## Semantics

- An `OperationCanceledException` thrown out of a step body **via the linked token** (`linked.IsCancellationRequested == true`) is treated as cancellation: no FAIL checkpoint is written, no retry is consulted, the exception propagates up. The `when (linked.IsCancellationRequested)` filter distinguishes our cancel from a stray user-thrown OCE.
- An `OperationCanceledException` thrown by user code for **unrelated reasons** (token never fired) continues to be treated as a normal step failure: FAIL checkpoint, retry per the configured `RetryStrategy`.
- SDK-internal writes (checkpoint serialization, batcher flush, response build) **do not** observe the workflow token. Successful work is never lost to teardown.
- Cached operation replays short-circuit before the user `Func` is invoked, so a cancelled token cannot affect replay determinism.

## Why now

We're in preview. Landing the breaking delegate-shape change before GA costs every preview user one mechanical lambda update; landing it post-GA would be much more disruptive.

## Public API impact

This is a **breaking** change to the user-facing delegate signatures on `IDurableContext`. Every existing user lambda must add a parameter (or `_`):

```csharp
// Before
await ctx.StepAsync(async step => await DoWork(), name: "work");

// After
await ctx.StepAsync(async (step, ct) => await DoWork(ct), name: "work");
```

Per `.autover/changes/`: bumped as a `Patch` because the package is unreleased preview.

## Tests

- All 318 existing unit tests updated and passing on net8.0.
- New `WorkflowCancellationTests` (10 tests) covering: token primitive, caller-token propagation, termination → cancel, user-OCE-as-failure, no-checkpoint-on-cancel, child-context propagation, condition-check token, replay short-circuit.
- All 34 integration test function projects + the AOT publish test compile cleanly with the new signatures.

## Docs

- XML doc updated on every changed `IDurableContext` `Func` parameter to describe the linked-token contract.
- `docs/core/{steps,child-contexts,callbacks,wait-for-condition}.md` examples updated to take and forward the token.
- New `docs/core/cancellation.md` documenting the contract and user guidance ("don't branch workflow logic on `IsCancellationRequested`", "don't catch and continue", "do pass `ct` to every cancellation-aware API").
- Design doc at `Libraries/src/Amazon.Lambda.DurableExecution/docs/design/cancellation-design.md`.

## Out of scope

- Source generator changes (the `Amazon.Lambda.Annotations` generator's emitted entry point will pass the workflow token into the user's top-level handler — separate change).
- Parallel/map operations (their cancellation rides on the same `WorkflowCancellation`; semantic decisions belong to that design).
- Lambda deadline timer (considered, rejected — Lambda's timeout is the deadline backstop).


normj (approved):
Approved assuming my understand is correct that when this is merged in [ParallelAsync](https://github.com/aws/aws-lambda-dotnet/pull/2375) PR that the `WorkflowCancellation will also be cancelled when we are awaiting the inflight tasks in the finally block.

philasmar (approved):

PR #2428, opened 2026-06-17
Adds Amazon.Lambda.DurableExecution.Testing

GarrettBeatty:
## Summary

Adds `Amazon.Lambda.DurableExecution.Testing` — a NuGet package for testing durable workflows.

Write your test once against `IDurableTestRunner<TInput, TOutput>` and run it either:
- **locally** (`DurableTestRunner`) — in-process on the real runtime engine with an in-memory backend (fast, no AWS), or
- **against a deployed function** (`CloudDurableTestRunner`) — invokes the real Lambda and polls for the result.

Both return the same `TestResult<TOutput>` for assertions (status, result, error, and per-step inspection via `GetStep`/`GetResult<T>`).

Supports time-skipping, the callback flow, and sibling-function registration. Adds an internal `IDurableServiceClient` seam to the runtime package (no public API change).

## Example

```csharp
using Amazon.Lambda.DurableExecution.Testing;
using Xunit;

public class OrderWorkflowTests
{
    [Fact]
    public async Task MultiStepWorkflow_Succeeds()
    {
        await using var runner = new DurableTestRunner<int, int>(
            handler: async (input, ctx) =>
            {
                var doubled = await ctx.StepAsync(async (_, _) => input * 2, name: "double");
                return await ctx.StepAsync(async (_, _) => doubled + 10, name: "add_ten");
            });

        var result = await runner.RunAsync(5);

        result.EnsureSucceeded();
        Assert.Equal(20, result.Result);

        // Inspect individual steps
        var step = result.GetStep("double");
        Assert.Equal(OperationStatus.Succeeded, step.Status);
        Assert.Equal(10, step.GetResult<int>());
    }
}
```

The same assertions run unchanged against a deployed function by swapping the runner:

```csharp
await using var runner = new CloudDurableTestRunner<int, int>(
    functionArn: "arn:aws:lambda:us-east-1:123456789012:function:order-processor:live");

var result = await runner.RunAsync(5);
result.EnsureSucceeded();
```

Day-long `WaitAsync` / callback workflows complete in milliseconds locally (time-skipping is on by default), and `ctx.InvokeAsync` siblings resolve in-process via `RegisterFunction` / `RegisterDurableFunction`. See the package README for the callback and sibling-invoke patterns.

## Test plan

- [x] 106 unit tests passing on net8.0 and net10.0
- [x] Portable scenarios verified end-to-end against real deployed functions
- [x] Existing runtime tests unaffected


Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/TestFunctions/ChildContextFailsFunction/Function.cs
```diff
@@ -21,28 +27,6 @@ public static async Task Main(string[] args)
...
-
-                throw new InvalidOperationException("intentional child context failure for integration test");
-            },
-            name: "phase",
-            config: new ChildContextConfig { SubType = "OrderProcessing" });
-
-        return new TestResult { Status = "should_not_reach" };
-    }
+        => DurableFunction.WrapAsync<StepsRequest, StepsResult>(
+            ChildContextWorkflows.FailsAsync, input, context);
```
GarrettBeatty, line 31: moved all of these common workflows to a different re-usable file so that the local and cloud runner tests can also re-use the existing integ tests

GarrettBeatty:
closing this to separate it into two prs to make release easier

Issue #2430, opened 2026-06-21
dotnet:8.v88 runtime breaks functions bundling Amazon.Lambda.Core < 3.0.0 (ReflectionTypeLoadException for ILambdaResponseStream)


jonathanbcsouza:
### Describe the bug

The `dotnet:8.v88` managed runtime ships `Amazon.Lambda.RuntimeSupport` 2.1.1.0, which contains `ImplLambdaResponseStream : ILambdaResponseStream` in `ResponseStreamLambdaCoreInitializerIsolated.cs`. When a function bundles `Amazon.Lambda.Core` < 3.0.0 (which `dotnet publish` does by default), the DLL in `/var/task` takes assembly resolution precedence over the runtime's copy in `/var/runtime`. The older DLL lacks `ILambdaResponseStream`.

Any code that calls `Assembly.GetTypes()` on the `Amazon.Lambda.RuntimeSupport` assembly (standard in ASP.NET Core DI, AutoMapper, MediatR) throws `ReflectionTypeLoadException` at INIT. The function never starts.

`GetTypes()` fails on the `Amazon.Lambda.RuntimeSupport` assembly (not `Amazon.Lambda.Core`) because `ImplLambdaResponseStream` implements `ILambdaResponseStream`. .NET tries to resolve the interface from the loaded `Amazon.Lambda.Core` (the older copy from `/var/task`), which doesn't have it.

The existing `try/catch` in `LambdaBootstrap.cs:244` protects `InitializeCore()` but does not protect downstream code scanning RuntimeSupport's types.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

Functions bundling older `Amazon.Lambda.Core` (< 3.0.0) should continue to initialize successfully on `dotnet:8.v88`, as they did on `dotnet:8.v84`. The runtime should be backward compatible with existing deployment packages.

### Current Behavior

Functions fail at INIT with:

```
INIT_START Runtime Version: dotnet:8.v88
INIT_REPORT Init Duration: 443.99 ms  Phase: init  Status: error  Error Type: Runtime.Unknown

System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
---> System.Reflection.ReflectionTypeLoadException: Unable to load one or more of the requested types.
Could not load type 'Amazon.Lambda.Core.ResponseStreaming.ILambdaResponseStream'
[...]

PR #2431, opened 2026-06-22
Fix ReflectionTypeLoadException for ILambdaResponseStream when bundli…

jonathanbcsouza:
*Issue #, if available:*
#2430

*Description of changes:*

The `dotnet:8.v88` managed runtime ships `Amazon.Lambda.RuntimeSupport` 2.1.1.0 which contains `ImplLambdaResponseStream : ILambdaResponseStream` in `ResponseStreamLambdaCoreInitializerIsolated.cs`. When a function bundles `Amazon.Lambda.Core` < 3.0.0 (which `dotnet publish` does by default), the older DLL in `/var/task` takes assembly resolution precedence over the runtime's copy in `/var/runtime`. Any code calling `GetTypes()` on the RuntimeSupport assembly triggers `ReflectionTypeLoadException` because .NET cannot resolve `ILambdaResponseStream` from the loaded older Core.

This breaks all .NET 8 Zip Lambda functions on Auto runtime management that use ASP.NET Core hosting, AutoMapper, MediatR, or any DI framework that scans types at startup. May also affect `dotnet:10.v36` and any any .NET managed runtime that shipped the response streaming RuntimeSupport.

**Fix:** Replace the direct `ILambdaResponseStream` interface implementation (`ImplLambdaResponseStream`) with a `DispatchProxy`-based dynamic implementation. No type in the RuntimeSupport assembly directly references `ILambdaResponseStream` at compile time, so `GetTypes()` no longer triggers the type load. The proxy is only created at runtime after confirming the interface exists in the loaded `Amazon.Lambda.Core`.

**Changes:**
- `ResponseStreamLambdaCoreInitializerIsolated.cs`: Removed `ImplLambdaResponseStream` nested class. Added reflection-based initialization with `DispatchProxy` to dynamically implement `ILambdaResponseStream` without a static type reference.
- `ResponseStreamProxy`: New `DispatchProxy` subclass that forwards interface calls to `ResponseStream`.
- `TestImplLambdaResponseStream.cs`: Test-only helper replacing the removed production class in unit tests.
- `LambdaResponseStreamingCoreTests.cs`: Updated to use `TestImplLambdaResponseStream`.

**Testing:**
- 284/286 unit tests pass locally (2 failures are `NativeAOTTests` requiring .NET 9/10 SDK, unrelated to this change)
- Reproduced the original issue and confirmed the fix concept in a test account
- End-to-end response streaming via DispatchProxy not verified locally (requires deploying to `/var/runtime/` which is immutable outside the Lambda team's build pipeline)

### DIFF
--- a/.autover/changes/5d462e9b-d94a-4005-9b7c-ddc43b5ab0d9.json
+++ b/.autover/changes/5d462e9b-d94a-4005-9b7c-ddc43b5ab0d9.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.RuntimeSupport",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Fixed ReflectionTypeLoadException when customer bundles Amazon.Lambda.Core \u003C 3.0.0 and runtime scans types. Replaced direct ILambdaResponseStream implementation with DispatchProxy to avoid exposing the type in assembly metadata."
+      ]
+    }
+  ]
+}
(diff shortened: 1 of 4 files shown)


normj:
Closing because this solution causes Native AOT issues. I have a new PR out that solves the problem in a different way that doesn't require any new reflection.

PR #2451, opened 2026-06-26
Speed up and stabilize integration tests (parallelism + deploy retry)

GarrettBeatty:
## Summary
TLDR: make tests run in parallel and fix flaky tests. was taking 60 minutes to run all tests now it takes 30

Started as a fix for a flaky CI failure in `TestCustomAuthorizerApp.IntegrationTests`, then grew into a broader effort to speed up and stabilize the integration-test phase (which dominated CI wall-clock by running everything serially), plus fixes for a few flaky unit tests surfaced along the way.

## Reliability

- **Retry custom-authorizer deployment on transient IAM role propagation.** The fixture's CloudFormation deploy intermittently rolled back with *"The role defined for the function cannot be assumed by Lambda"* — a transient IAM eventual-consistency race. `DeploymentScript.ps1` now retries the deploy (deleting the rolled-back stack between attempts, since `ROLLBACK_COMPLETE` can't be re-created) and surfaces CloudFormation failure events.

## Speed

- **Run the integration-test projects in parallel.** `run-integ-tests` now runs each `*.IntegrationTests.csproj` concurrently (`run-integ-tests-parallel.ps1`); each project deploys its own isolated stack, so they share no state.
- **Build the test projects once, up front.** The parallel runner builds all integration projects serially first, then runs `dotnet test --no-build` in parallel — so the concurrent runs don't each rebuild the shared `IntegrationTests.Helpers` project and race on its build output.
- **Stack-scoped Lambda lookup.** `LambdaHelper.FilterByCloudFormationStackAsync` uses CloudFormation `ListStackResources` instead of scanning every Lambda in the account and reading each function's tags — O(stack) instead of O(account), and no shared-account throttling.
- **In-project test parallelism.** `TestServerlessApp` and `TestCustomAuthorizerApp` share their single deployed-stack fixture across the assembly via `IAssemblyFixture` instead of one serial `[Collection]`, so the test classes run in parallel (stack still deploys once).
- **Durable suite parallelism.** Enabled parallel execution for the durable integ suite (was `maxParallelThreads=1`).
- **Publish durable test functions once, in a single MSBuild pass.** A generated traversal project (`Restore;Publish`, `BuildInParallel`) builds the shared dependency projects once and publishes every function to its own `bin/publish`; tests then only zip the output — replacing per-test cold publishing.

## Making durable parallelism safe (rate limits & races)

Enabling parallelism in the durable suite surfaced a series of shared-resource contention issues, fixed in layers:

- **IAM throttling** → replaced per-test IAM roles with a single shared execution role (created at most once per account, reused across runs); dispose no longer deletes roles.
- **Lambda control-plane throttling** → share the AWS clients statically so adaptive retry coordinates backoff across deployments, and cap concurrent control-plane calls (`CreateFunction`/`DeleteFunction`/`GetFunctionConfiguration`) with a suite-wide gate.
- **Shared-file races** → idempotent `dotnet tool install` across the parallel deploy scripts, and zip each function package to a unique temp path (a function used by more than one test was being zipped to a shared path concurrently).

## Developer experience

- **Live integ-test output.** The parallel runner streams each project's output line-by-line (prefixed with the project name) instead of buffering until completion; failed projects get a clean reprinted block.

## Flaky unit-test fixes (unrelated to the integ work, surfaced in CI)

- **Durable suspend tests** (`InvokeOperationTests` et al.): replaced fixed `Task.Delay` waits before asserting suspension with a deterministic await on the termination signal (`TerminationManager.TerminationTask`), bounded by a timeout — the fixed delays raced under CI thread-pool pressure.
- **`FileDescriptorLogStream` test**: the test helper trimmed trailing null bytes from captured output, which flaked ~1/256 of the time when a log header's timestamp ended in `0x00` (16-byte header read as 15). Now captures exactly the bytes written.
- **Streaming E2E test** (`StreamingE2EWithMoq`): `ResponseStreamFactory` tracks the active invocation in a static field (on-demand) or an `AsyncLocal` (multi-concurrency), and `GetCurrentContext()` prefers the `AsyncLocal`. Several factory tests set the `AsyncLocal` synchronously on reused xUnit worker threads, leaking into a later on-demand test so its response silently fell back to the buffered path (`CapturedHttpBytes` null). Fixed test-only (no shipping code changed): the multi-concurrency tests now write the `AsyncLocal` on isolated `Task.Run` flows so the mutation can't leak across threads.

## Testing

- Affected projects build clean; PowerShell scripts parse.
- Verified against AWS where the local environment allowed: custom-authorizer deploys its stack once and all 20 tests pass under the parallel `IAssemblyFixture` setup; the shared-role + single-pass publish path works (51/51 functions publish in one MSBuild pass).
- Flaky unit-test fixes verified by stress-running: the streaming test failed ~40% of full-assembly runs before the fix and 12/12 after; the durable suspend and log-stream tests are now deterministic.
- Remaining end-to-end durable-suite throttling/timing has been validated iteratively on real CI (each contention fix above was driven by a CI run).

### DIFF
--- a/Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/DurableFunctionDeployment.cs
+++ b/Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/DurableFunctionDeployment.cs
@@ -9,6 +9,7 @@ using Amazon.IdentityManagement;
 using Amazon.IdentityManagement.Model;
 using Amazon.Lambda;
 using Amazon.Lambda.Model;
+using Amazon.Runtime;
 using Xunit.Abstractions;
 
 namespace Amazon.Lambda.DurableExecution.IntegrationTests;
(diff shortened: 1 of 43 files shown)


Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/DurableFunctionDeployment.cs
```diff
@@ -36,24 +37,54 @@ internal sealed class DurableFunctionDeployment : IAsyncDisposable
...
+    // coordinate backoff across the parallel deployments.
+    private static readonly IAmazonLambda _lambdaClient = new AmazonLambdaClient(BuildClientConfig<AmazonLambdaConfig>());
+    private static readonly IAmazonIdentityManagementService _iamClient =
+        new AmazonIdentityManagementServiceClient(BuildClientConfig<AmazonIdentityManagementServiceConfig>());
+
+    // Lambda control-plane calls (CreateFunction/DeleteFunction/GetFunctionConfiguration) are
+    // account-rate-limited and are the next bottleneck once IAM is no longer per-test. Cap how many
+    // run concurrently across the whole suite so the parallel deployments don't collectively exceed
+    // Lambda's limits; data-plane calls (Invoke, durable-execution reads) are not gated.
+    private static readonly SemaphoreSlim LambdaControlPlaneGate = new(2, 2);
```
GarrettBeatty, line 55: even though it runs in parallel i still throttle it a bit in order to not hit rate limiting

philasmar (approved):

normj (approved):

PR #2453, opened 2026-07-01
Require ExecutionTimeout on [DurableExecution] and fix empty DurableConfig

GarrettBeatty:
## Summary

Makes `ExecutionTimeout` a **required**  https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-durableconfig.html constructor argument on `[DurableExecution]` and fixes the Annotations source generator so it never emits an invalid empty `DurableConfig`. Also adds real `deploy-serverless` → CloudFormation integration coverage for the durable annotations path.

## Why

The Lambda service rejects a durable function whose `DurableConfig` has no `ExecutionTimeout`:

```
InvalidParameterValueException: You cannot create a function with a durable
configuration without an executionTimeout
```

The generator previously emitted an empty `"DurableConfig": {}` when `[DurableExecution]` set neither property (see the old `CloudFormationWriter` fallback). That template passes the SAM transform but is **rejected by CloudFormation's `EarlyValidation` hook at deploy time** — so a `dotnet lambda deploy-serverless` of a plain `[DurableExecution]` function (exactly what the durable blueprint generates) fails. Since a durable function is invalid without an `ExecutionTimeout`, there is no valid "unset" state — it should be required.

This wasn't caught because no test deployed the generator-produced template through CloudFormation: the durable integ suite bypasses CFN by calling `CreateFunction` directly with a hardcoded `ExecutionTimeout`, and the generator unit tests never covered the no-args case.

## Changes

- **`DurableExecutionAttribute`**: `ExecutionTimeout` is now a required constructor argument (`[DurableExecution(executionTimeout: 300)]`); `RetentionPeriodInDays` stays an optional named property. `Validate()` always range-checks `ExecutionTimeout`.
- **`CloudFormationWriter`**: always emits `ExecutionTimeout`; removed the empty-`DurableConfig` fallback.
- **`DurableExecutionAttributeBuilder`**: reads the constructor argument.
- Updated all `[DurableExecution]` usages, unit tests, and the `Amazon.Lambda.DurableExecution` README.
- **New `TestDurableServerlessApp` + `TestDurableServerlessApp.IntegrationTests`**: a durable function deployed via `dotnet lambda deploy-serverless` → CloudFormation (us-east-1, Zip / `dotnet10`), asserting the generated `DurableConfig` template actually deploys. This closes the coverage gap. The now-redundant CFN-bypassing `AnnotationsClassLibraryFunction` test is retired.

## Breaking change

`[DurableExecution]` with no arguments no longer compiles — callers must supply `executionTimeout`. `Amazon.Lambda.Annotations` durable support is preview.

## Testing

- `Amazon.Lambda.Annotations.SourceGenerators.Tests`: **579/579 pass** (net8.0 + net10.0).
- `Amazon.Lambda.DurableExecution.Tests`: **404/404 pass**.
- New `TestDurableServerlessApp.IntegrationTests`: deployed to AWS (account 147997163238, us-east-1) — stack reached `CREATE_COMPLETE` and the deployed function has `DurableConfig.ExecutionTimeout = 300`. **1/1 pass**; resources cleaned up.


.autover/changes/durable-execution-required-timeout.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.Annotations",
+      "Type": "Patch",
```
GarrettBeatty: is patch fine here or should be minor? this is a breaking change making executiontimeout a required parameter

Libraries/test/Amazon.Lambda.DurableExecution.IntegrationTests/TestFunctions/AnnotationsClassLibraryFunction/AnnotationsClassLibraryFunction.csproj
```diff
@@ -1,22 +0,0 @@
-<Project Sdk="Microsoft.NET.Sdk">
-
-  <PropertyGroup>
-    <TargetFramework>net10.0</TargetFramework>
-    <!-- Class-library model: NO OutputType=Exe and NO GenerateMain. The managed dotnet10 runtime hosts
```
GarrettBeatty, line 5: this integration test wasnt actually working as intended before (it wasnt using the serverless.template) it was manually deploying still.


so i made a new test project similar to TestServerlessApp which actually uses deploymentscript to deploy serverless

Libraries/test/TestDurableServerlessApp.IntegrationTests/DurableAnnotationsTest.cs
```diff
@@ -0,0 +1,50 @@
...
+        [Fact]
+        public async Task DurableServerless_DeploysWithDurableConfig()
+        {
+            // The fixture already asserted the CloudFormation stack reached CREATE_COMPLETE and discovered a
+            // single Lambda function, so reaching here proves the generated DurableConfig template deployed.
+            Assert.False(string.IsNullOrEmpty(_fixture.DurableFunctionName), "A durable function should have been deployed.");
+
+            // Confirm the deployed function actually carries the durable configuration the generator emitted.
+            // ExecutionTimeout is required — an empty DurableConfig would have failed the deploy above.
+            var config = await _fixture.LambdaClient.GetFunctionConfigurationAsync(
```
GarrettBeatty, line 41: i dont actually execute the durable function here. i just valdiate that it gets deployed correctly and validate the config is correct. i figured there is already other integration tests (not in annotation library) which validate the execution history etc

.autover/changes/durable-execution-required-timeout.json
philasmar: Should be a Minor

philasmar (approved):
Approving assuming you make this a "Minor" change

normj (approved):

PR #2462, opened 2026-07-07
Set AutoPublishAlias on serverless.DurableFunction blueprint

GarrettBeatty:
## Description

Adds `AutoPublishAlias: live` to the function resource in the `serverless.DurableFunction` blueprint (vs2026).

## Motivation

When deploying the durable serverless blueprint as-is and then invoking the function, the invocation fails with a **no version published** error. Durable execution runs against a specific published function version, but the template published none.

With `AutoPublishAlias: live`, SAM publishes a new function version and points the `live` alias at it on every deploy, so the function is immediately invokable after `dotnet lambda deploy-serverless`.

## Changes

- `serverless.template`: add `"AutoPublishAlias": "live"` to the function `Properties`. This is a user-authored property that the Annotations source generator preserves across re-sync (it only manages `DurableConfig` / durable policy tokens).
- `Readme.md`: document the alias behavior in the Deploy section.
- Added AutoVer changefile (`Amazon.Lambda.Templates`, Patch).

## Notes

- Targets `dev`.
- Only the serverless blueprint is affected; the class-library `lambda.DurableFunction` blueprint has no `serverless.template` (deploys via `deploy-function`).

### DIFF
--- a/.autover/changes/durable-serverless-autopublishalias.json
+++ b/.autover/changes/durable-serverless-autopublishalias.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.Templates",
+      "Type": "Patch",
+      "ChangelogMessages": [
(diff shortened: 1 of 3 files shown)


GarrettBeatty:
"live" is what most of the AWS docs use in examples for the version which is why I chose it

normj (approved):

philasmar (approved):

PR #2492, opened 2026-07-22
Fix Test Tool v2 Blazor UI non-interactive on .NET 10 (static assets 404)

GarrettBeatty:
## Symptom

The AWS Lambda Test Tool v2 Blazor web UI had two static-asset serving bugs that broke interactivity, each specific to a target framework:

**(A) Non-interactive on .NET 10 (any .NET 9+).** Buttons don't respond, `@onclick`/`@bind` never fire; `window.Blazor` is `undefined`. The tell is subtle — the framework script returns HTTP **200 with a 0-byte body**, so status-code checks alone look green while the browser downloads nothing:

```
# net10, before this fix
_framework/blazor.web.js          -> HTTP 200, 0 bytes      <-- EMPTY
Amazon.Lambda.TestTool.styles.css -> HTTP 200, 0 bytes      <-- EMPTY
```

**(B) Code editor dead on .NET 8 (`dotnet run`).** The Razor class library assets for the Monaco code editor (`_content/BlazorMonaco/**`) return **404**, so the request/response editor never initializes — e.g. selecting an **example request does not populate the input**, and typing/formatting in the editor doesn't work:

```
# net8, dotnet run, before this fix
_framework/blazor.web.js                              -> HTTP 200, 187162 bytes  (ok)
_content/BlazorMonaco/lib/monaco-editor/min/vs/loader.js -> HTTP 404, 0 bytes    <-- MISSING
```

## Root cause

Static-asset serving changed in .NET 9: framework assets (`_framework/*`) and scoped CSS are served by the endpoint-routing **`app.MapStaticAssets()`** pipeline, which reads bytes from `IWebHostEnvironment.WebRootFileProvider`. Under `dotnet run`, neither the framework files nor RCL `_content/**` assets physically live in `wwwroot` — they live in the NuGet cache and are surfaced through the `*.staticwebassets.runtime.json` manifest. Two things kept that manifest from being reachable:

1. **Manifest not composed outside Development.** ASP.NET Core only composes the static-web-assets manifest into `WebRootFileProvider` automatically **when the environment is Development** (`WebHost.ConfigureWebDefaults` calls `StaticWebAssetsLoader.UseStaticWebAssets` only under `IsDevelopment()`). This tool runs as **Production by default**, so the web root was a bare `wwwroot` provider.
   - On **net9+** this meant `MapStaticAssets` served empty (0-byte) 200s for `_framework/*` → symptom (A).
   - On **net8** (no `MapStaticAssets`), the manifest was likewise never composed, and — see below — nothing read it anyway → symptom (B).

2. **Content root followed the current working directory.** As an installed global tool the process launches from an arbitrary cwd; `WebRootFileProvider` defaulted to `cwd/wwwroot`, a nonexistent path, so every asset returned empty 200s.

3. **The classic `UseStaticFiles` used an explicit bare-wwwroot provider.** It was pinned to a `PhysicalFileProvider(wwwroot)` and never read the manifest-composed `WebRootFileProvider`. On net9+ this was masked because `MapStaticAssets` reads `WebRootFileProvider`; on **net8 there is no such reader**, so manifest-mapped `_content/**` assets were unreachable → symptom (B).

## Fix

All in `Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/TestToolProcess.cs`:

- **Pin the content root** to the install dir so `WebRootFileProvider` resolves regardless of launching cwd:
  ```csharp
  var builder = WebApplication.CreateBuilder(new WebApplicationOptions
  {
      ContentRootPath = AppContext.BaseDirectory
  });
  ```
- **Compose the static-web-assets manifest on all target frameworks** (previously net9+-only). ASP.NET Core only does this automatically in Development; the tool runs as Production:
  ```csharp
  builder.WebHost.UseStaticWebAssets();   // now unconditional
  ```
- **Serve `_framework/*` + scoped CSS via the endpoints manifest on net9+:**
  ```csharp
  #if NET9_0_OR_GREATER
  app.MapStaticAssets();   // guarded: only when the manifest exists (test host has none)
  #endif
  ```
- **On .NET 8, add a second `UseStaticFiles` over `WebRootFileProvider`** so manifest-mapped RCL `_content/**` assets (BlazorMonaco) resolve. There is no `MapStaticAssets` on net8, and the explicit wwwroot provider can't see the composed manifest:
  ```csharp
  #if !NET9_0_OR_GREATER
  if (app.Environment.WebRootFileProvider is not null)
  {
      app.UseStaticFiles(new StaticFileOptions
      {
          FileProvider = app.Environment.WebRootFileProvider
      });
  }
  #endif
  ```

The existing wwwroot `UseStaticFiles` provider is preserved (authoritative for the installed tool, where all assets are published directly into `wwwroot`). For an installed tool the second net8 lookup resolves to the same `wwwroot`, so it is a harmless second pass.

## Verification — full 8-case matrix

Probed the three assets whose absence breaks the UI — `_framework/blazor.web.js` (interactive circuit), `_content/BlazorMonaco/.../loader.js` (Monaco editor → example-request fill), and `app.css` — asserting HTTP 200 **with non-zero bytes** across every combination of **{.NET 8, .NET 10} × {Production, Development} × {`dotnet run`, installed tool}**:

| Case | blazor.web.js | monaco loader.js | app.css | Result |
|---|---|---|---|---|
| run / net8.0 / Production | 200, 187162 | 200, 30061 | 200, 3946 | ✅ PASS |
| run / net8.0 / Development | 200, 187162 | 200, 30061 | 200, 3946 | ✅ PASS |
| run / net10.0 / Production | 200, 200575 | 200, 30061 | 200, 3946 | ✅ PASS |
| run / net10.0 / Development | 200, 200575 | 200, 30061 | 200, 3946 | ✅ PASS |
| installed / net8.0 / Production | 200, 187162 | 200, 30061 | 200, 3946 | ✅ PASS |
| installed / net8.0 / Development | 200, 187162 | 200, 30061 | 200, 3946 | ✅ PASS |
| installed / net10.0 / Production | 200, 200575 | 200, 30061 | 200, 3946 | ✅ PASS |
| installed / net10.0 / Development | 200, 200575 | 200, 30061 | 200, 3946 | ✅ PASS |

(Before this change: `run/net8.0/*` had `monaco loader.js -> 404` and `run/net10.0/*` had `blazor.web.js -> 200, 0 bytes`.) Real-browser confirmation via headless Chromium: `window.Blazor=true` and no console errors on `/`, `/documentation`, and `/durable-execution`. Both TFMs build with 0 warnings / 0 errors; the Test Tool unit tests (which invoke `Startup` directly) pass on net8.

### DIFF
--- a/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/TestToolProcess.cs
+++ b/Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/TestToolProcess.cs
@@ -36,9 +36,40 @@ public class TestToolProcess
     /// </summary>
     public static TestToolProcess Startup(RunCommandSettings settings, CancellationToken cancellationToken = default)
     {
-        var builder = WebApplication.CreateBuilder();
+        var builder = WebApplication.CreateBuilder(new WebApplicationOptions
+        {
+            // Pin the content root to the tool's install directory rather than the current
+            // working directory. As an installed global tool the process is launched from an
+            // arbitrary cwd, and the default content root is the cwd. On .NET 9+ the framework
+            // and scoped-CSS assets are served by MapStaticAssets from the WebRootFileProvider,
+            // which is derived from the content root (contentRoot/wwwroot). If the content root
+            // is a foreign cwd, that provider points at a nonexistent wwwroot and MapStaticAssets
+            // returns empty (0-byte) 200 responses, leaving the Blazor UI non-interactive.
+            ContentRootPath = AppContext.BaseDirectory
+        });
 
         Utils.ConfigureWebApplicationBuilder(builder);
+
+        // Static web assets (the *.staticwebassets.runtime.json manifest) map two kinds of files
+        // that do NOT physically live under wwwroot when running from a build output (dotnet run /
+        // dotnet build): the Blazor framework files (_framework/blazor.web.js) and Razor class
+        // library content such as BlazorMonaco's _content/BlazorMonaco/** editor assets. Both live
+        // in the NuGet cache and are surfaced via that manifest. ASP.NET Core only composes the
+        // manifest into the WebRootFileProvider automatically in the Development environment, but
+        // this tool runs in the Production environment by default, so without the call below the
+        // WebRootFileProvider is a bare wwwroot provider and those assets are unreachable.
+        //
+        // On .NET 9+ that leaves MapStaticAssets (added below) unable to find the framework files —
+        // it serves an empty (0-byte) HTTP 200, window.Blazor is never defined, and the whole UI is
+        // non-interactive. On .NET 8 the framework files are served by Blazor's own middleware, but
+        // the RCL _content/** assets are served through UseStaticFiles + WebRootFileProvider, so
+        // without the manifest the Monaco editor assets 404 and the code editor never initializes
+        // (e.g. selecting an example request cannot populate the input). Either way the fix is the
+        // same, so this runs on all target frameworks.
+        //
+        // When running as an installed global tool the manifest is absent (the framework and RCL
+        // content are published directly into wwwroot instead), so this is a harmless no-op there.
+        builder.WebHost.UseStaticWebAssets();
 
         builder.Services.AddSingleton<IRuntimeApiDataStoreManager, RuntimeApiDataStoreManager>();
         builder.Services.AddSingleton<IThemeService, ThemeService>();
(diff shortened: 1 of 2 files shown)


GarrettBeatty:
image
 also ran manually

GarrettBeatty:
i retested a second time now manually here is what i did.

Packed and installed the tool to
"~/.dotnet/tools/.store/amazon.lambda.testtool/0.15.0/amazon.lambda.testtool/0.15.0/tools/<TFM>/any/Amazon.Lambda.TestTool.dll

and then for each TFM (net 8/10), i launched the tool with

dev mode
 ASPNETCORE_ENVIRONMENT=Development; dotnet <path to tool dll>  start

prod mode (if environment is unset the default is prod as per microsfot docs)

unset  ASPNETCORE_ENVIRONMENT; dotnet <path to tool dll>  start

amd then verified in the ui triggering the function it works

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/TestToolProcess.cs
```diff
@@ -91,15 +122,61 @@ public static TestToolProcess Startup(RunCommandSettings settings, CancellationT
...
-        // directory. Without this, non-Production environments attempt to use the static web
-        // assets manifest which contains absolute paths from the build machine and will fail
-        // when running as an installed global tool on a different machine.
+        // Serve classic static files from the tool's install directory (wwwroot). This is the
+        // authoritative provider for an installed global tool, where every asset — app.css, the
+        // Blazor framework files, and RCL _content/** (e.g. BlazorMonaco) — is published directly
+        // into wwwroot. Pinning an explicit provider (rather than the default WebRootFileProvider)
+        // also avoids depending on the static-web-assets manifest, whose absolute build-machine
+        // paths would not resolve on another machine.
         app.UseStaticFiles(new StaticFileOptions
```
normj, line 113: Given the calls to `UseStaticFiles` and `MapStaticAssets` done below is the call still needed?

normj (changes requested):

Tools/LambdaTestTool-v2/src/Amazon.Lambda.TestTool/Processes/TestToolProcess.cs
GarrettBeatty: - **`UseStaticWebAssets()`** — tells the server *where* the framework/Monaco files actually are (the NuGet cache, via the manifest). Without it the server looks only in `wwwroot`, doesn't find them, and
  everything below has nothing to serve. This is the prerequisite for the other two.
  - **`UseStaticFiles(wwwroot)`** — serves the plain files that *do* live in `wwwroot`: `app.css`, images, favicon. Nothing else serves these, so without it the page loads unstyled.
  - **`MapStaticAssets()` (net9+)** — on .NET 9+ the framework files (`blazor.web.js`) and scoped CSS are served through this newer pipeline, not the classic one. Without it they come back empty,
  `window.Blazor` never loads, and the UI is dead. (.NET 8 has no `MapStaticAssets`, so a second `UseStaticFiles` over the manifest-composed provider does the same job.)


tried moving some of the code around and updating the comments to be easier to understand. but i tried testing without all of these options and some scenarios did not work

GarrettBeatty: Legend for asset kinds:
  - **classic** = `app.css`, images, favicon (physically in wwwroot)
  - **framework** = `_framework/blazor.web.js` + scoped CSS (needed for interactivity)
  - **RCL** = `_content/BlazorMonaco/**` (needed for the code editor)

  | Framework | Deployment | classic served by | framework served by | RCL (Monaco) served by | Extra thing that makes it work |
  |---|---|---|---|---|---|
  | **net10** | `dotnet run` | `UseStaticFiles(wwwroot)` | `MapStaticAssets()` | `MapStaticAssets()` | `UseStaticWebAssets()` — files are in the NuGet cache, not wwwroot |
  | **net10** | installed tool | `UseStaticFiles(wwwroot)` | `MapStaticAssets()` | `MapStaticAssets()` | `ContentRootPath` pin — tool launches from an arbitrary cwd |
  | **net8** | `dotnet run` | `UseStaticFiles(wwwroot)` | Blazor's own middleware | 2nd `UseStaticFiles(WebRoot)` | `UseStaticWebAssets()` — files are in the NuGet cache, not wwwroot |
  | **net8** | installed tool | `UseStaticFiles(wwwroot)` | Blazor's own middleware | `UseStaticFiles(wwwroot)` | `ContentRootPath` pin — everything's in wwwroot, so the base call covers it |

normj (approved):

dscpinheiro (approved):

PR #2506, opened 2026-07-30
Fix JSON log parameter ordering

### LINKED ISSUE #2468: JSON log format pairs ILogger message template placeholders with state values by position, garbling ASP.NET Core hosting logs
### Describe the bug

When a function uses `LoggingConfig.LogFormat: JSON` and bridges `Microsoft.Extensions.Logging.ILogger` to the Lambda logging API via `Amazon.Lambda.Logging.AspNetCore`, structured log events can come out with their JSON properties holding the **wrong values** (each property shifted to a neighboring placeholder's value).

Root cause, as far as I can tell, is a combination of two behaviors:

1. `LambdaILogger` extracts the message template from `{OriginalFormat}` and collects the state values **in state enumeration order** ([LambdaILogger.cs](https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.Logging.AspNetCore/LambdaILogger.cs)).
2. `JsonLogMessageFormatter` then pairs template placeholders with arguments **strictly by position** ([JsonLogMessageFormatter.cs](https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/Logging/JsonLogMessageFormatter.cs)).

The `Microsoft.Extensions.Logging` contract is that placeholders are resolved **by name** against the state key/value pairs; nothing guarantees that state enumeration order matches placeholder order. The most prominent violator ships in ASP.NET Core itself: [`HostingRequestStartingLog`](https://github.com/dotnet/aspnetcore/blob/main/src/Hosting/Hosting/src/Internal/HostingRequestStartingLog.cs) enumerates its state as `Protocol, Method, ContentType, ContentLength, Scheme, Host, PathBase, Path, QueryString`, while its template is:

```
Request starting {Protocol} {Method} {Scheme}://{Host}{PathBase}{Path}{QueryString} - {ContentType} {ContentLength}
```

So every `Request starting` log line of every ASP.NET Core app running on Lambda with JSON log format is silently garbled from the third placeholder on.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Template placeholders resolved by name, as with every other `Microsoft.Extensions.Logging` sink:

```json
{
    "message": "Request starting HTTP/1.1 DELETE https://api.example.com/events/123?preserve=true - application/json 0",
    "Method": "DELETE",
    "Scheme": "https",
    "Host": "api.example.com",
    "Path": "/events/123",
    "QueryString": "?preserve=true",
    "ContentType": "application/json",
    "ContentLength": 0
}
```

### Current Behavior

Actual CloudWatch log record (values anonymized). Note `Scheme` holding the content type, `Host` holding the content length, `Path` holding the host, `ContentType` holding the path, etc.:

```json
{
    "timestamp": "2026-07-04T22:43:57.529Z",
    "level": "Information",
    "requestId": "188cbe50-087d-4465-acf4-d4a6067ef0da",
    "message": "Request starting {Protocol} DELETE application/json://0httpsapi.example.com - /events/123 ?preserve=true",
    "Method": "DELETE",
    "Scheme": "application/json",
    "Host": 0,
    "PathBase": "https",
    "Path": "api.example.com",
    "QueryString": "",
    "ContentType": "/events/123",
    "ContentLength": "?preserve=true"
}
```

Each placeholder received the state value at the same index instead of the value with the matching name (`{Scheme}` ← `ContentType`, `{Host}` ← `ContentLength`, `{PathBase}` ← `Scheme`, ...). The rendered `message` is garbled the same way.

### Reproduction Steps

Minimal ASP.NET Core app:

```csharp
var builder = WebApplication.CreateBuilder(args);
builder.Services.AddAWSLambdaHosting(LambdaEventSource.RestApi);
builder.Logging.ClearProviders();
builder.Logging.AddLambdaLogger();
var app = builder.Build();
app.MapGet("/", () => "ok");
app.Run();
```

Deploy behind API Gateway with:

```yaml
LoggingConfig:
  LogFormat: JSON
```

Invoke any endpoint and inspect the `Request starting` record in CloudWatch: the `Scheme`, `Host`, `PathBase`, `Path`, `QueryString`, `ContentType` and `ContentLength` properties are shifted as shown above.

### Possible Solution

Either:

- In `LambdaILogger` (JSON branch): parse the placeholder names from `{OriginalFormat}` and emit the arguments **in placeholder order**, resolving each one by key against the state; or
- In `JsonLogMessageFormatter`: when arguments come from an `ILogger` state, match them to placeholders by name instead of by position.

### Additional Information/Context

Any `ILogger` event whose state enumeration order differs from its template placeholder order is affected — ASP.NET Core hosting logs are just the guaranteed-out-of-the-box case. Logs written through the standard `LoggerExtensions`/`LoggerMessage` paths happen to be unaffected because `FormattedLogValues` keeps both orders aligned, which is probably why this went unnoticed.

Amazon.Lambda.AspNetCoreServer 10.1.1
Amazon.Lambda.AspNetCoreServer.Hosting 2.1.0
Amazon.Lambda.Logging.AspNetCore 5.0.0
Amazon.Lambda.Core 3.1.0

### Targeted .NET Platform

.NET 10 (managed `dotnet10` runtime)

Amazon Linux 2023 (Lambda managed runtime)

Pybsama:
*Issue #, if available:*

Fixes #2468

*Description of changes:*

When the JSON logging path receives a custom `ILogger` state whose enumeration
order differs from `{OriginalFormat}`, order the callback arguments by message
template placeholder name instead of by state position.

The implementation:

- preserves missing placeholder slots and handles duplicate placeholder names;
- supports destructuring hints, alignment, format specifiers, escaped braces,
  and positional templates;
- keeps the existing fast path when state is already aligned, and uses the same
  1,024-entry soft cache cap as RuntimeSupport; and
- adds focused regressions for the ASP.NET Core hosting-log shape and the edge
  cases above, plus the required AutoVer patch change file.

Validation:

- `dotnet test Libraries/test/Amazon.Lambda.Logging.AspNetCore.Tests/Amazon.Lambda.Logging.AspNetCore.Tests.csproj --framework net10.0 --configuration Debug` (20 passed)
- `dotnet test Libraries/test/Amazon.Lambda.Logging.AspNetCore.Tests/Amazon.Lambda.Logging.AspNetCore.Tests.csproj --framework net10.0 --configuration Release` (20 passed)
- `dotnet build Libraries/test/Amazon.Lambda.Logging.AspNetCore.Tests/Amazon.Lambda.Logging.AspNetCore.Tests.csproj --configuration Release --framework net8.0` (0 warnings, 0 errors)
- repeated the same checks after applying the commit cleanly to current `master` (`fdb84c71`)
- AutoVer change-file validation and `git -c core.whitespace=cr-at-eol diff --check`

### DIFF
--- a/.autover/changes/58593c38-0c20-4fee-993c-15ea9f1f49fb.json
+++ b/.autover/changes/58593c38-0c20-4fee-993c-15ea9f1f49fb.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.Logging.AspNetCore",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Fix JSON structured logs whose state values use a different order than their message template."
+      ]
+    }
+  ]
+}
--- a/Libraries/src/Amazon.Lambda.Logging.AspNetCore/LambdaILogger.cs
+++ b/Libraries/src/Amazon.Lambda.Logging.AspNetCore/LambdaILogger.cs
@@ -1,10 +1,19 @@
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Globalization;
 
 namespace Microsoft.Extensions.Logging
 {
     internal class LambdaILogger : ILogger
     {
+        private const string ORIGINAL_FORMAT_KEY = "{OriginalFormat}";
+        private const int MESSAGE_TEMPLATE_PARSE_CACHE_MAXSIZE = 1024;
+
+        private static readonly ConcurrentDictionary<string, IReadOnlyList<string>> MESSAGE_TEMPLATE_PARSE_CACHE =
+            new ConcurrentDictionary<string, IReadOnlyList<string>>();
+        private static readonly char[] PROPERTY_NAME_DELIMITERS = { ',', ':' };
+
         // Private fields
         private readonly string _categoryName;
         private readonly LambdaLoggerOptions _options;
@@ -57,25 +66,31 @@ namespace Microsoft.Extensions.Logging
             if (IsLambdaJsonFormatEnabled && state is IEnumerable<KeyValuePair<string, object>> structure)
             {
                 string messageTemplate = null;
-                var parameters = new List<object>();
+                var properties = new List<KeyValuePair<string, object>>();
                 foreach (var property in structure)
                 {
-                    if (property is { Key: "{OriginalFormat}", Value: string value })
+                    if (property is { Key: ORIGINAL_FORMAT_KEY, Value: string value })
                     {
                         messageTemplate = value;
                     }
                     else
                     {
-                        parameters.Add(property.Value);
+                        properties.Add(property);
                     }
                 }
 
+                object[] parameters;
                 if (messageTemplate == null)
                 {
                     messageTemplate = formatter.Invoke(state, exception);
+                    parameters = GetPropertyValues(properties);
+                }
+                else
+                {
+                    parameters = OrderParametersByMessageTemplate(messageTemplate, properties);
                 }
 
-                Amazon.Lambda.Core.LambdaLogger.Log(lambdaLogLevel, exception, messageTemplate, parameters.ToArray());
+                Amazon.Lambda.Core.LambdaLogger.Log(lambdaLogLevel, exception, messageTemplate, parameters);
             }
             else
             {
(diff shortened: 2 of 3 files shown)


madmox:
Reporter of #2468 here. The approach looks right, two practical notes:

- It's still marked as **draft**, so it won't reach the maintainers' review queue.
- It targets `master`. `CONTRIBUTING.md` does say master, but every recently merged PR here (release PRs included) targets `dev`; worth confirming before this stalls.

Also worth knowing: #2474 inserts a block immediately before the `LambdaLogger.Log(...)` call this PR modifies. The two are behaviorally independent, but whichever lands second will need a rebase.

Pybsama:
Thanks for both notes. I checked them against the current repository state: #2545 retired the `dev` branch, so `master` is now the correct single-trunk target. I also applied this commit cleanly to current `master` (`fdb84c71`) and reran the focused suite: 20/20 tests in both Debug and Release on net10.0, plus a net8.0 Release build with 0 warnings and 0 errors.\n\n#2474 is still open. I confirmed that the changes are behaviorally independent but touch the same source/test regions, so whichever lands second will need a conflict resolution. If #2474 lands first, I will rebase this PR and add the combined category-prefix + reordered-parameters regression. Marking this ready for maintainer review now.

GarrettBeatty (commented):
⚠️ This is an automated AI review for a first pass — please check for accuracy.

Reviewed the diff for #2506 (JSON log parameter ordering). 6 findings — no confirmed blockers. Two `major` items are open questions that depend on `Amazon.Lambda.Core.LambdaLogger.Log`, whose body isn't in this diff.

**1. [major] `LambdaILogger.cs` · `OrderParametersByMessageTemplate`** — Appending unconsumed state values and inserting `null` for missing placeholders could alter the observable JSON (extra/renamed fields, or a `Missing: null` field that didn't exist before). Can't confirm from the diff because `LambdaLogger.Log`'s handling of surplus/null args isn't included. Can you confirm the downstream behavior — or whether reordering is the right layer vs. name-based assignment?

**2. [major] `LambdaILogger.cs`** — Duplicate-placeholder handling is asymmetric: when state has more same-name values than the template references, the extra is appended as an *unnamed trailing* value; when the template references more than state provides, the surplus becomes `null`. Neither direction has a test pinning the intended output.

**3. [major] `LambdaILogger.cs` · `UsesPositionalArguments`** — Strips a leading `@` but **not `$`** before `int.TryParse`, so `{$0}` is misclassified as non-positional and falls into the name-based path. No test covers `$`-prefixed positional templates.

**4. [minor] `LambdaILogger.cs` · `ParseTemplateProperties` cache** — `if (Count < MAXSIZE) TryAdd(...)` is check-then-act: under concurrency the cache can exceed 1024 (bounded overflow) and grow with high-cardinality templates. It's a process-lifetime static on a logging hot path.

**5. [minor] `LoggingTests.cs`** — The new facts assert the captured `object[]` parameter array + `MessageTemplate`, not serialized JSON. Correct seam for the reordering logic, but they don't prove observable output — relevant given finding #1.

**6. [minor] `.autover/changes/…json`** — Change type is `Patch`; if finding #1 confirms an observable output change, `Minor` would be more accurate.

PR #2559, opened 2026-09-03
feat(DurableExecution): context-aware serializer engine + Map/Parallel determinism

GarrettBeatty:
## Summary

Adds the **context-aware serializer engine** for durable operation results and extends the fresh-success serializer round-trip (already applied to Step and ChildContext) to **Map/Parallel**, with hardened terminal-failure handling. Stacked on `feature/per-step-serializer` (#2555); `FileSystemSerializer` (#2558) builds on this branch.

## What's in this PR

**1. Context-aware serializer infrastructure**
- Optional `IDurableResultSerializer` with `Serialize`/`Deserialize(…, DurableSerializationContext)`, where the context carries `EntityId` (operation id) + `DurableExecutionArn`.
- `LambdaSerializerHelper` dispatches to the context overload when the configured serializer implements it, otherwise falls back to plain `ILambdaSerializer` with **byte-identical** behavior. No config API change; existing serializers are unaffected.
- Wired at the result choke points: Step, ChildContext, WaitForCondition, and Map/Parallel per-item (per-item `EntityId = OperationId#index` so items don't collide).

**2. Map/Parallel fresh-success round-trip (Flat)**
- On a fresh (non-replay) success, each Flat per-item result is deserialized from its just-serialized checkpoint payload before entering the result — matching what replay reconstructs. This removes a fresh-vs-replay divergence for a non-round-tripping `ItemSerializer`. The returned value is a fresh deserialized instance; overflow (replay-children) results are unaffected.

**3. Map/Parallel inline reuse (Nested)**
- The parent now inlines each Nested unit's **original** SUCCEED-checkpoint payload verbatim, instead of re-serializing the already-round-tripped return value at the parent's per-unit id. This removes a double-transform divergence for non-round-tripping serializers (and, for a context-aware serializer, avoids writing a second copy at a different id).

**4. Terminal round-trip failure handling**
- A fresh-success round-trip **deserialize** failure is now terminal on the Step, ChildContext, and Flat Map/Parallel paths: a FAIL checkpoint is recorded, the retry strategy is **not** consulted, and the already-succeeded side-effecting body is never re-invoked. Replay reconstructs the terminal failure without re-running or re-deserializing.
- A Step post-SUCCEED checkpoint-enqueue failure routes through the same terminal path rather than propagating raw.

## Change files
- `12a4a1f7` — **Major**: Step/Child fresh-success round-trip (reclassified from Minor; breaking at the 1.0.0 baseline).
- `ca5a9844` — **Major**: Map/Parallel Flat per-item fresh-success round-trip.
- `f3a1520c` — **Major**: Map/Parallel Nested inline reuses child payload.
- `d12219eb` — **Major**: terminal round-trip deserialize failure (Child + Flat).
- `3691fc04` — **Major**: terminal Step round-trip failure (no retry).
- `c1cb8e3a` — **Patch**: terminal Step post-SUCCEED enqueue failure.

## Testing
- **439/439** unit tests pass on net8.0; builds clean under `TreatWarningsAsErrors`.

## Note
Two `<see cref="FileSystemSerializer"/>` doc links render as `<c>FileSystemSerializer</c>` here (the type doesn't exist on this branch) and are restored to `cref` links in #2558, which introduces the type.

### DIFF
--- a/.autover/changes/3691fc04-ad66-49c5-a383-f9592a6da0ce.json
+++ b/.autover/changes/3691fc04-ad66-49c5-a383-f9592a6da0ce.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.DurableExecution",
+      "Type": "Major",
+      "ChangelogMessages": [
+        "Behavior change: when a step\u0027s fresh-success serializer round-trip fails (e.g. an asymmetric/broken StepConfig.Serializer that cannot deserialize its own output), the step now fails TERMINALLY (FAIL checkpoint \u002B StepException) instead of being routed through the configured RetryStrategy. Previously a non-null RetryStrategy would re-invoke the already-succeeded, side-effecting step body on each attempt (duplicating side effects) until attempts exhausted. This mirrors ChildContextOperation\u0027s terminal handling of a round-trip failure."
+      ]
+    }
+  ]
+}
--- a/.autover/changes/c1cb8e3a-cca9-415e-b9ca-08289658df08.json
+++ b/.autover/changes/c1cb8e3a-cca9-415e-b9ca-08289658df08.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.DurableExecution",
+      "Type": "Patch",
+      "ChangelogMessages": [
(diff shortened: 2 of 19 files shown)


.autover/changes/3691fc04-ad66-49c5-a383-f9592a6da0ce.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Amazon.Lambda.DurableExecution",
+      "Type": "Major",
```
GarrettBeatty, line 5: since this is a behavior change i put it as major

normj (approved):

PR #2561, opened 2026-09-04
Feature/durable result serializer

GarrettBeatty:
merges already approved prs for per operation serializer and file system serializer


jnunn-aws (commented):
Please update empty PR description.

Libraries/src/Amazon.Lambda.DurableExecution/docs/core/steps.md
```diff
@@ -146,3 +147,50 @@ var result = await ctx.StepAsync(
...
+
+```csharp
+config: new StepConfig
+{
+    // No inner: the durable runtime binds the global serializer as the inner.
+    Serializer = new FileSystemSerializer(basePath: "/mnt/efs/durable")
+}
+```
+
+An explicitly-supplied inner always wins over the global one. The inner is bound only when `FileSystemSerializer` is used through a per-operation serializer slot; used without a bound inner (for example as the assembly-registered serializer) it throws `InvalidOperationException`. `FileSystemStorageMode` (`Always` vs. `Overflow`) controls whether every value is written to a file or only those that would overflow the inline checkpoint; `FileSystemPathEncoding` controls how paths are derived. `FileSystemSerializer` never deletes result files, so pair the base path with an external retention policy (an EFS lifecycle policy, an S3 lifecycle rule, or a scheduled cleanup). This works the same way for any per-operation serializer slot — `CallbackConfig`/`InvokeConfig`/`WaitForConditionConfig<TState>`/`ChildContextConfig.Serializer` and `MapConfig<TItem>`/`ParallelConfig.ItemSerializer`.
```
jnunn-aws, line 196: **`FileSystemSerializer` does not work on Callback/Invoke slots**

This paragraph says the offload mechanism "works the same way for any per-operation serializer slot — `CallbackConfig`/`InvokeConfig`/`WaitForConditionConfig<TState>`/`ChildContextConfig.Serializer` and `MapConfig<TItem>`/`ParallelConfig.ItemSerializer`." That's not true for Callback and Invoke.

`CallbackOperation.cs:214` and `InvokeOperation.cs:138,147` call the serializer through the plain `ILambdaSerializer` methods (`_serializer.Serialize`/`Deserialize`), with no `DurableSerializationContext`. Those methods on `FileSystemSerializer` throw `NotSupportedException` (`FileSystemSerializer.cs:270-279`), and the inner-less variant additionally throws at `RequireInner()` (`287-292`). So a `FileSystemSerializer` in a Callback/Invoke slot fails at runtime.

Offloading may also be a less natural fit for these two: the Invoke payload is read by the callee, and the Callback payload is supplied by the external system, so a local result file has no counterpart there. Could scope the offload docs to the Step/ChildContext/WaitForCondition/Map/Parallel **result** slots, and note that Callback/Invoke support a plain `ILambdaSerializer` override.

Libraries/src/Amazon.Lambda.DurableExecution/FileSystemSerializer.cs
```diff
@@ -0,0 +1,483 @@
...
+    {
+        if (!IsWithinBase(dir))
+            throw new InvalidOperationException(
+                $"FileSystemSerializer: refusing to write an offloaded payload outside the configured " +
+                $"base path. Resolved directory '{dir}' does not resolve under '{_basePath}'. This can " +
+                "happen when the durable execution ARN contains path-traversal segments.");
+    }
+
+    // True when the fully-resolved candidate path lies under the fully-resolved base path.
+    private bool IsWithinBase(string candidate)
```
jnunn-aws, line 393: **Containment check rejects all reads/writes when the base path's leaf is a symlink**

`IsWithinBase` runs both the base and the candidate through `ResolveReal`, but `ResolveReal` only follows symlinks on the leaf and its immediate parent. If `basePath` is itself a symlink (e.g. `/mnt/link -> /mnt/real`, a symlinked EFS mount):

- Base resolves: `fullBase = /mnt/real/`.
- Candidate `/mnt/link/<func>/<exec>/<inv>/<entity>.bin` — its leaf/parent are far below the `/mnt/link` symlink, which `ResolveReal` never touches — stays `/mnt/link/...`.
- `fullPath.StartsWith(fullBase)` is `false`, so `ValidateWriteDirWithinBase` and `ValidatePathWithinBase` reject **every** write and read.

The macOS `/var -> /private/var` case the comment cites is safe only because that symlink sits *above* the base and is left unresolved symmetrically on both sides; a symlinked base *leaf* is not. Recommend resolving `_basePath` to its real path once (the constructors already normalize with `Path.GetFullPath` at lines 131/162 — add a `ResolveReal` there) and comparing candidates against that, so both sides are normalized identically. Also worth a test with a symlinked base — current tests cover traversal and out-of-base pointers but not this.

Libraries/src/Amazon.Lambda.DurableExecution/docs/core/steps.md
GarrettBeatty: >That's not true for Callback and Invoke.

thats true and intentionally this way. the reason being is these call other lambda functions and it doesnt make sense to use the file system serializer in this case

peterrsongg (approved):

Libraries/src/Amazon.Lambda.DurableExecution/FileSystemSerializer.cs
GarrettBeatty: this is merged in now as part of #2563

Libraries/src/Amazon.Lambda.DurableExecution/docs/core/steps.md
GarrettBeatty: ill leave this for now and update the wording in a follow up if needed.

jnunn-aws (approved):

PR #2576, opened 2026-09-14
Add LMI worker and execution max concurrency counter debug log statement.

normj:
*Issue #, if available:*

*Description of changes:*

Adds a one-time structured DEBUG log emitted during init that reports how many worker (invoke processing) tasks the runtime started and the execution environment's max concurrency. This gives observability into the worker pool sizing when running in Lambda managed instances (multi-concurrency) mode.

**Behavior**

- Emitted only when both conditions hold:
  - The function is running in multi-concurrency mode (`AWS_LAMBDA_MAX_CONCURRENCY` set), and
  - The JSON log format is configured (`AWS_LAMBDA_LOG_FORMAT` / the .NET RIC log-format variable resolves to `Json`).
- Emitted at `Debug` level, so it only surfaces when the function log level is `Debug`/`Trace` (filtering handled by the console logger writer).
- Uses the event name `runtime_worker_pool_initializing`, matching the Java and Python Lambda runtimes so the log is queryable consistently across languages.
- Structured JSON fields: `event`, `workerCount`, `executionEnvironmentMaxConcurrency`.

**Changes**

- `Utils.cs`: new `IsJsonLogFormat` and `EmitWorkerPoolInitializingLog` helpers, plus the event-name and message-template constants.
- `LambdaBootstrap.cs`: calls the emitter once during processing-loop setup, after the worker count is determined.
- Tests: `UtilsTest`, `LogMessageFormatterTests`, and `LambdaBootstrapMultiConcurrencyTests`, plus a new `CapturingConsoleLoggerWriter` test helper.
- Adds an autover changelog entry (patch) for `Amazon.Lambda.RuntimeSupport`.


GarrettBeatty (approved):

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/Utils.cs
```diff
@@ -21,12 +21,59 @@ namespace Amazon.Lambda.RuntimeSupport.Helpers
...
+        /// (Lambda managed instances) and the JSON log format is configured. The DEBUG level means the log is only
+        /// surfaced when the function log level is set to Debug or Trace; that filtering is handled by the console
+        /// logger writer so it is not re-checked here.
+        /// </summary>
+        internal static void EmitWorkerPoolInitializingLog(IConsoleLoggerWriter consoleLogger, IEnvironmentVariables environmentVariables, int workerCount, int maxConcurrency)
+        {
+            if (consoleLogger == null)
+                return;
+
+            if (!IsUsingMultiConcurrency(environmentVariables) || !IsJsonLogFormat(environmentVariables))
```
ashishdhingra, line 66: nit: Is it safe to assume that `environmentVariables` is always non-null when passed to this method? I do see existing methods lacking the check, so assuming it is always non-null.

ashishdhingra (approved):

Libraries/src/Amazon.Lambda.RuntimeSupport/Helpers/Utils.cs
normj: It is always true.

PR #2578, opened 2026-09-17
Force disable proxy when communicating with Lambda runtime API.

normj:
*Description of changes:*

The `HttpClient` that `Amazon.Lambda.RuntimeSupport` uses to communicate with the Lambda Runtime API (RAPID) now forces proxy usage off. Its underlying `SocketsHttpHandler` is created with `UseProxy = false`.

**Problem:** `SocketsHttpHandler` honors `HttpClient.DefaultProxy`, which is populated from the `HTTP_PROXY`/`HTTPS_PROXY` environment variables. When a customer configures a proxy, internal Runtime API calls were being routed through it. That fails because the RAPID endpoint is not necessarily on a loopback address (only `localhost` is implicitly exempt in most proxy implementations), so the runtime could not talk to the Runtime API.

**Fix:** Disable the proxy on the Runtime API handler only. This is scoped to the `HttpClient` used for Runtime API communication and has no impact on HTTP calls made from customer handler code.

This mirrors the equivalent fix made in the Java Lambda runtime client.

*Testing:*
* Added `RuntimeApiHttpClientBypassesProxy` unit test asserting the Runtime API handler has `UseProxy == false`.


ashishdhingra (approved):
LGTM

GarrettBeatty (approved):

dscpinheiro (approved):

